PR #528, opened 2013-12-11
Added ability to enable Webdav support to a vhost directory

chriscowley:
As I wanted to enable Webdav on certain directories in my environment I decided to roll-up my sleeves and get it done.

This should also enable HTTP access to Git repos to be controlled in Puppet too.


janvonde:
Anybody there that can update those tests to get this pull request accepted?

PR #838, opened 2014-09-04
fix allowes for directories

pondohva:
There is typo in template for vhost, in apache configuration allow requires two arguments, 'from' and ip/hostname.


faker-:
I agree that the `from` should be in the template, but it's currently documented (https://github.com/puppetlabs/puppetlabs-apache/#allow) to be used like this:

      directories => [
        { path  => '/path/to/directory',
          allow => 'from example.org',
        },

This change would be breaking all manifests currently using that syntax.

mhaskel:
Good catch @faker- !

@pondohva I'm going to close this since I'm not sure we want to make this backwards incompatible change. Thanks for the contribution though!

igalic:
@mhaskel maybe it could've been a candidate for 2.0.x?

PR #848, opened 2014-09-12
Modify behaviour of MODULES-1294

toote:
This change makes it so that "Require all granted" is only added to directory entries if the require property is undefined, keeping default behaviour. But if the require property is defined as an empty string or false (or any "falsey" value) it will not add the line.

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -39,7 +39,7 @@
     <%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
       <%- if directory['require'] and directory['require'] != '' -%>
     Require <%= Array(directory['require']).join(' ') %>
-      <%- else -%>
+      <%- elsif directory['require'].nil? -%>
     Require all granted
       <%- end -%>
     <%- else -%>


mhaskel:
@toote ping?

toote:
my bad, this fell through the cracks as I moved on to other issues... will try to take another jab at it during the weekend

mhaskel:
Great, thanks @toote. Let us know if you need a hand with anything.

toote:
@mhaskel I think I have a first attempt at tests here... but it is not clear to me how to best do the rebase. Should I do it on a different branch? If so, how do we avoid losing this discussion?

mhaskel:
Hi @toote ,

If you fetch the code from the upstream (puppetlabs) repo you should be able to do a `git rebase upstream/master` when you have the `directory_require_MODULES-1294` branch checked out. Resolve merge conflicts, then you can `git push -f origin directory_require_MODULES-1294` and then this PR will get the updated code.

DavidS:
Hi @toote, thanks for the work you have put in to this pull request. However due to inactivity and no tests added this pull request is being closed. If you want to re-submit this PR with the necessary changes, we would gladly look at the new pull request.

PS: See http://git-scm.com/book/en/Git-Branching-Rebasing for advice on rebasing.

PR #880, opened 2014-10-01
(#1391) Correct Ubuntu Trusty mod_prefork package name

eshamow:
This module calls apache2-prefork-dev as the package
name for the mod_prefork module for all Debian-flavored Linuxes. However
Ubuntu 14.04 changes this to apache2-dev and makes apache2-prefork-dev
an alias.

The result is that calling this resource, even with the package
installed, always results in a package changed purged to present
message, as it's scanning dpkg for apache2-prefork-dev, finding the
package missing, and then installing the alias via apt.


mhaskel:
@eshamow ping?

PR #886, opened 2014-10-09
Updated README.md documentation for setenv to include an example array. ...

DevoKun:
* Updated README.md by adding an example for the **setenv** vhost variable to clarify the Array usage.
* With the recent discovery of the **POODLE** exploit (http://www.net-security.org/secworld.php?id=17495) changed the CipherSuite default to be the Mozilla Foundation recommendation (https://wiki.mozilla.org/Security/Server_Side_TLS) and the Protocol default to be TLSv1.


mhaskel:
Hi @DevoKun, thanks for the contribution, and apologies it took so long to review this. We merged https://github.com/puppetlabs/puppetlabs-apache/pull/917 to allow for fixing things like POODLE. We don't want to change defaults to be the Mozilla Foundation recomendations though, so I'm going to close this PR now.

PR #907, opened 2014-10-22
RFC: Adding SSL Stapling Support

dteirney:
This is a very rough attempt to get feedback on whether this is the right approach to take to add SSL Stapling support.

Based on recommended best practices at https://wiki.mozilla.org/Security/Server_Side_TLS#Apache

### DIFF
--- a/manifests/mod/ssl.pp
+++ b/manifests/mod/ssl.pp
@@ -2,7 +2,6 @@ class apache::mod::ssl (
   $ssl_compression = false,
   $ssl_options     = [ 'StdEnvVars' ],
   $ssl_cipher      = 'HIGH:MEDIUM:!aNULL:!MD5',
-  $ssl_stapling    = false,
   $apache_version  = $::apache::apache_version,
   $package_name    = undef,
 ) {
@@ -52,6 +51,7 @@ class apache::mod::ssl (
   # $ssl_compression
   # $ssl_options
   # $session_cache,
+  # $stapling_cache,
   # $ssl_mutex
   # $apache_version
   #
--- a/manifests/mod/ssl.pp
+++ b/manifests/mod/ssl.pp
@@ -13,9 +13,9 @@ class apache::mod::ssl (
   }
 
   $stapling_cache = $::osfamily ? {
-    'debian'  => "\${APACHE_RUN_DIR}/ocsp(128000)", # TODO: Better way?
-    'redhat'  => '/var/run/ocsp(128000)', # TODO: Confirm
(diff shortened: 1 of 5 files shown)


libb-city:
Hi David

My apologies for taking ions getting back to you.

I will need to play around with the configuration to be able to answer your
questions as best as i can.

I haven't been following the module changelog as of late.

Will answer you soonest.

Kind Regards

templates/mod/ssl.conf.erb
```diff
@@ -10,6 +10,9 @@
   SSLPassPhraseDialog  builtin
   SSLSessionCache "shmcb:<%= @session_cache %>"
   SSLSessionCacheTimeout 300
+<% if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
+  SSLStaplingCache "shmcb:<%= @stapling_cache %>"
```
kronn: I may be wrong, but shouldn't the `SSLStaplingCache` only be set if ssl_stapling is activated?

I propose this conditional:

```ruby
if @ssl_stapling && scope.function_versioncmp([@apache_version, '2.4']) >= 0
```

libb-city: And your correct as far as i know.

if ssl stapling cache directive set without ssl stapling on, apache would
fail to start.

Changes look ok. Seems fair enough.

manifests/mod/ssl.pp
```diff
@@ -11,6 +11,12 @@
     'freebsd' => '/var/run/ssl_scache(512000)',
   }
 
+  $stapling_cache = $::osfamily ? {
+    'debian'  => "\${APACHE_RUN_DIR}/ocsp(128000)",
+    'redhat'  => '/var/cache/mod_ssl/ocsp(128000)',
+    'freebsd' => '/var/run/ocsp(128000)',
```
mhaskel: Is there any reasonable value we can set as a default case? Otherwise this will cause issues with strict_variables.

libb-city: Im failing to understand why the cache value would cause issues
with strict_variables.

Please note that the cache value i have used is the apache default. It can
be tweaked.

mhaskel: @libb-city The issue is there's no default case for the selector.

libb-city: Hmmm. might i suggest we enable cache to be set as a variable that can be
set to desired value during config.

I think this then bypasses the need to set default

jonnytpuppet:
Hi There

I'm closing this PR due to inactivity and also due to the inability to merge. If you wish to have this feature included, please re-submit your PR after rebasing. Thanks!

PR #915, opened 2014-10-24
Modules-1458 mod_wsgi package and module name/path

jantman:
mod_wsgi must be compiled specifically for the python version in use. This change allows apache::mod::wsgi to take a specified package name and module name/path, in order to use a non-default python (i.e. the ones provided by the IUS Community repository, to use newer-than-default python versions on RHEL/CentOS). Behavior is unchanged if the params are left at their defaults.

### DIFF
--- a/README.md
+++ b/README.md
@@ -729,6 +729,22 @@ For customized parameters, which tell Apache how Python is currently configured 
     }
 ```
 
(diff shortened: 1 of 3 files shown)


igalic:
thank you @jantman!

PR #923, opened 2014-10-30
Add support for mod_auth_cas module configuration

pcfens:
Add support to configure mod_auth_cas within Apache.

### DIFF
--- a/manifests/mod/auth_cas.pp
+++ b/manifests/mod/auth_cas.pp
@@ -1,0 +1,48 @@
+class apache::mod::auth_cas (
+  $cas_login_url,
+  $cas_validate_url,
+  $cas_cookie_path          = $::apache::params::cas_cookie_path,
+  $cas_version              = 2,
+  $cas_debug                = 'Off',
+  $cas_validate_depth       = undef,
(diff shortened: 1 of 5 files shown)


manifests/mod/auth_cas.pp
```diff
@@ -0,0 +1,48 @@
+class apache::mod::auth_cas (
+  $cas_login_url            = undef,
+  $cas_validate_url         = undef,
```
mhaskel: Since these two variables are required, they should not be set to undef but rather have no default value.

mhaskel:
@pcfens thanks for the contribution, and sorry it took so long for us to get back to this. I think this is pretty close to being ready to merge. I made one inline comment that needs to be addressed, and it looks like tests are failing against 1.8.7, but that's possibly been fixed upstream, so if you rebase against master we'll have a better idea. thanks!

pcfens:
@mhaskel Thanks for the feedback - I've made the suggested changes and fixed the test issues as well.

Thanks again.

mhaskel:
Looks good, thanks @pcfens !

PR #928, opened 2014-11-11
Omit priority prefix if {priority => false} passed to vhost/custom_config

jbrunner:
Do not prefix filenames in apache::vhost and apache::custom_config when
priority is set to false.


mhaskel:
I picked this up with a rebase and some fixup in #1014

mhaskel:
Thanks for the contribution @jbrunner ! This was merged in #1014

PR #845, opened 2014-09-10
Update _additional_includes.erb

cristi1979:
fix Include statement for apache 2.4

PR #937, opened 2014-11-19
Fix additional_includes on Apache 2.4

mcanevet:
Fix include on Apache 2.4 when no matching files exist

### DIFF
--- a/templates/vhost/_additional_includes.erb
+++ b/templates/vhost/_additional_includes.erb
@@ -1,5 +1,10 @@
 <% Array(@additional_includes).each do |include| -%>
   
   ## Load additional static includes
-  Include "<%= include %>"
+  <IfVersion < 2.4>
+    Include "<%= include %>"
+  </IfVersion>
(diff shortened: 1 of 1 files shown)


mremy:
We have the following use case:
we have several vhosts, and need to let other people manage their own configuration per vhost.
Without this patch, we have to write a file with only a comment in each vhost conf directory.
It works, but it's annoying.

I understand your opinion. Would you agree with a parameter like "allow_empty_conf_dir" to switch to `IncludeOptional` with `Include` remaining the default ?

igalic:
*ack*

can someone (@mcanevet @mremy @ckaenzig) here add some documentation for this, then i'll hit the merge button!

mhaskel:
@mcanevet thanks for the contribution!

@mcanevet @mremy @ckaenzig I merged #845 earlier this morning and I believe that covers the use case you're looking for. I'm going to close this PR for now, but if that doesn't cover you're needed usecase please feel free to resubmit. Thanks again!

PR #940, opened 2014-11-19
Adding the ability to manage AuthExternal parameter.

bengoa:
Hey guys.

This patch save my day when I tried to authenticated against unixgroup users and I was unable to manage this parameter trough the original module, so I've created this little patch.

I hope this could be helpful for you.

Cheers,
Alberto


mhaskel:
Hi @bengoa . Thanks for the contribution, and apologies it took so long to get back to you on this. `apache::mod::authz_unixgroup.load` is not a valid puppet class name, so that should be called something else. The two new `apache::mod::*` classes should also be listed in the README. Can you fix those and squash this down to a single commit? Thanks again!

jonnytpuppet:
@bengoa I'm closing this PR due to inactivity. Please feel free to open a new PR with the suggestions from @mhaskel 

Thanks!

PR #943, opened 2014-11-26
make icons directorylisting configurable

nbeernink:
This should allow people to disable directory index on the icons/ dir via an override.

### DIFF
--- a/manifests/mod/alias.pp
+++ b/manifests/mod/alias.pp
@@ -1,5 +1,6 @@
 class apache::mod::alias(
-  $apache_version = $apache::apache_version
+  $apache_version = $apache::apache_version,
+  $icons_options  = 'Indexes MultiViews',
 ) {
   $ver24 = versioncmp($apache_version, '2.4') >= 0
 
--- a/templates/mod/alias.conf.erb
+++ b/templates/mod/alias.conf.erb
@@ -1,7 +1,7 @@
 <IfModule alias_module>
 Alias /icons/ "<%= @icons_path %>/"
 <Directory "<%= @icons_path %>">
-    Options Indexes MultiViews
+    Options <%= @icons_options %>
     AllowOverride None
 <%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
     Require all granted
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
         * [Class: apache::default_mods](#class-apachedefault_mods)
         * [Defined Type: apache::mod](#defined-type-apachemod)
         * [Classes: apache::mod::*](#classes-apachemodname)
+        * [Class: apache::mod::alias](#class-apachemodalias)
         * [Class: apache::mod::event](#class-apachemodevent)
         * [Class: apache::mod::info](#class-apachemodinfo)
         * [Class: apache::mod::pagespeed](#class-apachemodpagespeed)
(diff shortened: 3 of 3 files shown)


mhaskel:
@nbeernink thanks for the contribution, and sorry it took so long for us to review it. I have a couple of comments on your PR, but I'll add them inline. Thanks again!

manifests/mod/alias.pp
```diff
@@ -6,6 +6,7 @@
     'redhat'  => '/var/www/icons',
     'freebsd' => '/usr/local/www/apache22/icons',
   }
+  $icons_index = 'Indexes Multiviews'
```
mhaskel: I think this should be added as a parameter for `apache::mod::alias` if you actually want to be able to set it.

templates/mod/alias.conf.erb
```diff
@@ -1,7 +1,7 @@
 <IfModule alias_module>
 Alias /icons/ "<%= @icons_path %>/"
 <Directory "<%= @icons_path %>">
-    Options Indexes MultiViews
+    <%= @icons_index %>
```
mhaskel: Should this be `Options <%= @icons_index %>` ?

nbeernink:
I patched the code but ran out of time trying to solve the problem of not being able to push to this PR. Will try again later.

manifests/mod/alias.pp
```diff
@@ -11,6 +12,7 @@
     },
     'freebsd' => '/usr/local/www/apache22/icons',
   }
+  $icons_options = $icons_options
```
ekohl: Isn't this redundant?

nbeernink: Ah you're right.

ekohl:
Test failures are unrelated and should be fixed with https://github.com/puppetlabs/puppetlabs-concat/pull/263

manifests/mod/alias.pp
```diff
@@ -1,5 +1,6 @@
 class apache::mod::alias(
-  $apache_version = $apache::apache_version
+  $apache_version = $apache::apache_version,
+  $icons_options  = 'Indexes Multiviews',
```
mhaskel: This should be `'Indexes MultiViews'`

mhaskel:
@nbeernink I commented above on one more code update to not change default behavior. Could you also update unit tests [here](https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/alias_spec.rb) to exercise the new parameter, and update the README to indicate that `apache::mod::alias` has parameters that can be set (should just be adding a `*`). Thanks!

nbeernink:
@mhaskel, added the docs and corrected the multiviews option, but I don't know how I would write the tests you require. Hope you could point me in the right direction or add them for me. :)

igalic:
@nbeernink n.b.: you also need to [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing) this pr now…

as for tests, take a look at other [spec tests for mod_alias](https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/alias_spec.rb)

mhaskel:
@nbeernink you'll need to add a new block in the tests where you add a :params block like https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/auth_cas_spec.rb#L4-L9 and check that the file contains the values you're setting (like https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/authnz_ldap_spec.rb#L26-L28)

PR #945, opened 2014-11-28
Moved vhost default for overide and options to parent class.

brentclark:
Good day

I make heavy use of hiera in my environment.

As opposed to having to define apache parameters options and override for each virtual host.
I would like to propose that it could be defined it in one location (i.e.  hieras default look up variable (e.g. common.yaml)).

I hope I have made my case clear, and that you will give my pull request consideration.

Kind Regards
Brent Clark


mhaskel:
@brentclark thank you for your contribution, and apologies on the delay in getting back to you. A couple of comments. For future contributions, you may find it useful to use the [GitHub Flow workflow](https://guides.github.com/introduction/flow/). 

It seems like the initial intent of your PR was the changes in 40dbca0. While I can see where that would be useful, that is not a change I'd like to see in the module at this time. I worry that this would lead to too many parameters being added to `class apache` over time, so your best bet is the `create_resources` default option @igalic mentioned above.

It looks like you added some commits for mod_expires. Ideally those should have been submitted as part of a separate pull request. I have a few comments on those additions, but I'll make those inline.

If you want to address the comments for `mod_expires`, rebase this pull request, and remove 40dbca0 I think we should be able to get this merged. Alternatively you can submit a new pull request with only the commit for `mod_expires`.

Thanks again!

manifests/mod/expires.pp
```diff
@@ -1,3 +1,24 @@
-class apache::mod::expires {
+class apache::mod::expires(
```
mhaskel: Since you added new parameters they should get documented in the README.

mhaskel: There should also be unit test coverage added for the new parameters.

jonnytpuppet:
@brentclark I'm closing this due to inactivity. Please feel free to open a new PR with the suggestions from @mhaskel

Thanks!

PR #950, opened 2014-12-04
MODULES-1554: update error docs and icons path for RHEL7-based systems

jlambert121:
https://tickets.puppetlabs.com/browse/MODULES-1554


igalic:
thank you @jlambert121!

PR #951, opened 2014-12-04
MODULES-1382: support multiple access log directives

jlambert121:
MODULES-1382: https://tickets.puppetlabs.com/browse/MODULES-1382


igalic:
thank you @jlambert121!

PR #968, opened 2014-12-23
design plan: implmenet mod_macro

igalic:


igalic:
@mhaskel well, i'm a big fan of rebasing, but *that* would require `--force` :D

mhaskel:
ok, yeah, then pulls would be best. `--force` is the worst :)

PR #977, opened 2015-01-03
Set keepalive to On by default

mmoll:
According to Apache documentation this is ```On``` by default since at least httpd 2.0.

Disclaimer: I want this change because I have a situation where "include apache" is there multiple times and so it's not possible to set that parameter by normal class parameter ways (no hiera involved).

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ Changes the base location of the configuration directories used for the apache s
 
 #####`keepalive`
 
-Enables persistent connections.
+Enables persistent connections. Defaults to 'On'.
 
 #####`keepalive_timeout`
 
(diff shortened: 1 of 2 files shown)


brentclark:
I second this.
Seems silly to have it Off.

igalic:
@mhaskel: we should merge this into 2.0.x

mhaskel:
Manually merged into 2.0.x at https://github.com/puppetlabs/puppetlabs-apache/commit/49054018ffe4524c36c7ef59b8ed65518d0ed3c4

mhaskel:
@mmoll thanks for the contribution!

PR #978, opened 2015-01-05
Allow includes necessary for basic authentication

Micsi:
allow some include apache::mod:: statements needed for basic_auth. The modules could not be inclued if you set default_mods => false. Includes were present in default_modules definition, though.
Tested on Debian wheezy 7.7


igalic:
@Micsi this is really in dire need of a proper [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing)

PR #982, opened 2015-01-16
Documentation fix: ssl_protocol expects a string, not an array.

Farzy:
The README.md sample uses an array where a string is expected.


igalic:
iirc, i asked for both to be possible, since it's actually quite easy to achieve.
i guess it was lost in the migration to concat… >_> @mhaskel :p

PR #983, opened 2015-01-16
add support for PassengerEnabled, avoid adding blank SSL values

hdeadman:
PassengerEnabled option added b/c:
```
  RailsAutoDetect, RackAutoDetect and WsgiAutoDetect
  These options have been removed in version 4.0.0 as part of an optimization. You should use PassengerEnabled instead.
```

In the SSL module, changed template to avoid setting values if they are blank. Trying to use a module that passes in a value for these but I want to override it in hiera with a blank string so it doesn't get set in the config. Maybe I could have set the value to false in hiera but my change is consistent with what was in place for another optional SSL value in the template.

### DIFF
--- a/templates/vhost/_ssl.erb
+++ b/templates/vhost/_ssl.erb
@@ -4,22 +4,22 @@
   SSLEngine on
   SSLCertificateFile      "<%= @ssl_cert %>"
   SSLCertificateKeyFile   "<%= @ssl_key %>"
-  <%- if @ssl_chain -%>
+  <%- if @ssl_chain && @ssl_chain != '' -%>
   SSLCertificateChainFile "<%= @ssl_chain %>"
   <%- end -%>
   <%- if @ssl_certs_dir && @ssl_certs_dir != '' -%>
   SSLCACertificatePath    "<%= @ssl_certs_dir %>"
   <%- end -%>
-  <%- if @ssl_ca -%>
+  <%- if @ssl_ca && @ssl_ca != '' -%>
   SSLCACertificateFile    "<%= @ssl_ca %>"
   <%- end -%>
-  <%- if @ssl_crl_path -%>
+  <%- if @ssl_crl_path && @ssl_crl_path != '' -%>
   SSLCARevocationPath     "<%= @ssl_crl_path %>"
   <%- end -%>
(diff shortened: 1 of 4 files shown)


templates/vhost/_ssl.erb
```diff
@@ -4,22 +4,22 @@
   SSLEngine on
   SSLCertificateFile      "<%= @ssl_cert %>"
   SSLCertificateKeyFile   "<%= @ssl_key %>"
-  <%- if @ssl_chain -%>
+  <%- if @ssl_chain && @ssl_chain != '' -%>
```
igalic: why don't we just check for `@ssl_chain.empty?` instead?

igalic:
i'm a bit unhappy with the fact that you're munging a number of different flavours of changes in one pr. for one, for the other, there's a merge-commit in the middle of them. that shouldn't be.

templates/vhost/_ssl.erb
hdeadman: I will defer to you, I am new to Ruby and was just copying what was already done in the template for SSLCACertificatePath.

hdeadman:
Thanks for the link to github flow, I am new to this and will follow that for future patch requests. I had been trying to get the puppet-puppet module from ploperations working on Centos 7 and I have given up so I may not have time in the short term to re-do this PR. Feel free to take the changes or I can just close this.

templates/vhost/_ssl.erb
igalic: \*nod\*
in that case, that template, well, our templates in general, need some fixing ;)

mhaskel:
@hdeadman Thanks for the contribution! Since it sounds like you aren't going to have time to work on this in the short term I'm going to close this PR for now. If/when you get a chance to pick up this work again, here are a few extra tips to get this merged in more easily.
* This should definitely be broken out into 2 separate pull requests (one for PassengerEnabled, one for empty values in SSL) (the versioncmp change has already been merged)
* Since the PassengerEnabled change is a new feature, there should be docs updates in the README, and at least unit test coverage for the new parameter (added to https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/passenger_spec.rb)

Thanks again for the contribution!

PR #985, opened 2015-01-18
MODULES-1622: Allow multiple Deny directives in a directory

faker-:
Allow multiple `Deny` directives in the same way as `Allow` already works, example:  

    directories => [
      { 'path' => '/var/ww/first',
        'deny' => [ 'from 127.0.0.1', 'from 127.0.0.2', ],
      },

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -147,11 +147,20 @@ describe 'apache::vhost', :type => :define do
           'access_log_format'           => '%h %l %u %t \"%r\" %>s %b',
           'access_log_env_var'          => '',
           'aliases'                     => '/image',
-          'directories'                 => {
-            'path'     => '/var/www/files',
-            'provider' => 'files',
-            'deny'     => 'from all'
-          },
+          'directories'                 => [
+            {
+              'path'     => '/var/www/files',
+              'provider' => 'files',
+              'allow'    => [ 'from 127.0.0.1', 'from 127.0.0.2', ],
+              'deny'     => [ 'from 127.0.0.3', 'from 127.0.0.4', ],
+            },
+            {
+              'path'     => '/var/www/foo',
+              'provider' => 'files',
+              'allow'    => 'from 127.0.0.1',
+              'deny'     => 'from all',
+            },
+          ],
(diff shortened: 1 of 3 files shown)


igalic:
@faker- i *think* we have tests for this.
would you mind updating those too?

mhaskel:
@faker- there are some tests in https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/defines/vhost_spec.rb. It would be great if you could add a case with `deny` set to an array just to make sure nothing explodes. Otherwise this looks good. Thanks for the contribution!

faker-:
Sorry for the delay, I had trouble setting up an environment for the tests. 
Can you check my last commit?
I'm not entirely my changes are all that is required.

mhaskel:
@faker- if there could also be a test case with something like `deny => 'from all'` just to make sure no array-specific logic isn't properly wrapped that would be great. Thanks!

faker-:
Makes sense, added with my last commit.

igalic:
@faker- would you mind [squashing](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) your commits down a bit?

faker-:
@igalic squashed them to one commit

spec/defines/vhost_spec.rb
```diff
@@ -147,11 +147,20 @@
...
+              'allow'    => [ 'from 127.0.0.1', 'from 127.0.0.2', ],
+              'deny'     => [ 'from 127.0.0.3', 'from 127.0.0.4', ],
+            },
+            {
+              'path'     => '/var/www/foo',
+              'provider' => 'files',
+              'allow'    => 'from 127.0.0.1',
+              'deny'     => 'from all',
+            },
+          ],
```
igalic: i just realized, you're not checking whether this sets or changes anything down the line

faker-:
@igalic you are right, I just extended the existing "deny from all" test.
Turns out that deny was not getting applied anyway (since Apache version in that block is set to 2.4).

Can you check my last commit again?

mhaskel:
Looks good, thanks @faker- !

PR #986, opened 2015-01-19
add gitignore rule to ignore vim swap files...

bechtoldt:
... again that have been removed by 8bd6b6ce8e5f224ee2ab0739446ff1aa2ead048b

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ spec/fixtures/
 coverage/
 .idea/
 *.iml
+.*sw*


igalic:
@bechtoldt i was told the reason is that we, vim users, (unlike [intellij users](https://github.com/puppetlabs/puppetlabs-apache/pull/970) ;p ) can add those to the [global git ignore](http://git-scm.com/docs/gitignore) (~/.config/git/ignore).

bechtoldt:
Well I don't see a serious use case where I want to have swap files in the index or ignore swap files on a global level.

mhaskel:
@bechtoldt Thanks for the contribution. I'm not sure if there's some greater reason for `.swp` files to not be in `.gitignore`, but it seems like a reasonable addition. However, `.gitignore` is managed by https://github.com/puppetlabs/modulesync_configs/blob/master/moduleroot/.gitignore so we can keep it consistent across modules. I'm going to close this PR, and please submit this change as a PR against https://github.com/puppetlabs/modulesync_configs. Thanks again!

PR #987, opened 2015-01-20
MODULES-1684: Specify mod_proxy_connect module for Apache >= 2.3.5

holser:
mod_proxy_connect was moved to own module since Apache >= 2.3.5.
AllowConnect directive won't work without this module on Ubuntu 14.04 or
Debian 8.

### DIFF
--- a/manifests/mod/proxy_connect.pp
+++ b/manifests/mod/proxy_connect.pp
@@ -1,0 +1,8 @@
+class apache::mod::proxy_connect (
+  $apache_version  = $::apache::apache_version,
+) {
+  if versioncmp($apache_version, '2.4') >= 0 {
+    Class['::apache::mod::proxy'] -> Class['::apache::mod::proxy_connect']
+    ::apache::mod { 'proxy_connect': }
+  }
+}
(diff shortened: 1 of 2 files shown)


manifests/mod/proxy_connect.pp
```diff
@@ -0,0 +1,8 @@
+class apache::mod::proxy_connect (
+  $apache_version  = $::apache::apache_version,
+) {
+  if versioncmp($apache_version, '2.4') >= 0 {
```
igalic: i'm confused. this means we "succeed" to "load" this module?
shouldn't we fail here?

holser: We shouldn't fail here. If we succeed then we should load module

igalic:
okay, thanks @holser!

igalic:
@hunner just drew my attention to [MODULES-1874](https://tickets.puppetlabs.com/browse/MODULES-1874) so i checked httpd's svn history, and, mod_proxy_connect has been around since 2005, and is|was definitely in 2.2.

we have to remove that whole `if`, and just load the module regardless.

PR #989, opened 2015-01-20
Add mod_remoteip module

dsavineau:
http://httpd.apache.org/docs/2.4/mod/mod_remoteip.html

### DIFF
--- a/manifests/mod/remoteip.pp
+++ b/manifests/mod/remoteip.pp
@@ -1,0 +1,27 @@
+class apache::mod::remoteip (
+  $header            = 'X-Forwarded-For',
+  $proxy_ips         = [ '127.0.0.1' ],
+  $proxies_header    = undef,
+  $trusted_proxy_ips = undef,
+  $apache_version    = $::apache::apache_version
+) {
+  if versioncmp($apache_version, '2.4') < 0 {
+    fail('mod_remoteip is only available in Apache 2.4')
+  }
+
+  ::apache::mod { 'remoteip': }
+
+  # Template uses:
+  # - $header
+  # - $proxy_ips
+  # - $proxies_header
+  # - $trusted_proxy_ips
+  file { 'remoteip.conf':
+    ensure  => file,
+    path    => "${::apache::mod_dir}/remoteip.conf",
+    content => template('apache/mod/remoteip.conf.erb'),
+    require => Exec["mkdir ${::apache::mod_dir}"],
+    before  => File[$::apache::mod_dir],
+    notify  => Service['httpd'],
+  }
+}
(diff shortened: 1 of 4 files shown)


manifests/mod/remoteip.pp
```diff
@@ -0,0 +1,27 @@
+class apache::mod::remoteip (
+  $header            = 'X-Forwarded-For',
+  $proxy_ips         = [ '127.0.0.1' ],
+  $proxies_header    = undef,
+  $trusted_proxy_ips = undef,
```
igalic: this class has n configurable parameters parameters exposed, none of which are documented in the README.

manifests/mod/remoteip.pp
```diff
@@ -0,0 +1,27 @@
...
+  $apache_version    = $::apache::apache_version
```
igalic: i'm not sure it's necessary to pass this as parameter, rather just use it as variable, external or localized.

mhaskel: +1 to removing `$apache_version` here

mhaskel:
@dsavineau thanks for the contribution! Other than some documentation for the parameters you added and removing the `apache_version` parameter this looks good.

manifests/mod/remoteip.pp
dsavineau: @igalic no problem I will add those parameters to the README.

manifests/mod/remoteip.pp
dsavineau: @igalic @mhaskel I'm not sure to understand why we need to remove apache_version here because in most of the others apache module manifests (worker, itk, etc...) it's used like this.
The main reason is to allow to override the apache version if the auto detection doesn't match with the setup :
example : the auto detection on Debian Wheezy (7.x) returns 2.2 but if you have custom packages with 2.4 version this manifest will not be applied

mhaskel: ah, you have a good point.

manifests/mod/remoteip.pp
mhaskel: Actually, the pattern in the readme for apache::mod::* classes is just to have a * after the name if they have parameters, no further docs in the README, so this looks good!

mhaskel:
@dsavineau Ok, I think the only thing I'd like to see changed for this is add [validate_array](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/validate_array.rb) calls for the two parameters that you're expecting arrays for. Once that's added we'll be able to get this merged. Thanks!

igalic:
I'm surprised a global override doesn't fix that

jonnytpuppet:
ping @dsavineau is there any update on this? Thanks!

igalic:
@jonnytpuppet not sure an update is necessary, after merging  #1065

jonnytpuppet:
@igalic good catch!

Closing due to #1065

PR #900, opened 2014-10-17
Rewrite_map option added to vhost

gerapeldoorn:
Added rewrite_map option, needed this for use with fail2ban.

Readme also updated.

PR #990, opened 2015-01-21
Add RewriteMap support

soerenbe:
Here is the patch to support apache RewriteMap.

Since this is my first contribution/pull request every feedback is very wellcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1327,7 +1327,7 @@ Modifies collected [request headers](http://httpd.apache.org/docs/current/mod/mo
 
 #####`rewrites`
 
-Creates URL rewrite rules. Expects an array of hashes, and the hash keys can be any of 'comment', 'rewrite_base', 'rewrite_cond', or 'rewrite_rule'. Defaults to 'undef'.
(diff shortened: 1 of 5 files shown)


soerenbe:
Associated ticket: https://tickets.puppetlabs.com/browse/MODULES-1689

igalic:
@soerenbe thank you very much for this contribution.
i *think* we have spec thests for rewrite_stuff. would you mind adding rewrite_map to those tests as well?

soerenbe:
@igalic I will try. Last time i tried it, i didn't get all the spec tests to run on my Ubuntu 14.04 box because of some fancy ruby error. So I gave up after a hour. I will trie it tomorrow.

soerenbe:
@igalic Ok - I got it.
It took me hours to get vagrant running, but finally I got it. I added a the test and it passes.

mhaskel:
Hi @soerenbe, this looks good, thanks! Could you please squash this down to a single commit, then we should be good to merge!

igalic:
@soerenbe ping

soerenbe:
@igalic : Done. 
When you need something else drop me a line.

I just realized that there is a similar (old?) pull request here: https://github.com/puppetlabs/puppetlabs-apache/pull/900

igalic:
@soerenbe thanks you.

re #900: sometimes pull requests are forgotten, either by us, or by their authors :(

PR #991, opened 2015-01-21
Quoted params to versioncmp in mod/alias.pp to avoid type error

gibbsoft:
I've logged a JIRA ticket for this, please see 'MODULES-1692'.

### DIFF
--- a/manifests/mod/alias.pp
+++ b/manifests/mod/alias.pp
@@ -1,7 +1,7 @@
 class apache::mod::alias(
   $apache_version = $apache::apache_version
 ) {
-  $ver24 = versioncmp($apache_version, 2.4) >= 0
+  $ver24 = versioncmp($apache_version, '2.4') >= 0
 
   $icons_path = $::osfamily ? {
     'debian'  => '/usr/share/apache2/icons',


manifests/mod/alias.pp
```diff
@@ -1,7 +1,7 @@
 class apache::mod::alias(
   $apache_version = $apache::apache_version
 ) {
-  $ver24 = versioncmp($apache_version, 2.4) >= 0
+  $ver24 = versioncmp("$apache_version", "2.4") >= 0
```
cmurphy: For consistency with the other places this is used, could you change "2.4" to '2.4' and drop the quotes around $apache_version? It's then the responsibility of the user to input a string for the version. See https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/event.pp#L51 for example.

cmurphy:
Could you also squash it into one commit?

gibbsoft:
I've merged the two commits, does that do it Colleen?

Best regards



Nigel
___


Nigel Gibbs
Cloud Architecture Consultant / Linux DevOps Engineer
Gibbsoft Ltd
w: gibbsoft.com
e: nigel@gibbsoft.com
t: +44 (0) 777 197 4018

cmurphy:
@gibbsoft sorry, what I meant was to make the two commits f84847a and 6efbd68 into one single commit so that the git history is clean. Now there seem to be four commits. What you could do now is, on your master branch:

```
$ git reset HEAD^
```

That should leave you with just a4cbf79 which contains the correct change with an accurate commit message. You can verify with

```
$ git log
```

to see what changes made by you are the history. You can then do

```
$ git push -f origin master
```

to override the remote commits with new single commit.

gibbsoft:
got it

cmurphy:
Perfect, thank you!

PR #992, opened 2015-01-21
modules-1559 apache module no service refresh

tphoney:
adding the option to disable the service refresh after configuration changes.


spec/defines/vhost_spec.rb
```diff
@@ -285,7 +285,7 @@
         'owner'   => 'root',
         'mode'    => '0644',
         'require' => 'Package[httpd]',
-        'notify'  => 'Service[httpd]',
+        'notify'  => 'Class[Apache::Service]',
```
mhaskel: trailing whitespace

tphoney: so should i make a change and commit again ? or wait for it to run through ?

PR #993, opened 2015-01-21
modules-1559 Apache module no service refresh

tphoney:
Apache module no service refresh


mhaskel:
\o/ thanks @tphoney !

PR #994, opened 2015-01-22
MODULES-1680 - sort php_* hashes for idempotency

mhaskel:
They were already being sorted in templates/vhost/_php*.erb, but not in
templates/vhost/_directories.erb


lgbff:
it to run for me, thx.

PR #995, opened 2015-01-22
MODULES-1696: ensure mod::setenvif is included if needed

jlambert121:


manifests/vhost.pp
```diff
@@ -390,6 +390,12 @@
     }
   }
 
+  if ($setenv and ! empty($setenv)) or ($setenvif and ! empty($setenvif)) {
```
igalic: this seems redundant:
```
empty(undef) ==> true
empty([]) ==> true
empty('') ==> true
```

jlambert121: This is just copying https://github.com/jlambert121/puppetlabs-apache/blob/modules_1696/manifests/vhost.pp#L393 - I think those two should match.  The default for these is [], should it instead be undef?  Otherwise this would by default always be true.

igalic:
thanks @jlambert121!

PR #996, opened 2015-01-23
change default version in freebsd to 2.4

sethlyons:
freebsd updated the default version of apache to 2.4 in july of 2014:

(from /usr/ports/updating):
```
20140713:
  AFFECTS: users of www/apache22
  AUTHOR: ohauer@FreeBSD.org

  The default version was changed from www/apache22 to www/apache24,
  pre-build apache modules and web applications will also reflect this!

  In case ports are build by yourself and apache22 is required
  use the following command to keep apache22 as default.

  # echo "DEFAULT_VERSIONS+=apache=2.2" >> /etc/make.conf
```

i think these are all the changes but am waiting to see the travis tests.  i am already using this in production on freebsd 10.1 without any issues.  once we're sure it's ready to merge, i'll rebase down to one commit.


manifests/mod/peruser.pp
```diff
@@ -8,66 +8,63 @@
...
-  }
-
-  ::apache::peruser::multiplexer { '01-default': }
 
   case $::osfamily {
     'freebsd' : {
-      class { '::apache::package':
-        mpm_module => 'peruser'
-      }
+      fail("Unsupported osfamily ${::osfamily}")
```
igalic: do i understand this correctly, that freebsd 2.4 no longer supports peruser?

igalic:
this looks good to me, but then again, i'm no freebsd user, so i wouldn't know.

manifests/mod/filter.pp
```diff
@@ -0,0 +1,3 @@
+class apache::mod::filter {
+  ::apache::mod { 'filter': }
+}
```
mhaskel: This should have a README update

mhaskel:
@sethlyons thanks for the contribution! I found one new class that should have a corresponding addition in the README, but otherwise this looks good (trusting your FreeBSD knowledge :) ).

Also, unfortunately since this is such a big change it's going to need a rebase, looks like we merged something with a conflict. If you ping me once the README is updated and this is rebased I'll try to get it merged right away so we don't have to go through the rebase again. Thanks again!

sethlyons:
changes made.

mhaskel:
Great, thanks @sethlyons !

PR #998, opened 2015-01-27
add support for conf-enabled/*.conf in newer Debian/Ubuntu apache 2.4 setups

Seegras:
Newer Debian/Ubuntu systems have a conf-available/conf-enabled system in which things like 
charset.conf, other-vhosts-access-log.conf,  serve-cgi-bin.conf, localized-error-pages.conf, security.conf and other such necessities reside that used to be in conf.d.

So puppetlabs::apache is broken on Ubuntu/Debian with apache 2.4 without this.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -40,6 +40,7 @@ class apache (
   $httpd_dir              = $::apache::params::httpd_dir,
   $server_root            = $::apache::params::server_root,
   $conf_dir               = $::apache::params::conf_dir,
+  $conf_enable_dir        = $::apache::params::conf_enable_dir,
   $confd_dir              = $::apache::params::confd_dir,
   $vhost_dir              = $::apache::params::vhost_dir,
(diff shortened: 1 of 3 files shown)


igalic:
i wouldn't say *broken*, but definitely not as fully-fledged :p

Seegras:
Hi. I've seen there is another pull request open which would allow the specification of additional_optional_includes https://github.com/puppetlabs/puppetlabs-apache/pull/845 which basically would alleviate this issue, without the need for specifying purge_config false and setting confd_dir to conf-enabled. 
And I fixed the email; it was a wrong one, one of the old employer.

mhaskel:
@Seegras I merged #845 this morning, do you still need this PR or is the needed functionality covered?

Seegras:
I'd guess this solves it.

igalic:
okay so, closing this one then

PR #999, opened 2015-01-28
ssl: Added HSTS settings option for Strict-Transport-Security enablement.

jpds:
Added HSTS settings option for Strict-Transport-Security enablement.

Can be tested doing: curl --insecure -si https://www.example.com/


jpds:
Just realised I could do all this with the header option of ::vhost.

PR #1000, opened 2015-01-28
Symlinks on all distros

HT43-bqxFqB:
Create symlinks on all distros (not just Debian) when a separate $vhost_enable_dir is used. Without this a different $vhost_enable_dir as conf.d on centos does not create symlinks to enable vhosts.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -440,7 +440,7 @@ define apache::vhost(
     require => Package['httpd'],
     notify  => Class['apache::service'],
   }
-  if $::osfamily == 'Debian' {
+  if $::apache::vhost_enable_dir {
     $vhost_enable_dir = $::apache::vhost_enable_dir
     $vhost_symlink_ensure = $ensure ? {
       present => link,
--- a/spec/acceptance/vhost_spec.rb
+++ b/spec/acceptance/vhost_spec.rb
@@ -92,6 +92,39 @@ describe 'apache::vhost define', :unless => UNSUPPORTED_PLATFORMS.include?(fact(
     end
   end
 
+  context 'new vhost with enable symlink' do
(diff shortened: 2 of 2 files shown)


igalic:
uhm… how did this ever work then?
shouldn't our tests catch that??
what about… \*dun-dun-dun\* acceptance tests??

HT43-bqxFqB:
I'll try to find out how to write that into vhost_spec.rb. May take a while... :)

igalic:
thanks, @HT43-bqxFqB! that would be great.

HT43-bqxFqB:
So far it seems that none of the spec/acceptance/vhost_spec.rb get tested in https://travis-ci.org/puppetlabs/puppetlabs-apache/jobs/48768476. What i have found is the following:

* spec/classes/apache_spec.rb contains a specific $vhost_enable_dir part (Line 251), but only for RedHat5 and no symlinking to this directory is tested
* spec/defines/vhost_spec.rb conatins testing for creating a vhost an symlinking it (Line 55), but only for Debian based systems

So far i would guess I'll have to move my test into spec/defines/vhost_spec.rb into a section unspecific of osfamily to test symlink creation, as spec/acceptance/vhost_spec.rb doesn't seem to get tested. Would that be correct?

mhaskel:
@HT43-bqxFqB everything under spec/acceptance gets tested, it just runs on our (currently) internal jenkins setup

mhaskel:
@HT43-bqxFqB this looks good! Can you please squash this down to a single commit, then we can get it merged. Thanks for the contribution!

HT43-bqxFqB:
Today i learned how to squash and reset. Now it is in one commit :)

mhaskel:
Great, thank you @HT43-bqxFqB !

PR #1001, opened 2015-01-31
ssl.pp: Allow setting of SSLRandomSeed option.

jpds:


templates/mod/ssl.conf.erb
```diff
@@ -1,8 +1,16 @@
 <IfModule mod_ssl.c>
   SSLRandomSeed startup builtin
-  SSLRandomSeed startup file:/dev/urandom 512
   SSLRandomSeed connect builtin
+<% if @ssl_random_seeds and ! @ssl_random_seeds.empty? -%>
```
igalic: this first line is pretty much unnecessary, because
```ruby
Array(undef).each do ..
```
skips the whole loop (and is easier to read).

igalic:
General comments: This pr will need documentation & spec test updates!

templates/mod/ssl.conf.erb
jpds: This I took from templates/vhost/_header.erb, but OK, I'll edit it for this one.

igalic: no worries.
i think we should do this clean-up in [2.0.x](https://github.com/puppetlabs/puppetlabs-apache/blob/2.0.x/DESIGN.md)

jpds: So, removing this if statement - how do I set the default pre-existing template options? Putting them in ssl.pp?

igalic: in general, we set the pre-existing default options as *default* value for the class parameter.
which you appear to have already done, since it's about *additional* values of SSLRandomSeed

jpds: Not necessarily, for instance - I set urandom to 1024 bits on my maschinen, removing the default 512 value.

igalic: okay! then just set the default value (512…) in the ssl class.

jpds:
OK, docs and tests should be in place now.

README.md
```diff
@@ -740,6 +740,7 @@ Installs Apache SSL capabilities and uses the ssl.conf.erb template. These are t
       ssl_options            => [ 'StdEnvVars' ],
       ssl_pass_phrase_dialog => 'builtin',
       ssl_protocol           => 'all -SSLv2 -SSLv3',
+      ssl_random_seeds       => [ 'startup builtin', 'startup file:/dev/urandom 512', 'connect builtin', 'connect file:/dev/urandom 512' ],
```
igalic: please break this line up:
```puppet
  ssl_random_seeds => [
    'startup builtin',
    'startup file:/dev/urandom 512',
    'connect builtin',
    'connect file:/dev/urandom 512'
  ],
```

jpds: Was thinking of doing that, done.

igalic:
anyway. this patch *still* needs a [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing)
***
```shell
# assuming you have your own fork checked out as origin:
% git remote -v
origin  git@github.com:jpds/puppetlabs-apache (fetch)
origin  git@github.com:jpds/puppetlabs-apache (push)
% git remote add upstream git@github.com:puppetlabs/puppetlabs-apache # add a new remote: puppetlabs repo, as "upstream"
% git remote update
% git rebase upstream/master # rebase
```
***
also, a [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) down to a single commit would be good.

jpds:
It seems pull #1007 clashes with this one, though I prefer my approach as at least here, the random source and bytes can be tweaked at the same time.

jpds:
There, all done.

igalic:
thank you @jpds!

igalic:
aah :|

PR #1002, opened 2015-02-02
add proper array support for require

Griphon:
now valid access rules are generated from directory template

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -38,7 +38,9 @@
     <%- end -%>
     <%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
       <%- if directory['require'] and directory['require'] != '' -%>
-    Require <%= Array(directory['require']).join(' ') %>
+        <%- Array(directory['require']).each do |require| -%>
+    Require <%= require %>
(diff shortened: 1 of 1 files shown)


igalic:
given that this file seems a frequent cause of issues, we should probably add tests for it…
***
general comments: @Griphon: you may wanna [add your email address to your github settings](https://help.github.com/articles/setting-your-email-in-git/), so your commits are associated with your account.
i also suggest using the [github flow](https://guides.github.com/introduction/flow/index.html) workflow for your next pull request.

Griphon:
@igalic, ok, should I write tests for this patch? It does work for my local puppet environment and array of "Require" will work on any apache

mhaskel:
@Griphon Thanks for the contribution! If you could update tests in https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/defines/vhost_spec.rb to check that the expected file contents are there if `require` is an array or a string. Let us know if you need any additional help!

mhaskel:
@Griphon I updated the tests in #1047, but thanks for the contribution!

Griphon:
@mhaskel Thank you, I'm pleased to contribute but sorry for absence of tests, I had hard time at my work.

PR #1006, opened 2015-02-04
MODULES-1744: use bool2httpd for server_signature and trace_enable

jlambert121:


hunner:
Perhaps instead we should validate that they are actually on/off, as in https://github.com/puppetlabs/puppetlabs-apache/blob/1.2.0/manifests/init.pp#L101 ?

hunner:
As the readme would indicate that those are the values that SHOULD be passed, though not explicitly https://github.com/puppetlabs/puppetlabs-apache#trace_enable https://github.com/puppetlabs/puppetlabs-apache#server_signature

jlambert121:
I thought about validating, but the function already existed so I went with more options for the user.  I'm happy to switch it out for validation, bools make more sense to me though.

PR #1007, opened 2015-02-04
(MODULES-1719) Add parameter for SSLRandomSeed bytes

hunner:


hunner:
The acceptance tests don't work on this yet...

PR #1009, opened 2015-02-09
ssl_protocol expects a string, not an array.

sathieu:
This was lost with #1007 (e8c565).


mhaskel:
Good catch @sathieu ! Thanks for the contribution!

PR #1014, opened 2015-02-11
Fixup 928 - optionally omit priority

mhaskel:
Fixes puppetlabs/puppetlabs-apache#928

### DIFF
--- a/manifests/custom_config.pp
+++ b/manifests/custom_config.pp
@@ -23,8 +23,15 @@ define apache::custom_config (
 
   validate_bool($verify_config)
 
+  if $priority {
+    $priority_prefix = "${priority}-"
+  } else {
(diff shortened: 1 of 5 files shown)


manifests/custom_config.pp
```diff
@@ -23,8 +23,15 @@
 
   validate_bool($verify_config)
 
+  if $priority {
+    $priority_prefix = "${priority}-"
+  } else {
+    $priority_prefix = ''
+  }
```
igalic: i would, as always, prefer this to be a [selector](https://docs.puppetlabs.com/puppet/3.7/reference/lang_conditional.html#selectors)

igalic:
damn. it's already merged.

PR #1021, opened 2015-02-13
add ProxyErrorOverride directive

timogoebel:
This PR adds support for the ProxyErrorOverride directive. http://httpd.apache.org/docs/2.2/mod/mod_proxy.html#proxyerroroverride

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -192,6 +192,7 @@ describe 'apache::vhost', :type => :define do
           'php_admin_values'            => ['true', 'false'],
           'no_proxy_uris'               => '/foo',
           'proxy_preserve_host'         => true,
+          'proxy_error_override'        => true,
           'redirect_source'             => '/bar',
           'redirect_dest'               => '/',
           'redirect_status'             => 'temp',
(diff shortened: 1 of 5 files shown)


mhaskel:
@timogoebel Thanks for the contribution! Could you update the [vhost unit tests](https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/defines/vhost_spec.rb) to exercise this new parameter?

timogoebel:
@mhaskel: I added the same tests as for proxy_preserve_host. What do you think?

spec/defines/vhost_spec.rb
```diff
@@ -192,6 +192,7 @@
           'php_admin_values'            => ['true', 'false'],
           'no_proxy_uris'               => '/foo',
           'proxy_preserve_host'         => true,
+          'proxy_error_override'        => true,
```
igalic: you're not checking this anywhere ;)

timogoebel: Well yes, however proxy_preserve_host is not checked anywhere as well, is it?

igalic: [inconsolable sobbing]

timogoebel: I was unter the impression, that the context was just to "set everything!" and check whether the manifest compiles. What tests do you think are necessary?

timogoebel:
@igalic @mhaskel : Any more feedback on this? Are the tests ok? Can this be merged?

igalic:
@timogoebel i'll merge this one, and i've created an issue to track our broken tests: [MODUlES-1774](https://tickets.puppetlabs.com/browse/MODULES-1774)

thank you for this patch!

PR #1024, opened 2015-02-15
Update README.md

cristi1979:
params from proxy_pass needs to be a hash (considering the erb templates/vhost/_proxy.erb)


cristi1979:
params from proxy_pass needs to be a hash (considering the erb templates/vhost/_proxy.erb)

mhaskel:
Thanks @cristi1979 ! good catch!

PR #1027, opened 2015-02-17
Feature/extentions

stevenpost:
Hi,

While upgrading the puppet apache module from an old version to 1.2.0, I decided to port our custom changes and make them available to all, dropping patches were the functionality had been implemented upstream (here). These changes frequently do not have documentation and never any tests. However I do not have the time to add them.
I add this pull request in hopes that it may be useful to some, or even implemented upstream.

The branch is based on top of tag 1.2.0 (since that is what we use), I'll update it whenever we upgrade the module.  It also includes a commit that is already in master, cherry-picked and adapted to work with this version.
Because of the cherry-picking and changing around, some commit messages also need cleaning before being merged into master.

Regards,
Steven

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -27,6 +27,7 @@ class apache (
   $default_ssl_crl_path   = undef,
   $default_ssl_crl        = undef,
   $default_ssl_crl_check  = undef,
+  $default_type           = 'none',
   $ip                     = undef,
   $service_enable         = true,
   $service_manage         = true,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -17,6 +17,7 @@ class apache (
   $service_name           = $::apache::params::service_name,
   $default_mods           = true,
   $default_vhost          = true,
+  $default_charset        = undef,
   $default_confd_files    = true,
   $default_ssl_vhost      = false,
   $default_ssl_cert       = $::apache::params::default_ssl_cert,
@@ -69,6 +70,7 @@ class apache (
   $package_ensure         = 'installed',
   $use_optional_includes  = $::apache::params::use_optional_includes,
 ) inherits ::apache::params {
+
   validate_bool($default_vhost)
   validate_bool($default_ssl_vhost)
   validate_bool($default_confd_files)
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -55,7 +55,7 @@ class apache (
   $group                  = $::apache::params::group,
   $keepalive              = $::apache::params::keepalive,
   $keepalive_timeout      = $::apache::params::keepalive_timeout,
-  $max_keepalive_requests = $apache::params::max_keepalive_requests,
+  $max_keepalive_requests = $::apache::params::max_keepalive_requests,
   $logroot                = $::apache::params::logroot,
   $logroot_mode           = $::apache::params::logroot_mode,
   $log_level              = $::apache::params::log_level,
--- a/README.md
+++ b/README.md
@@ -434,6 +434,10 @@ Determines whether the HTTPD service is enabled when the machine is booted. Defa
 
 Determines whether the service should be running. Valid values are 'true', 'false', 'running', or 'stopped' when Puppet should manage the service. Any other value sets ensure to 'false' for the Apache service, which is useful when you want to let the service be managed by some other application like Pacemaker. Defaults to 'running'.
 
+#####`default_charset`
+
+If defined, the value will be set as `AddDefaultCharset` in the main configuration file (`httpd.conf`). It is undefined by default.
+
 #####`service_name`
 
 Name of the Apache service to run. Defaults to: 'httpd' on RedHat, 'apache2' on Debian and Gentoo, and 'apache22' on FreeBSD.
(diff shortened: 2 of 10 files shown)


manifests/init.pp
```diff
@@ -53,7 +56,7 @@
   $group                  = $::apache::params::group,
   $keepalive              = $::apache::params::keepalive,
   $keepalive_timeout      = $::apache::params::keepalive_timeout,
-  $max_keepalive_requests = $apache::params::max_keepalive_requests,
+  $max_keepalive_requests = $::apache::params::max_keepalive_requests,
```
igalic: ow did that slip through?

igalic:
hrm… that's a lot to review, and i just realized, i should review it on a commit-by-commit base…

a general comment: you may want to add (all) your email addresses to your github profile, so your commits are all associated with your account

stevenpost:
Hi,

Thanks for wanting to do a quick review. Yes, in essence, non of the commits are related to each other.

For the instances github does not recognise my e-mail address, it is probably because the address is invalid. That happened for a while, so the committer is correct (because of the recent cherry-picking), but the author is not. I'll quickly do a mass rebase on that.

stevenpost:
Hi,

Some minor changes:
- rebased against the 1.3.0 tag
- dropped patch for the manage_service parameter (applied upstream, was a cherry-pick from master)

igalic:
@stevenpost you should always rebase against master

mhaskel:
@stevenpost @igalic is right, this needs to be rebased against master. And just a general comment to make the pull requests easier to review: we're a big fan of the [GitHub Flow](https://guides.github.com/introduction/flow/), so each of these commits should be on a separate branch/a separate pull request. All new features need to come with doc and test updates.

stevenpost:
Hi @mhaskel ,

I'll quickly cover you comments:
- documentation: agreed, it should be added
- github flow: I agree on that sentiment in general, but no here, I'm not pushing for inclusion, I'm making my work available so others
- rebase on master: normally yes, but again, I just want to run this on a production environment, it doesn't build upon master, it builds upon a specific tag. I'll leave the rebasing of that to you (should be fairly trivial).
- tests: while I am a strong proponent of testing, your test suite doesn't work with ruby 1.8.7, the version included on Red Hat 6 systems. I'm not writing tests I cannot execute. The module itself does work with this version. I believe the tests should mimic production as mush as possible (that also means the puppet version and the ruby version)

daenney:
> github flow: I agree on that sentiment in general, but no here, I'm not pushing for inclusion, I'm making my work available so others

If you want your work included you should follow the flow that the people needing to review and merge your changes need. It makes their life easier and your contribution more likely to land.

> tests: while I am a strong proponent of testing, your test suite doesn't work with ruby 1.8.7, the version included on Red Hat 6 systems. I'm not writing tests I cannot execute. The module itself does work with this version. I believe the tests should mimic production as mush as possible (that also means the puppet version and the ruby version)

The fact that you're on an outdated system is no reason not to write tests. You're not the only person using this module and your features and changes could have consequences for others.

> rebase on master: normally yes, but again, I just want to run this on a production environment, it doesn't build upon master, it builds upon a specific tag. I'll leave the rebasing of that to you (should be fairly trivial).

New features and changes are always introduced through the master branch. This is how it's always been done, I don't see any good reason to grant anyone an exception.

stevenpost:
Hi,

- rebased on current master
- fix mod_passenger support (the old patch wasn't enough)
- remove some temporary Debian thing for passenger as it broke Red Hat systems and was to be removed anyway (according to the comments)

This now actually runs in our production environment (based on 1.3.0).

On your previous comments:
- splitting the series: WIP
- tests: this module indicates it is supported and compatible on both RHEL 5 and 6 systems, if your test suite can't handle that, fix the suite, don't blame the other guy. Tests should be able to run on a production-like system. As I cannot run them, I don't write them.

daenney:
And why exactly can't you run the tests? You say that you can't, but you don't show any reason except for "I'm telling you it doesn't". The tests run on Ruby 1.8.7 just fine so that's not the problem.

stevenpost:
If they are do run fine, I suppose it is a documentation issue?

From the 'quick start' on the bottom of the readme:
$ bundle install
Fetching gem metadata from https://rubygems.org/........
Fetching version metadata from https://rubygems.org/..
Using rake 10.4.2
Using CFPropertyList 2.3.0

Gem::InstallError: i18n requires Ruby version >= 1.9.3.
An error occurred while installing i18n (0.7.0), and Bundler cannot continue.
Make sure that `gem install i18n -v '0.7.0'` succeeds before bundling.

That is on a RHEL 6 system.

mhaskel:
@stevenpost for the tests, you only need i18n for running acceptance tests, and your changes should only require updates to unit tests. You can install the bundle using `bundle install --without system_tests` and then run tests with `bundle exec rake spec`

cmurphy:
@stevenpost it looks like some of these things are fixed or you have put them up for review in separate pull requests, which makes things a lot easier for us. Can we close this pull request in favor of the separate smaller ones?

PR #1034, opened 2015-02-20
Add support for Passenger's PassengerAppEnv setting

liff:
(#1776) Add support for specifying PassengerAppEnv
either as a global setting to `apache::mod::passenger` or
per `apache::vhost`.


mhaskel:
This looks great, thanks @liff !

PR #1036, opened 2015-02-20
MODULES-1784 check for deprecated options and fail when they are unsupported

faker-:
See https://tickets.puppetlabs.com/browse/MODULES-1784 for more information.
Basically if you specify allow, deny, order or satisfy on a system running Apache 2.4, those options get silently dropped. 

With this pull request those manifests would fail and alert the user that they are not using supported options.

Feedback very much welcome.
I tried to solve this in the manifest but failed since $directories can be an array of hashes and I found no way to check for the existence of a key in such a construct.


mhaskel:
@faker- could you add some tests for this? thanks!

faker-:
@mhaskel added tests with last commit

mhaskel:
Great, thanks @faker- !

PR #1039, opened 2015-02-22
Fixed default mpm_event config warning

HT43-bqxFqB:
This should remove the warning and change the values to the defaults set by apache during the config check.
> AH00513: WARNING: MaxRequestWorkers of 256 is not an integer multiple of ThreadsPerChild of 25, decreasing to nearest multiple 250, for a maximum of 10 servers.


igalic:
thank you @HT43-bqxFqB!

PR #1040, opened 2015-02-24
Added basic content types to deflate with apache

leopoiroux:


manifests/mod/deflate.pp
```diff
@@ -1,9 +1,9 @@
 class apache::mod::deflate (
   $types = [
-    'text/html text/plain text/xml',
+    'text/html text/plain text/xml text/x-js',
```
igalic: does javascript work these days when it's deflated?

leopoiroux: For me text/x-js is like application/javascript, no ?

mhaskel:
I'm hesitant to change defaults unless there's a really good (documented) reason.

leopoiroux:
Hello, can you at least add "application/json" ? It seems logical to me.
For other "mimetype" I do not have a really good (documented) reason :)

mhaskel:
@igalic can you chime in here? I'm not finding any documentation indicating that 'application/json' would need to be a default, but I'm not familiar enough with mod_deflate to make a call here.

igalic:
from my experience (opinion?), application/json is not dangerous to compress, but i've had bad experiences compressing javascript. i blame browsers. :(

anyway, i don't think making this a default here is a good idea.
people who need this, should test it for themselves and see if it works well for them.

PR #1042, opened 2015-02-26
(#1391) Correct Debian jessie mod_prefork dev package name

sathieu:
This module calls apache2-prefork-dev as the package
name for the mod_prefork module for all Debian-flavored Linuxes. However
Ubuntu 14.04 changes this to apache2-dev and makes apache2-prefork-dev
an alias.

The result is that calling this resource, even with the package
installed, always results in a package changed purged to present
message, as it's scanning dpkg for apache2-prefork-dev, finding the
package missing, and then installing the alias via apt.

Relates to 25be287baf for the Ubuntu part


mhaskel:
Thanks @sathieu !

PR #1043, opened 2015-02-26
Fixed vhost proxy_pass params documentation

grafjo:
brace was missing


mhaskel:
Woo, good catch, thanks @grafjo !

grafjo:
you're welcome!

PR #1044, opened 2015-02-26
Corrected error in documentation for ssl_protocol and ssl_cipher.

tdiscuit:
The README.md file contains inaccurate information for the ssl_protocol setting. It has it listed as a string, but in reality an array is expected.

I have also fixed some contradiction on the ssl_protocol and ssl_cipher definitions.


mhaskel:
Thanks @tdiscuit !

PR #1047, opened 2015-02-26
add proper array support for require

mhaskel:
Fixes #1002


templates/vhost/_directories.erb
```diff
@@ -38,7 +38,9 @@
     <%- end -%>
     <%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
       <%- if directory['require'] and directory['require'] != '' -%>
-    Require <%= Array(directory['require']).join(' ') %>
+        <%- Array(directory['require']).each do |require| -%>
+    Require <%= require %>
```
hunner: Mmm, `@require`?

hunner: Oh no, local variable. Anyway, something other than a ruby keyword would be good.

PR #1048, opened 2015-03-03
Allow templates to be used for security.conf and security_crs.conf

MrAvenger:
I'm trying to override some of the default rules in security.conf but am currently unable to.


mhaskel:
Hi @MrAvenger , thanks for the contirbution! So, I'm thinking rather than allowing a template to be passed in, parameters should be added and the template should be modified for the rules you're trying to override.

This will also need [docs](https://github.com/puppetlabs/puppetlabs-apache#class-apachemodsecurity) and [test](https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/security_spec.rb) updates.

cmurphy:
@MrAvenger any update on this?

mhaskel:
Hi @MrAvenger, thanks for the contribution. We haven't heard from you in a while, so I'm going to close this PR for now. If this is functionality you're still interested in having added to the module please address the concerns from the comments and resubmit. Thanks again!

PR #1049, opened 2015-03-04
include mod_filter when needed instead of instantiating it

sethlyons:
mod_filter gets included in a few places, and when some of them include and other instantiate, we end up with duplicate resources.


mhaskel:
Thanks @sethlyons !

PR #1053, opened 2015-03-06
Adds the AddDefaultCharset option

stevenpost:
Please make sure to review the test, I'm not that familiar with rspec.


igalic:
it seems that travis is happy with your test!

igalic:
thank you @stevenpost!

stevenpost:
I know the test result is ok, but is the test actually testing what it should be testing?
Possible values are On, Off, <about any character set>.
That last one seems a bit more difficult to test.

PR #1054, opened 2015-03-06
Ensure resources notify Apache::Service class

butlern:
(#MODULES-1829) Fix resources that notify Service[httpd]

With the introduction of the service_manage bool, the Service[httpd]
class will not be part of the catalog if false. However there are still
some resources which notify Service[httpd] instead of the wrapper
Apache::Service class and these choke on undefined Service[httpd]
resource if service_manage is false.


igalic:
super! thank you @butlern!

PR #1056, opened 2015-03-09
no longer enforce the default value for $apache_name

fraenki:
The documentation states that the parameter `${apache::apache_name}` is available to install a different apache package. It is not working as expected, because the default values are still enforced in several places, ignoring the `apache_name` parameter.

This patch fixes this issue.

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -44,10 +44,10 @@ class apache::package (
           before => Package['httpd'],
         }
       }
-      $apache_package = $::apache::params::apache_name
+      $apache_package = $::apache::apache_name
     }
     default: {
-      $apache_package = $::apache::params::apache_name
+      $apache_package = $::apache::apache_name
     }
   }
 
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -121,7 +121,7 @@ define apache::vhost(
     fail('You must include the apache base class before using any apache defined resources')
   }
 
-  $apache_name = $::apache::params::apache_name
+  $apache_name = $::apache::apache_name
 
   validate_re($ensure, '^(present|absent)$',
   "${ensure} is not supported for ensure.


igalic:
@fraenki do you think you can fix the breaking spec test?

```
 1) apache::dev on a FreeBSD OS

Failure/Error: it { is_expected.to contain_class("apache::params") }

Puppet::Error:

Undefined variable "::apache::apache_name"; class ::apache has not been evaluated at /home/travis/build/puppetlabs/puppetlabs-apache/spec/fixtures/modules/apache/manifests/package.pp:47 on node testing-worker-linux-380b67c9-1-13901-linux-10-53647779
```

stevenpost:
Apache 2.4 support is one of the reasons I had to upgrade this module to a more recent version, so I'm very interested in fixing this as well.

A rebased version of this branch doesn't fail the tests on my test machine, but I can see why it mights fail (random ordering of resources by puppet). What would be a proper fix for this? Is it feasable to have package.pp inherit from ::apache instead of ::apache::params (params would still be included because ::apache inherits it)?

igalic:
in general, given that package is only ever called from apache (?) it should "inherit" apache, i.e.: it should be possible to access its variables.

stevenpost:
doing a quick grep, users are:
manifests/dev.pp: it says it requires the package, nothing more
manifests/mpm.pp: only for the freebsd osfamily, the class apache::package is declared as a resource
various tests.

Of course that is only inside of the Apache module.

I'll write a patch on top of the rebased branch and let the tests run, unless there are objections.

stevenpost:
Doing that gives me more errors on the tests:
rspec ./spec/classes/dev_spec.rb:56 # apache::dev on a FreeBSD OS
rspec ./spec/classes/dev_spec.rb:71 # apache::dev on a Gentoo OS
rspec ./spec/classes/mod/dev_spec.rb:24 # apache::mod::dev on a RedHat OS
rspec ./spec/classes/mod/dev_spec.rb:24 # apache::mod::dev on a Debian OS
rspec ./spec/classes/mod/dev_spec.rb:24 # apache::mod::dev on a FreeBSD OS

Looking at the actual test, it is the one that actually checks for ::apache::params.

thoughts?

stevenpost:
Well... this may be a problem...
https://docs.puppetlabs.com/puppet/latest/reference/lang_classes.html#inheritance
#####
Note: This version of Puppet does not support using parameterized classes for inheritable base classes. The base class must have no parameters.
######

stevenpost:
I think I found the error, it seems to be in the test itself, rather that the manifest code.
I'll let my current test run finish and create a pull request.
Missing preconditions.

fraenki:
Close this in favour of puppetlabs/puppetlabs-apache/pull/1070. Thanks @stevenpost.

PR #1057, opened 2015-03-09
make $lib_path configurable

fraenki:
In conjunction with `${apache::apache_version}` it is sometimes required to also change `$lib_path` to a non-default location. This patch introduces `lib_path` as a new parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,10 @@ Sets the amount of time the server waits for subsequent requests on a persistent
 
 Sets the limit of the number of requests allowed per connection when KeepAlive is on. Defaults to '100'.
 
+#####`lib_path`
+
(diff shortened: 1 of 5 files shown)


igalic:
are those uses in mod_wsgi and mpm the only ones?
should we generally make this a template for modules?

bmjen:
Please add a test for the new param.

cmurphy:
I don't think there's an especially obvious place to add a test for this, and I'd be fine with going without. I would still like to see @igalic's questions answered.

mhaskel:
@fraenki ping?

fraenki:
@igalic These are the only occurences of `::apache::params::lib_path`. I don't think there's more to do here, since this is a somewhat special use-case.
@mhaskel Sorry for my late reply. :-(

PR #1059, opened 2015-03-11
Add support for the proxy-nokeepalive setting

stevenpost:
As the title says.
I added the option into the 'all options' test for vhosts, but I don't see where to check it properly.


igalic:
i'm a bit confused about this. i think it needs better documentation.

stevenpost:
My comment was about the tests, not the actual change.
I'll see if I can come up with some better description for the README.md file, actual documentation from Apache here: http://httpd.apache.org/docs/2.2/mod/mod_proxy.html#envsettings
A typical use case here: https://wiki.jenkins-ci.org/display/JENKINS/Running+Jenkins+behind+Apache

stevenpost:
Thinking about this, perhaps there should be a more generic way to set environment variables with SetEnv inside the proxy parts?

stevenpost:
I'm closing this pull request as it is superseeded by https://github.com/puppetlabs/puppetlabs-apache/pull/1061

PR #1060, opened 2015-03-11
give a lower priority to mod_passenger

stevenpost:
By giving mod_passenger a lower priority, mod_proxy will be loaded before mod_passenger, allowing a single vhost to direct certain requests using mod_proxy and direct the others to mod_passenger.

Use case: puppetmaster where the CA requests are proxied to a single server (without the need to configure the agents for different servers).

### DIFF
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -15,9 +15,9 @@ class apache::mod::passenger (
   $passenger_app_env              = undef,
   $mod_package                    = undef,
   $mod_package_ensure             = undef,
-  $mod_lib                        = undef,
+  $mod_lib                        = 'mod_passenger.so',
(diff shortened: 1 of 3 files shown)


igalic:
yes! having proxy before application server modules is a  wise choice.

@stevenpost, can you please [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) those 4 commits down to one?

stevenpost:
*squashed*

This is actually used in our environment, hardware load balancer with 2 puppetmasters, 1 of them is CA.

igalic:
thanks @stevenpost!

mhaskel:
This broke acceptance testing. Reimplemented in #1072.

PR #1061, opened 2015-03-12
Allow setting environment variables inside the proxy locations

stevenpost:
This patch adds support to set variables using 'SetEnv' inside proxy
locations. This is needed to support some backends that don't properly
support HTTP/1.1, as it allows a fallback by the Apache httpd server.
For details see [1] (Apache httpd documentation) and [2] (use case).

[1] http://httpd.apache.org/docs/2.2/mod/mod_proxy.html#envsettings
[2] https://wiki.jenkins-ci.org/display/JENKINS/Running+Jenkins+behind+Apache

### DIFF
--- a/README.md
+++ b/README.md
@@ -1335,12 +1335,15 @@ apache::vhost { 'site.name.fdqn':
       'params' => { 'retry' => '0', 'timeout' => '5' }, },
     { 'path' => '/e', 'url' => 'http://backend-a/e',
       'keywords' => ['nocanon', 'interpolate'] },
+    { 'path' => '/f', 'url' => 'http://backend-f/',
+      'setenv' => ['proxy-nokeepalive 1','force-proxy-request-1.0 1']},
   ],
 }
 ```
 
 `reverse_urls` is optional and can be an array or a string. It is useful when used with `mod_proxy_balancer`.
 `params` is an optional parameter. It allows to provide the ProxyPass key=value parameters (Connection settings).
+`setenv` is optional and is an array to set environment variables for the proxy directive, for details see http://httpd.apache.org/docs/current/mod/mod_proxy.html#envsettings
 
 #####`rack_base_uris`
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1335,12 +1335,14 @@ apache::vhost { 'site.name.fdqn':
       'params' => { 'retry' => '0', 'timeout' => '5' }, },
     { 'path' => '/e', 'url' => 'http://backend-a/e',
       'keywords' => ['nocanon', 'interpolate'] },
+    { 'path' => '/f', 'url' => 'http://backend-f/', 'setenv' => ['proxy-nokeepalive 1','force-proxy-request-1.0 1']},
```
igalic: please insert a break here for readability

README.md
```diff
@@ -1335,12 +1335,14 @@ apache::vhost { 'site.name.fdqn':
...
   ],
 }
 ```
 
 `reverse_urls` is optional and can be an array or a string. It is useful when used with `mod_proxy_balancer`.
 `params` is an optional parameter. It allows to provide the ProxyPass key=value parameters (Connection settings).
+`setenv` is optional and is an array to set environment variables for the proxy directive, for details see http://httpd.apache.org/docs/2.2/mod/mod_proxy.html#envsettings
```
igalic: please use /current/ for links to the documentation.

igalic:
looks much betterer now!
thanks for the update.

stevenpost:
*squashed the commits into a single patch*

stevenpost:
Hi,

Do you know why this test fails on travis, only for some environments? https://travis-ci.org/puppetlabs/puppetlabs-apache/builds/54071349

The build is from before the squash, but no actual changes are made with the squash.

stevenpost:
nevermind, I see it... the other environments are just taking longer, I made an error in one of the tests, will push an update.

PR #1062, opened 2015-03-12
Remove Debian workaround as it broke Red Hat systems

stevenpost:
On Red Hat systems it was entirely possible to have the path for 'passenger_package.conf' file equal to that of the 'passenger.conf' file, causing an error. This was only the case when the '$passenger_conf_package_file' variable was not set (the default behaviour).

It was due to be removed anyway.


igalic:
@stevenpost can you add, or fix? the acceptance tests, should it exist

stevenpost:
Only the rspec test I removed in the patch seems to check for the workaround.
The acceptance tests don't seem to check for the filenames, although I cannot run them.

PR #1063, opened 2015-03-12
add parameters to configure expires globally

igalic:
allow global configuration of all of mod_expires Parameters.

### DIFF
--- a/templates/mod/expires.conf.erb
+++ b/templates/mod/expires.conf.erb
@@ -1,0 +1,11 @@
+ExpiresActive <%= scope.function_bool2httpd([@expires_active]) %>
+<%- if ! @expires_default.nil? and ! @expires_default.empty? -%>
+ExpiresDefault "<%= @expires_default %>"
+<%- end -%>
+<%- if ! @expires_by_type.nil? and ! @expires_by_type.empty? -%>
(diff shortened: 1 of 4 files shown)


templates/mod/expires.conf.erb
```diff
@@ -0,0 +1,11 @@
+ExpiresActive <%= scope.function_bool2httpd([@expires_active]) %>
+<%- if ! @expires_default.nil? and ! @expires_default.empty? -%>
+ExpiresDefault "<%= @expires_default %>"
+<%- end -%>
+<%- if ! @expires_by_type.nil? and ! @expires_by_type.empty? -%>
+<%-   [@expires_by_type].flatten.each do |line| -%>
```
igalic: if we force this to [array], and flatten it, we don't need to check if it's nil/empty.

igalic: yeah, except, then `line.map` will fail, if `expires_by_type` is `undef`

igalic:
:+1: in this great monologue.

igalic:
anyone wanna merge my monologue?
@mhaskell, @cmurphy?

igalic:
I'm already using it in production, btw

PR #1064, opened 2015-03-13
Apache: allow setting the default type

stevenpost:
This setting is only valid on Apache httpd 2.2 (and possibly lower)

Please make sure to review the test extensively, I may have done something wrong about the position I placed that test.


igalic:
thank you @stevenpost!

PR #1065, opened 2015-03-13
Remoteip module

igalic:
Add mod_remoteip module, fixing some issues on #989 that @mhaskel pointed out


mhaskel:
@igalic tests fail.

igalic:
yeah, and i broke them ~_~

PR #1066, opened 2015-03-16
Apache: add module for mod_geoip

stevenpost:
Only actually tested on RHEL 6.

Debian *should* be ok according to https://packages.debian.org/wheezy/all/geoip-database/filelist (this package is recommended by a direct dependency of the Apache module, so most systems install it automatically with the Apache module)

Edit: typo

### DIFF
--- a/templates/mod/geoip.conf.erb
+++ b/templates/mod/geoip.conf.erb
@@ -1,0 +1,3 @@
+GeoIPEnable <%= scope.function_bool2httpd([@enabled]) %>
+
+GeoIPDBFile <%= @db_file %>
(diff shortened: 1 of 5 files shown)


templates/mod/geoip.conf.erb
```diff
@@ -0,0 +1,5 @@
+<%- if @enabled -%>
+GeoIPEnable On
```
igalic: you could use the same as here https://github.com/igalic/puppetlabs-apache/blob/expires/templates/mod/expires.conf.erb#L1
```ruby
GeoIPEnable <%= scope.function_bool2httpd([@enabled]) %>
```

cmurphy:
@stevenpost could you rebase this against master?

The tests seem to have failed due just from running for too long, rebasing will kick off another try.

stevenpost:
No, this functionality has been merged to master in 0cba3904c5ba02f79dea17750bd71cf73cc6c068, thus causing a merge conflict.
This was a different pull request.

PR #1067, opened 2015-03-16
Apache - mod_passenger: allow setting PassengerMinInstances

stevenpost:


igalic:
i just realised: we don't allow to enable/disable passenger globally…
we should change that

igalic:
@stevenpost maybe we should validate that these things are numbers, might be more sensible errors than the ones we would otherwise get from puppet trying and failing to reload httpd… just a thought, might not be really necessary… or out of scope for this pr

stevenpost:
Definitely out of scope for this pr, but I tend to agree on that.
On the other hand, do you want this module to be a replacement for apachectl configtest?
If you start validating everything (combinations, strings, etc.) things may go out of hand quickly.

Perhaps it is better to be able to call 'apachectl configtest' as a dependency on the service restart (exec with refresh_only => true), but making it optional. That way, a running httpd doesn't get shut down when misconfiguring something.

igalic:
yeah, the only problem with that is that we'll need to depend on a newer version of concat, which means a major version bump.
so maybe this is for https://github.com/puppetlabs/puppetlabs-apache/blob/2.0.x/DESIGN.md ?

stevenpost:
I fail to see how the concat module has anything to do with this.

igalic:
https://github.com/puppetlabs/puppetlabs-concat/#validate_cmd

https://github.com/puppetlabs/puppetlabs-concat/pull/263

which has only been supported since [1.2.0](https://github.com/puppetlabs/puppetlabs-concat/tree/1.2.0) (compare the [previous release](https://github.com/puppetlabs/puppetlabs-concat/tree/1.1.2))

this module [depends on >=1.1.1](https://github.com/puppetlabs/puppetlabs-apache/blob/master/metadata.json#L71)

changing dependencies means bumping the major version, according to semver.

PR #1069, opened 2015-03-16
Apache: add support for the ProxyPassMatch directive

stevenpost:
This would require more tests, although due to how the tests for the normal ProxyPass directive are written, they match ProxyPassMatch as well, making it a bit more difficult.

### DIFF
--- a/templates/vhost/_proxy.erb
+++ b/templates/vhost/_proxy.erb
@@ -32,6 +32,29 @@
   <%- end -%>
   </Location>
 <% end -%>
+<% [@proxy_pass_match].flatten.compact.each do |proxy| %>
+  ProxyPassMatch <%= proxy['path'] %> <%= proxy['url'] %>
+  <%- if proxy['params'] -%>
+    <%- proxy['params'].each_pair do |key, value| -%> <%= key %>=<%= value -%>
+    <%- end -%>
+  <%- end -%>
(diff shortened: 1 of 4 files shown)


templates/vhost/_proxy.erb
```diff
@@ -32,6 +32,29 @@
...
+    ProxyPassReverse <%= reverse_url %>
+    <%- end -%>
+  <%- end -%>
+  <%- if proxy['setenv'] -%>
+    <%- Array(proxy['setenv']).each do |setenv_var| -%>
+    SetEnv <%= setenv_var -%>
+    <%- end -%>
+  <%- end -%>
+  </Location>
+<% end -%>
```
igalic: i think my brain just melted.

:(

stevenpost: Well... The original patch was really simple, only 3 lines instead of that whole block, but since functionality is the same as 'ProxyPass', I copied the whole ProxyPass block (and changing the variable name). The original patch was written for version 0.x of this module.

igalic: i think this whole mess must go… it was one of my first contributions to this module, and now that i looke at it, it just makes no sense at all :(

igalic: bug again, that's out of scope of this pr…

PR #1072, opened 2015-03-16
Give a lower priority to mod_passenger

mhaskel:
This was initially implemented in #1060.

### DIFF
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -15,9 +15,9 @@ class apache::mod::passenger (
   $passenger_app_env              = undef,
   $mod_package                    = undef,
   $mod_package_ensure             = undef,
-  $mod_lib                        = 'mod_passenger.so',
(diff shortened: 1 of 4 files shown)


manifests/mod/passenger.pp
```diff
@@ -56,13 +56,14 @@
...
   $_path = $mod_path
-  ::apache::mod { 'zpassenger':
+  ::apache::mod { 'passenger':
     package        => $_package,
     package_ensure => $_package_ensure,
     lib            => $_lib,
     lib_path       => $_lib_path,
     id             => $_id,
     path           => $_path,
+    loadfile_name  => 'zpassenger.load',
```
igalic: oh, yeah, that makes things easier…

stevenpost: cool, I fully agree.
In my defense, that parameter didn't exist when I first wrote that patch :)

PR #1074, opened 2015-03-16
1.4.0 prep

mhaskel:


mhaskel:
This is reverting #1001 for the 1.4.0 release as that was a backwards incompatible change.

stevenpost:
Would be nice if #1070 and #1066 could make it into this release.
For 1066 some tests timed out, others succeeded. For 1070, it *should* be fixed now but the Travis job is still running.

Any chance these could still make it?

mhaskel:
@stevenpost sorry, I think those are going to need to wait until the next release. It's a bit late to pull in new features.

PR #1080, opened 2015-03-19
Added support for TraceEnable directive in vhosts.pp manifest

bschonec:
Since TraceEnable directive can be put in either the Apache server config or virtual host config, it makes sense to be able to have Puppet manage the directive in either place.


bschonec:
Added ability to override Apache main configuration file (typically httpd.conf) variable $conf_file.  This give the user the ability to manage multiple configuration files on a single servers.  This is advantages if the server is running both PHP4 and PHP5.  An example would be to have a 'httpd.conf.php4' and 'httpd.conf.php5' to distinguish between the two concurrent installs of PHP.

manifests/mod/php4.pp
```diff
@@ -0,0 +1,62 @@
...
+  validate_array($extensions)
+
+  if $source and ($content or $template != 'apache/mod/php4.conf.erb') {
+    warning('source and content or template parameters are provided. source parameter will be used')
+  } elsif $content and $template != 'apache/mod/php4.conf.erb' {
+    warning('content and template parameters are provided. content parameter will be used')
+  }
+
+  $manage_content = $source ? {
+    undef   => $content ? {
```
bmjen: Needs Alignment

manifests/init.pp
```diff
@@ -66,10 +68,12 @@
   $apache_version         = $::apache::version::default,
   $server_tokens          = 'OS',
   $server_signature       = 'On',
-  $trace_enable           = 'On',
+  $trace_enable           = undef,
```
cyberious: Why are we changing the default behavior, this is a breaking change.

manifests/init.pp
```diff
@@ -244,11 +263,12 @@
...
   }
 
-  if $::apache::conf_dir and $::apache::params::conf_file {
+if $::apache::manage_conf_file {
+  if $::apache::conf_dir and $::apache::conf_file {
     case $::osfamily {
       'debian': {
         $pidfile              = "\${APACHE_PID_FILE}"
-        $error_log            = 'error.log'
+        $error_log            = 'error_log'
```
cyberious: Another breaking change. Why?

manifests/mod/php4.pp
```diff
@@ -0,0 +1,62 @@
+class apache::mod::php4 (
```
cyberious: This is a good idea, but should be a separate PR

templates/mod/ssl.conf.erb
```diff
@@ -9,17 +9,17 @@
 
   SSLPassPhraseDialog <%= @ssl_pass_phrase_dialog %>
   SSLSessionCache "shmcb:<%= @session_cache %>"
-  SSLSessionCacheTimeout <%= @ssl_sessioncachetimeout %>
+  SSLSessionCacheTimeout 300
```
tphoney: why are you hardcoding these values, they were previously variables

checksums.json
```diff
@@ -0,0 +1,289 @@
+{
```
jonnytpuppet: This file should not be commited

manifests/init.pp
jonnytpuppet: Would this cause a backwards incompatibility?

bschonec:
I'm quite new at all of this and I apologize for the 'noise' in the code.  I'll pull the latest version and then clean up and resubmit.

igalic:
@bschonec: no worries about noise!

PR #1083, opened 2015-03-22
MODULES-1789 add initial mod_geoip support

roman-mueller:
This adds support for mod_geoip.
I've tested it on CentOS 6 and 7.

There is one problem with this: if you want to set `GeoIPEnable` in a `directories` parameter you must declare the class `apache::mod::geoip` yourself.  
As the `directories` parameter can be an array of hashes I cannot check for the value of `geoip_enable` here:  
```puppet
    apache::vhost { 'first.example.com':
      docroot     => '/var/www/first',
      directories => [
        { path         => '/var/www/first',
          geoip_enable => 'On',
        },
      ],
    }
```

And because of this, I cannot automatically include it.

### DIFF
--- a/manifests/mod/geoip.pp
+++ b/manifests/mod/geoip.pp
@@ -1,22 +1,22 @@
 class apache::mod::geoip (
-  $enable                  = 'Off',
-  $dbfile                  = '/usr/share/GeoIP/GeoIP.dat',
-  $flag                    = 'Standard',
-  $output                  = 'All',
-  $enableutf8              = undef,
-  $scanproxyheaders        = undef,
-  $uselastxforwarededforip = undef,
+  $enable                     = false,
+  $db_file                    = '/usr/share/GeoIP/GeoIP.dat',
+  $flag                       = 'Standard',
+  $output                     = 'All',
+  $enable_utf8                = undef,
+  $scan_proxy_headers         = undef,
+  $use_last_xforwarededfor_ip = undef,
 ) {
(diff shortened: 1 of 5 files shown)


manifests/mod/geoip.pp
```diff
@@ -0,0 +1,29 @@
+class apache::mod::geoip (
+  $enable                  = 'Off',
+  $dbfile                  = '/usr/share/GeoIP/GeoIP.dat',
+  $flag                    = 'Standard',
+  $output                  = 'All',
+  $enableutf8              = undef,
+  $scanproxyheaders        = undef,
+  $uselastxforwarededforip = undef,
```
igalic: i know it'll make the variables longer, but can you please under_score them?

manifests/mod/geoip.pp
```diff
@@ -0,0 +1,29 @@
+class apache::mod::geoip (
+  $enable                  = 'Off',
```
igalic: i'd rather this was a boolean, then we can just use `bool2httpd()` in the template

roman-mueller:
@igalic thanks for feedback, didn't know about bool2httpd until now.
Can you check my last commit?

I'm not sure why the build failed now, it says it exceeded 50 minutes.
That sounds like Travis CI is overloaded rather than my change breaking it.

igalic:
it looks like travis is overloaded, i could restart the jobs and see if that helps…

igalic:
thanks @roman-mueller!

PR #1084, opened 2015-03-23
Test and future parser updates

cmurphy:


.travis.yml
```diff
@@ -1,21 +1,28 @@
 ---
+sudo: false
```
igalic: \o/ finally

manifests/dev.pp
```diff
@@ -4,8 +4,10 @@
   }
   include ::apache::params
   $packages = $::apache::params::dev_packages
-  package { $packages:
-    ensure  => present,
-    require => Package['httpd'],
+  if $packages { # FreeBSD doesn't have dev packages to install
+    package { $packages:
```
igalic: i used this as horrible, horrible hack in one of my modules…

spec/classes/apache_spec.rb
```diff
@@ -794,7 +794,7 @@
 
     it do
       expect {
-       should compile
+       catalogue
```
igalic: where's that from?

mhaskel: undocumented in rspec-puppet? @hunner found it.

igalic: how is the undocumented thing better?

PR #1086, opened 2015-03-25
Fixed setting multiple env_var in a location block for proxy pass.

btreecat:
This fix should allow people using multiple env_var because each var being set needs to be on it's own line.


jonnytpuppet:
Hi There

Please submit an acceptance test with a manifest using multiple env_var

Thanks!

cmurphy:
I think this is a minor enough change to merge without new acceptance tests.

btreecat:
Thanks! I'll be sure to include tests next time if the changes are less minor.

PR #1089, opened 2015-03-25
log_level ^ and $ removed so that log_levels like info ssl:debug are allowed

mrgum:
I had the need to debug ssl in apache for just one vhost

So I copy and pasted from init.pp

I was not able to run the test suite, I'll try harder but please consider nevertheless

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -186,7 +186,8 @@ define apache::vhost(
   Allowed values are 'directory' and 'absent'.")
 
   if $log_level {
-    validate_re($log_level, '^(emerg|alert|crit|error|warn|notice|info|debug)$',
+    $valid_log_level_re = '(emerg|alert|crit|error|warn|notice|info|debug)'
+    validate_re($log_level, $valid_log_level_re,
     "Log level '${log_level}' is not one of the supported Apache HTTP Server log levels.")
   }
 


jonnytpuppet:
Hi There

Thank you for your contribution!

We would need a basic acceptance test for this to be accepted. Please look here for current tests and simply add one with a manifest with a log level that did not pass before your PR:

https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/acceptance/vhost_spec.rb

Thanks!

cmurphy:
@mrgum Looking at http://httpd.apache.org/docs/2.4/mod/core.html#loglevel it looks like this type of syntax is meant to be able to support multiple log levels, as it's setting it per module. If we allow a regex like 'ssl:debug' then it will only set the log level for the ssl mod. I think this change is an incomplete solution, and I would rather change the parameter to support an array of log levels like that.

mrgum:
@cmurphy my proposal relaxes the regex to ensure that the string merely has one of those values in it. the code is copied from the parent init.pp lines 127-130
i'm just making the vhost define loglevel match the apache class loglevel

I'd fear by making it an array we'd have to track changes to apache closely

cmurphy:
According to the docs, "Specifying a level without a module name will reset the level for all modules to that level. Specifying a level with a module name will set the level for that module only." So what happens if I set `log_level => 'ssl:debug'` but I want the log level for a different module, or for everything else, to be `info`?

I'm leaning toward merging this since you pointed out that init.pp already does this, but I think it should be fixed in both places.

@igalic @mhaskel @hunner want to weigh in?

mrgum:
The check just says "does the string have one of the valid levels in it" it doesn't mind how many of them there are or if there are different ones. I currently use this patch with the setting 'info ssl:debug'

whilst I agree an array might better model the concept, I would dislike it as I prefer just having a single string in my hiera and I can not see that I'd

mrgum:
Its probably a bad idea to have the same code for checking the values in two places and LogLevel is also valid in directory settings from 2.3.6 so that would make three.

maybe it would be better to have a validate_apache_loglevel check function and use that in the init.pp and twice in vhost.pp (once for the main vhost and once for the directories)

mrgum:
btw the regexp needs changing to support the trace1-trace8 levels

igalic:
@mrgum do you wanna pick that up also (re trace-levels)

mrgum:
I have redo as a new parser function supporting trace levels and having tests

I'll close this and start a new request

PR #1094, opened 2015-04-01
Allow to fully remove a vhost with special configurations

cjeanneret:
This commit allows apache::vhost to fully take knowledge of the vhost
status.
There were two places where the "ensure => absent" caused some failure,
puppet wanting to create some subdirectory in the vhost when it was
removed.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -251,8 +251,12 @@ define apache::vhost(
   # This ensures that the docroot exists
   # But enables it to be specified across multiple vhost resources
   if ! defined(File[$docroot]) and $manage_docroot {
+    $docroot_ensure = $ensure? {
+      present => 'directory',
+      default => 'absent',
+    }
     file { $docroot:
-      ensure  => directory,
+      ensure  => $docroot_ensure,
       owner   => $docroot_owner,
       group   => $docroot_group,
       mode    => $docroot_mode,
@@ -263,8 +267,12 @@ define apache::vhost(
 
   # Same as above, but for logroot
   if ! defined(File[$logroot]) {
+    $logroot_ensure_real = $ensure? {
+      present => $logroot_ensure,
+      default => 'absent',
+    }
     file { $logroot:
-      ensure  => $logroot_ensure,
+      ensure  => $logroot_ensure_real,
       mode    => $logroot_mode,
       require => Package['httpd'],
       before  => Concat["${priority_real}${filename}.conf"],
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -453,7 +453,10 @@ describe 'apache::vhost', :type => :define do
       it { is_expected.to_not contain_class('apache::mod::proxy_http') }
       it { is_expected.to_not contain_class('apache::mod::passenger') }
       it { is_expected.to_not contain_class('apache::mod::headers') }
-      it { is_expected.to contain_file('/var/www/foo') }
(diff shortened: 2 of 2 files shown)


igalic:
this needs documentation. a proper warning that says this will destroy your data (if you want;)

manifests/vhost.pp
```diff
@@ -263,8 +267,12 @@
 
   # Same as above, but for logroot
   if ! defined(File[$logroot]) {
+    $logroot_ensure_real = $ensure? {
+      present => $logroot_ensure,
+      default => 'absent',
+    }
     file { $logroot:
-      ensure  => $logroot_ensure,
+      ensure  => $logroot_ensure_real,
```
mhaskel: I don't like this. The reason we added a `$logroot_ensure` parameter rather than just using `$ensure` is we didn't want to remove logs when a vhost is being removed.

manifests/vhost.pp
```diff
@@ -251,8 +251,12 @@
   # This ensures that the docroot exists
   # But enables it to be specified across multiple vhost resources
   if ! defined(File[$docroot]) and $manage_docroot {
+    $docroot_ensure = $ensure? {
+      present => 'directory',
+      default => 'absent',
+    }
```
mhaskel: Rather than basing this on ensure could you add a `docroot_ensure` parameter to the define?

cjeanneret: hmm, we might expect that "ensure => absent" removes all vhost resources, don't we?
I guess (not sure if possible) ```docroot_ensure``` default value set to ```$ensure``` might be for the best. Will give it a try.

manifests/vhost.pp
cjeanneret: same thing here as the ```docroot_ensure``` comment. defaulting to ```$ensure``` instead of the current ```'directory'``` might correct it in a smooth, convenient way.

cjeanneret:
Seeing comments, I'll add some doc in order to warn people, and try to default values to ```$ensure``` value instead of fixed one.

I'll do a second commit in order to do so. If this suits you, I'll then cleanup the commits in order to provide only one, so that you have a clean tree.

manifests/vhost.pp
mhaskel: I'm not really comfortable changing the default behavior for logroot, given the possible security implications of the change (security folks I have worked with were very particular about not having logs removed, and this could accidentally break that. I think it's best to require people to be explicit)

manifests/vhost.pp
mhaskel: The issue with defaulting to have the `$docroot_ensure` value default to `$ensure` is that the module supports multiple vhosts in the same docroot, so you could end up accidentally blowing away other things. I think I like this being added as a parameter with a default of `directory`, and then the user just needs to be explicit if they want to have the docroot removed.

tphoney:
pint @cjeanneret, thanks for being willing to make the changes.

jonnytpuppet:
ping @cjeanneret is there any update on this? Thanks!

cjeanneret:
Hello,

Yes, sorry ­— had a lot of things to do, I hope I can submit something either today, or next week. Sorry for the delay :(

tphoney:
ping @cjeanneret is there any update on this? Thanks!

mhaskel:
Hi @cjeanneret, thanks for the contribution! We're going to close this for now due to inactivity, but if this is something you're still interested in having added to the module please address the comments and resubmit. Thanks again!

PR #1096, opened 2015-04-09
Do not offload overriding LogFormats to httpd

igalic:
if an admin overrides one of the pre-existing formats, do not offload
that to httpd: check for it, and replace it at configuration
generation time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,7 +357,7 @@ Specifies the location where apache module files are stored. It should not be co
 
 #####`loadfile_name`
 
-Sets the file name for the module loadfile. Should be in the format *.load.  This can be used to set the module load order.
+Sets the file name for the module loadfile. Should be in the format \*.load.  This can be used to set the module load order.
 
 #####`log_level`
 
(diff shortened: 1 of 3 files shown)


igalic:
this *might* needs tests or documentation.

mhaskel:
@igalic thanks! Yeah, this definitely needs tests and docs. Do you think this would break existing configurations (i.e., is this backwards incompatible?)

igalic:
nope. shouldn't be.

igalic:
once again, i have no idea how to test for this, la la la…
really, i don't know how tests work. i'll just wait for @DavidS to fix rspec-puppet ;P

DavidS:
You need this rspec-puppet PR (https://github.com/rodjek/rspec-puppet/pull/292) and then you can use this pattern (https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/type_spec.rb#L10-L13) without the puppetlabs_spec_helper mess.

mhaskel:
thanks @igalic

PR #1097, opened 2015-04-09
unify log_level checking and allow trace1-8 levels

mrgum:
this request came out of discussion on "log_level ^ and $ removed so that log_levels like info ssl:debug are allowed #1089"

added validate_apache_log_level function to apply same test to both apache class and vhost - trace levels also supported

still todo is think about supporting log_level checking in directories


igalic:
let's merge this as it is now and see what breaks :0

PR #1099, opened 2015-04-14
Fix typo in mod_expires docs

mlandewers:


manifests/mod/expires.pp
```diff
@@ -6,9 +6,9 @@
   ::apache::mod { 'expires': }
 
   # Template uses
-  # $expries_active
-  # $expries_default
-  # $expries_by_type
+  # $expires_active
+  # $expires_default
+  # $expires_by_type
```
igalic: i'm dyslexic ._. thank you for catching that.

PR #1103, opened 2015-04-16
MODULES-1937: fixes mismatching of Apache version on Amazon Linux

BurritoAlPastor:
Fixes https://tickets.puppetlabs.com/browse/MODULES-1937 by identifying Amazon Linux before checking any version numbers.

I'm afraid I'm not able to add tests for this.

### DIFF
--- a/manifests/version.pp
+++ b/manifests/version.pp
@@ -3,7 +3,7 @@
 # Try to automatically detect the version by OS
 #
 class apache::version {
-  # This will be 5 or 6 on RedHat, 6 or wheezy on Debian, 12 or quantal on Ubuntu, 3 on Amazon, etc.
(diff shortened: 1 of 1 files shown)


manifests/version.pp
```diff
@@ -12,7 +12,9 @@
 
   case $::osfamily {
     'RedHat': {
-      if ($::operatingsystem == 'Fedora' and versioncmp($distrelease, '18') >= 0) or ($::operatingsystem != 'Fedora' and versioncmp($distrelease, '7') >= 0) {
+      if ($::operatingsystem == 'Amazon') {
```
igalic: how come this isn't caught by the else?

BurritoAlPastor: It... should be? I can't reproduce a situation in which it doesn't.

tphoney:
Can you please fix the failing acceptance tests.

PR #1104, opened 2015-04-17
Use concat 1.2.x

mhaskel:


igalic:
:(
when can we switch?

Jeoffrey54:
Same issue for me :(.

mhaskel:
switching to concat 2.x is a backwards-incompatible change. I'm guessing at some point we'll have releases of everything dependent on the newer concat, but as that hasn't been released (and hasn't fully stabilized yet) we need to pin our dependencies so things don't explode.

domcleal:
It seems that 2.x ended up being pretty much compatible (which makes me very happy), so I've opened #1126 to relax the deps and fix deprecations.

PR #1107, opened 2015-04-23
(#1971) new $service_restart parameter to influence httpd.

traylenator:
Adds a new parameter to apache and apache::service to
permit restart paramter to the httpd service to be set.

For instance this allows to a restart to be configured
as apachectl graceful for instance.

Fixes https://tickets.puppetlabs.com/browse/MODULES-1971

the change is fully backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,6 +456,11 @@ Name of the Apache service to run. Defaults to: 'httpd' on RedHat, 'apache2' on 
 
 Determines whether the HTTPD service state is managed by Puppet . Defaults to 'true'.
 
+#####`service_restart`
+
+Determines whether the HTTPD service restart command should be anything other than the default managed by Puppet.  Defaults to undef.
+
+
 #####`trace_enable`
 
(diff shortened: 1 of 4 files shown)


bmjen:
@traylenator Thanks for the well put together PR! This will be a welcomed param addition.

tenajsystems:
@traylenator,  how do i go about setting the service_restart parameter?  I have tried setting `service_restart` like the code below but when i run `puppet agent -t` , no changes are made
```
class { 'apache':
  service_restart => '/usr/sbin/apachectl -k graceful',
}
```
Any thoughts on why the code I have above doesn't work?

traylenator:
There has to be a configuration chnage somewhere?

tenajsystems:
I thought it's supposed to make configuration changes but upon taking a second look, it appears it's not supposed to make any configuration changes and setting `service_restart => '/usr/sbin/apachectl -k graceful'` works as intended. Thanks!

PR #1112, opened 2015-04-30
fix puppet-lint errors and warnings

costela:
Minor commit to shut-up puppet-lint

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -339,11 +339,11 @@ define apache::vhost(
     }
   }
   if $add_listen {
-    if $ip and defined(Apache::Listen["${port}"]) {
+    if $ip and defined(Apache::Listen[$port]) {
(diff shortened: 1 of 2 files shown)


igalic:
i'm fascinated by now *everything* is failing on the future parser.

mhaskel:
Hi @costela, thanks for the contribution. The quotes in `manifests/vhost.pp` are required for future parser, so the correct way to fix that would be to add a `# lint:ignore:only_variable_string` comment on those lines.

costela:
Hi @mhaskel, 
I'm not sure I understand it: I *am* using the future parser (albeit on 3.7) and it doesn't seem to choke on the lack of quotes. Is using quotes in this case the "new right thing"? (maybe I'm using an outdated version of puppet-lint; can't check right now)

mhaskel:
Hi @costela, the issue is that the `$port` and `$listen_addr_port` support either a stringified integer or an integer, and the future parser is not ok with resources with a numeric title (so, like `apache::listen { 80: }`)

costela:
aaah, ok, gotcha! now it makes sense! ;)
thanks @mhaskel! I'll close this PR then

PR #1113, opened 2015-04-30
MODULES-1968 - Update the template to warn if using deprecated options

mhaskel:


roman-mueller:
Why were the tests removed and not changed to check for the warning?
Ugh, I spent 90% of the initial PR's time on figuring out how to add those tests and now they get ripped out.

igalic:
@mhaskel do you wanna restore the tests?

mhaskel:
@roman-mueller AFAIK rspec-puppet can't actually detect warnings :(

Aethylred:
@roman-mueller @mhaskel rspec-puppet can't detect warnings, but it can detect notify

igalic:
joy. ^ @DavidS ping

DavidS:
You need this rspec-puppet PR (https://github.com/rodjek/rspec-puppet/pull/292) and then you can use this pattern (https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/type_spec.rb#L10-L13) without the puppetlabs_spec_helper mess.

PR #1119, opened 2015-05-08
Fix additional includes in vhost

ckaenzig:
The parameter `use_optional_includes` was recently added to the module, but it was not working in configuration generated by `apache::vhost`. This PR means to fix that.

Should both commits be squashed or does it make sense to keep them separate?

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -105,6 +105,7 @@ define apache::vhost(
   $fastcgi_socket              = undef,
   $fastcgi_dir                 = undef,
   $additional_includes         = [],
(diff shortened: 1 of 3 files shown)


igalic:
@ckaenzig sorry for getting back to you this late, and then only with: This needs documentation.

you might also wanna [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) those two down to one.

ckaenzig:
Oops, it looks like I forgot about this pull request, sorry about that. The same change has now been committed as 8b00664, so this PR is redundant, I'm closing it.

PR #1122, opened 2015-05-17
removing the default declaration to allow override.

daveseff:
Having these entries did not allow for a calling module to override this declaration.  The agent would throw the following error: 

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Duplicate declaration: Apache::Mod[expires] is already declared in file /etc/puppetlabs/puppet/environments/production/modules/apache/manifests/default_mods.pp:57; cannot redeclare at....


igalic:
i'm a little confuzzled about this.

can you post a snippet to repro this?
maybe even add that as a spec test?

jonnytpuppet:
ping @daveseff

PR #1123, opened 2015-05-17
Adding <Files> Section

daveseff:
Allows for adding The following example section:
<Files foo.php>
    Order deny,allow
    Deny from all
    Allow from 127.0.0.1
</Files>


igalic:
what can you do with `$files`, that you cannot do with `$directories => { provider => files}` ?

mhaskel:
ping @daveseff ?

daveseff:
Ok I haven't tried that. I'll give it a go.

PR #1126, opened 2015-05-20
Support puppetlabs-concat 2.x

domcleal:
2.0.0 is compatible with current 1.x usage, so permit either 1.x or 2.x
to satisfy the concat dependency.  Move all ensure parameters from
concat::fragment to concat itself.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,9 +1,7 @@
 fixtures:
   repositories:
     stdlib: "git://github.com/puppetlabs/puppetlabs-stdlib.git"
-    concat:
-      repo: "git://github.com/puppetlabs/puppetlabs-concat.git"
-      branch: "1.2.x"
+    concat: "git://github.com/puppetlabs/puppetlabs-concat.git"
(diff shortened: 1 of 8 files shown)


mcanevet:
:+1:

mhaskel:
@domcleal thanks for the contribution! We're currently investigating some possible issues with concat 2.0.0, putting this PR on hold until those investigations are complete.

domcleal:
Was that https://tickets.puppetlabs.com/browse/MODULES-2061?

mhaskel:
@domcleal that was one of them, but there's also a ticket for weird interactions with PE and Concat 2.0.0 and apache: https://tickets.puppetlabs.com/browse/MODULES-2054

duritong:
This patch also misses 2 concat::fragment to remove ensure => present. in init.pp and namedvirtualhost.pp

domcleal:
@duritong thanks, well-spotted!  Updated.

bmjen:
:+1:

PR #1130, opened 2015-05-29
create resources with hiera

bleuchtang:
With this PR we can create vhosts, balancers, mods, npms, and custom_configs directly with hiera. here an example:

   ---
     apache::vhosts:
      'supervision.example.net':                                                          
        port: 80                                                                    
        docroot: '/usr/share/graphite-web/static/'                                  
        wsgi_daemon_process: '_graphite'                                            
        wsgi_daemon_process_options:                                                
          processes: 5                                                              
          threads: 5                                                                
          display-name: "%{literal('%')}{GROUP}"                                    
          inactivity-timeout: 120                                                   
          user: '_graphite'                                                         
          group: '_graphite'                 
    classes:                                                                        
      - apache

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -72,6 +72,15 @@ class apache (
   $allow_encoded_slashes  = undef,
   $package_ensure         = 'installed',
   $use_optional_includes  = $::apache::params::use_optional_includes,
+
+  ### START Hiera Lookups ###
+  $vhosts                 = {},
+  $balancers              = {},
+  $mods                   = {},
+  $npms                   = {},
+  $custom_configs         = {},
+  ### END Hiera Lookups ###
+
 ) inherits ::apache::params {
   validate_bool($default_vhost)
   validate_bool($default_ssl_vhost)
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -72,6 +72,15 @@
   $allow_encoded_slashes  = undef,
   $package_ensure         = 'installed',
   $use_optional_includes  = $::apache::params::use_optional_includes,
+
+  ### START Hiera Lookups ###
+  $vhosts                 = {},
+  $balancers              = {},
+  $mods                   = {},
+  $npms                   = {},
```
igalic: what are [npms](https://github.com/puppetlabs/puppetlabs-apache/search?q=npm&type=Code&utf8=%E2%9C%93)?

igalic:
as i have already mentioned in many puppetlabs modules', we prefer not to add hiera lookups.

this is the kind of code that one abstract in their profile. if we put it into a base-block module, we'd have to argue with people over what goes and what doesn't go into these kind of modules, and how exactly to compose the results.

manifests/init.pp
bleuchtang: Oups that a typo, it should be the plural of mpm (mpms or mpm_array)

jonnytpuppet:
Hi There

I'm closing this as we do not accept submissions that pertain to Heira by default. If you would like this feature, perhaps re-submit a PR that does this inside a separate class, and we can maybe take this further.

Thanks!

PR #1133, opened 2015-06-03
Amazon Linux does not support systemd. Having the versioncmp makes th…

cliles:
…is fail on newer versions (which do not yet support systemd). Run this on 2015.03 and you end up with an apache server that will not start.

### DIFF
--- a/manifests/default_mods.pp
+++ b/manifests/default_mods.pp
@@ -12,7 +12,7 @@ class apache::default_mods (
       if versioncmp($apache_version, '2.4') >= 0 {
         # Lets fork it
         # Do not try to load mod_systemd on RHEL/CentOS 6 SCL.
-        if ( !($::osfamily == 'redhat' and versioncmp($::operatingsystemrelease, '7.0') == -1) and !($::operatingsystem == 'Amazon' and versioncmp($::operatingsystemrelease, '2014.09') <= 0  ) ) {
+        if ( !($::osfamily == 'redhat' and versioncmp($::operatingsystemrelease, '7.0') == -1) and !$::operatingsystem == 'Amazon' ) {
           ::apache::mod { 'systemd': }
         }
         ::apache::mod { 'unixd': }


manifests/default_mods.pp
```diff
@@ -12,7 +12,7 @@
       if versioncmp($apache_version, '2.4') >= 0 {
         # Lets fork it
         # Do not try to load mod_systemd on RHEL/CentOS 6 SCL.
-        if ( !($::osfamily == 'redhat' and versioncmp($::operatingsystemrelease, '7.0') == -1) and !($::operatingsystem == 'Amazon' and versioncmp($::operatingsystemrelease, '2014.09') <= 0  ) ) {
+        if ( !($::osfamily == 'redhat' and versioncmp($::operatingsystemrelease, '7.0') == -1) and !$::operatingsystem == 'Amazon' ) {
```
nibalizer: Probably this should be RedHat

nibalizer: The original implementation seems to imply that Amazon Linux after 2014.09 uses systemd. Is this incorrect?

cliles:
we are running 2015.03 and it's not on systemd.

nibalizer:
Huh. Can you boot a 2014 node and verify that it is/is not running systemd?

cliles:
2014.09 is not using systemd AFAIK. I don't see it running on a 2014.09 ami grabbed from the marketplace. Looking at the function reference for versioncmp it would return -1 if the version is lower than 2014.09 and 0 if the version was equal to 2014.09. 

```
( !($::osfamily == 'redhat' and versioncmp($::operatingsystemrelease, '7.0') == -1) and !($::operatingsystem == 'Amazon' and versioncmp($::operatingsystemrelease, '2014.09') <= 0  ) )
```
When running on a version higher than 2014.09 you get.
```
! (false and false) and !(true and false) 
```
This will cause mod_systemd to be loaded, which does not exists, thus rendering an apache configuration that will not start.

I see this as the original author "hoped" that versions > 2014.09 would be on systemd, which is not the case I am encountering.

nibalizer:
I'm hoping https://github.com/puppetlabs/puppetlabs-stdlib/pull/473 makes this easier.

bmjen:
I've confirmed on our AWS account that 2014.09 and 2015.03 indeed... are not on systemd. Thanks for the find @cliles !

We are going to re-PR this fix into 1.5.x branch for the upcoming release and close this PR, and it should propagate back down to master.

mhaskel:
I manually merged this into 1.5.x, and it will get merged back into master once the release is finalized.

PR #1155, opened 2015-07-07
mod::alias should be included when the aliases parameter is used

pcfens:
`mod::alias` should be included when `aliases` is set, otherwise Apache fails to start.

I've added a simple rspec test around this as well, though it seems to stand out from the other tests around vhosts. Ideally acceptance tests would catch things like this as well, but a separate test for each parameter seems like a rather large undertaking (though I'm happy too look at adding one for this case if desired).

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -363,7 +363,7 @@ define apache::vhost(
   }
 
   # Load mod_alias if needed and not yet loaded
-  if ($scriptalias or $scriptaliases != []) or ($redirect_source and $redirect_dest) {
+  if ($scriptalias and $scriptaliases != []) or ($aliases and $aliases != []) or ($redirect_source and $redirect_dest) {
     if ! defined(Class['apache::mod::alias'])  and ($ensure == 'present') {
       include ::apache::mod::alias
     }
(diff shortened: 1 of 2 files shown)


manifests/vhost.pp
```diff
@@ -362,7 +362,7 @@
   }
 
   # Load mod_alias if needed and not yet loaded
-  if ($scriptalias or $scriptaliases != []) or ($redirect_source and $redirect_dest) {
+  if ($scriptalias or $scriptaliases != []) or ($aliases or $aliases != []) or ($redirect_source and $redirect_dest) {
```
hunner: Looks like a logic issue... could you make this `if ($scriptalias and $scriptaliases != []) or ($aliases and $aliases != []) or ($redirect_source and $redirect_dest) {` ? Or am I just misunderstanding what this is supposed to do?

bmjen: Seems the original logic was wrong. Can you update this line to:
`if ($scriptalias and $scriptaliases != []) or ($aliases and $aliases != []) or ($redirect_source and $redirect_dest) {`

pcfens: Changing to `and` makes more sense, but the initial patch was imitating what was already in place (or). I'll update, and push shortly.

igalic:
thanks @pcfens!

PR #1158, opened 2015-07-17
mod_passenger: Allow setting PassengerSpawnMethod

wubr:


igalic:
@wubr this will need documentation and perhaps some mild validation to catch the most erregious of errors

wubr:
Good call.  How do the new commits look?

igalic:
nice, @wubr!
would you mind squashing down those commits a bit?

wubr:
Squashed.  Thanks for bearing with me!

igalic:
thanks a lot @wubr

PR #1159, opened 2015-07-22
Add support for user modifiable installation of mod_systemd and pidfile locations:

vamegh:
default_mods.pp --> added use_systemd option, if running a custom build of apache on
                    centos7, user modifiable boolean option provided

init.pp --> added pidfile and use_systemd option both of which go to params.pp and allow
            these values to be set in hiera. changed all osfamily choices for conf_dir
            to draw values from the $pidfile variable.

params.pp --> added $use_systemd, by default set to true to mimic original behaviour.
              added $pidfile, set to mimic exact same behaviour as used in init.pp,
              except now user changeable via hiera.

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,7 +448,7 @@ apache::vhost { 'suphp.example.com':
   suphp_configpath => '/etc/php5/apache2',
   directories      => [
     { 'path'  => '/home/appuser/myphpapp',
-      'suphp' => { 
+      'suphp' => {
         user  => 'myappuser',
         group => 'myappgroup',
       },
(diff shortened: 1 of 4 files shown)


tphoney:
Can you remove the edit to the metadata.json, this will be updated as part of a release. 
The release will include many changes. so it may not be a x.y.1 release

vamegh:
Sure, this should be done now.

tphoney:
sorry more stuff, apologies for getting back to this slowly. This will need a readme update and possibly some tests surrounding this new functionality

vamegh:
Hi, I've updated the readme to include the 2 new options, they are really only options to be passed ideally via hiera, if required. I don't believe tests are necessary, but If you wish to write the tests, feel free. By default it will keep the apache module behaviour exactly the same, it just provides the user with more customisation, especially if using custom built rpms.

mhaskel:
@vamegh Thanks for the contribution! Looks like we won't need tests for this, but the code will need to be rebased. Thanks!

vamegh:
Hi, 

Sorry for the delay.. It should be rebased now.

tphoney:
@vamegh Thanks for the work.

PR #1160, opened 2015-07-22
add possibility to set icons_path to false so no alias will be set for it

tjikkun:
It is not always that we want Alias /icons in our config, but right now if the alias mod gets pulled in there is no way around that. This commit fixes that.


tphoney:
@tjikkun thanks for the work, but this is currently failing TravisCI, can you please look at the failure.

tjikkun:
Hi @tphoney, sorry for that. Should be better now but let's wait and see. If Travis still complains I will fix it tomorrow.

igalic:
thanks @tjikkun!

PR #1162, opened 2015-07-24
Adding use_optional_includes parameter to vhost define.

cropalato:
I was testing with debian jessie connecting to a puppetmaster 3.7.3-1 + hiera and apache/templates/vhost/_additional_includes.erb was reporting @use_optional_includes as undef.


igalic:
thank you @cropalato!

PR #1165, opened 2015-07-30
fixes timing of mod_security tests for aio

bmjen:
While running AIO tests on RHEL6, the httpd service takes a bit longer
to init. Since Apache init scripts return and completes init asynchronously
this causes tests to fail. Adding a short sleep here to wait for the service
to init compensates for the slower init.


igalic:
this makes me sad :(

bmjen:
@igalic :(

PR #1166, opened 2015-07-30
we also want to have groups with '-' like vhost-users

sunflower9973:
extended regex for suexec users/groups to accept '-' additionally to word characters

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -149,7 +149,7 @@ define apache::vhost(
   # Input validation begins
 
   if $suexec_user_group {
-    validate_re($suexec_user_group, '^\w+ \w+$',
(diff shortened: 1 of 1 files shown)


igalic:
@sunflower9973 if you insist on magic internet points, you may want to either change your committer email in this commit, or else add it to your GitHub profile.

from `groupadd(8)` (on my linux box)

> CAVEATS
> 
> It is usually recommended to only use groupnames that begin with a lower case letter or an underscore, followed by lower case letters, digits, underscores, or dashes. They can end with a dollar sign. In regular expression terms: [a-z_][a-z0-9_-]*[$]?

so from that pov we're safe. (on linux)

igalic:
thanks @sunflower9973!

PR #1120, opened 2015-05-08
SecRuleEngine parameter for modsecurity

unixninja:
Added a parameter that sets mod security SecRuleEngine. Allowing for DetectionOnly or off.

PR #1168, opened 2015-07-30
[#puppethack] Adding ability to enable/disable the secruleengine through a parameter

igalic:
rebase / squash of #1120.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1079,6 +1079,10 @@ Name of package to install containing crs rules
 
 Directory to install the modsec configuration and activated rules links into
 
+#####`modsec_secruleengine`
+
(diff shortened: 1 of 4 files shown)


igalic:
still no tests though, cuz, i'm finding them quite pointless in this case.
@bmjen seens to disagree…

bmjen:
@igalic I don't think I ever disagreed. :P 

I think after looking at the spec tests for mod_security, a unit test is likely not necessary. Will merge once travisci is done.

jackdpeterson:
Can this be merged in? I'd like to be the first test-case :-)

igalic:
yes, yes it can. (now that travis is done)

PR #1171, opened 2015-07-30
Move the description of the timeout parameter to the correct module.

timmooney:
The timeouts parameter was under the wrong module in README.md.  The change moves it from the description of apache::mod::version to apache::mod::reqtimeout


cyberious:
@timmooney closing, resolved as part of readme rewrite

PR #1173, opened 2015-07-30
RedHat derived systems >= 7 use mod_dir of conf.modules.d, not conf.d

timmooney:
I'm not crazy about using $::apache::version::distrelease as the selector to determine the default for mod_dir, but that's what was done for other RHEL7 vs. earlier decisions.

I kind of feel like using the $::apache::version itself would be better, since then we don't need to update all of the code when RHEL8 comes out.  Then again, that will be years, so maybe it's not worth worrying about right now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,7 +416,7 @@ Setting this to 'false' stops the user resource from being created. This is for 
 
 #####`mod_dir`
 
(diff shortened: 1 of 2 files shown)


timmooney:
I'll work on getting the error fixed and an amended pull request.  I don't work with git much, so it might take me a while.

PR #1175, opened 2015-07-31
Fix apache::mod::cgid so it can be used with the event MPM

MasonM:
Before this, attempting to include `apache::mod::cgid` when using the event MPM
raised the following error:
`Could not find resource 'Class[Apache::Mod::Worker]' for relationship on 'Class[Apache::Mod::Cgid]'`


mhaskel:
Great, thanks @MasonM !

PR #1177, opened 2015-08-03
Added option to disable indexes globally.

tthayer:
This adds "Options -Indexes" to the base apache configuration file, which will disable indexes on directories that don't explicitly have "Options Indexes" set.


tthayer:
I'm not seeing where the conflicts are and github isn't helping.

igalic:
i think this makes for a bad name, or option, in general.

in my opinion this should be an `option` parameter which takes a string or an array of options.

tthayer:
That is a good point. This is the only use case I've seen using the Option outside of a directory definition so I hadn't considered that use case.

hunner:
The conflicts should become apparent if you do a rebase onto the current master.

tthayer:
OK, rebased and updated now.

tthayer:
And lost my changes. Time to read about how to git again.

igalic:
here's a nice rundown of git by @eevee: http://eev.ee/blog/2015/04/24/just-enough-git-to-be-less-dangerous/

PR #879, opened 2014-09-30
Need fcgid to load after unixd on RHEL7

underscorgan:

PR #1178, opened 2015-08-10
load unixd before fcgid on all operating systems (see #879)

sethlyons:


sathieu:
@rvrangel Does #1331 fix it for you too?

PR #1153, opened 2015-07-06
Make Docroot optional

robbat2:
Implements
https://tickets.puppetlabs.com/browse/MODULES-2120

PR #1179, opened 2015-08-11
Improve docroot flexibility with additional_includes

roidelapluie:
If DocumentRoot is in one of the files defined by additional_includes,
then we should NOT include it in the configuration we output.

Permit this only if $manage_docroot is false, $docroot is explicitly
undef, and $additional_includes is non-empty.

Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>

Rebase of #1153

### DIFF
--- a/README.md
+++ b/README.md
@@ -1823,7 +1823,7 @@ Sets the list of resources to look for when a client requests an index of the di
 
 ##### `docroot`
 
-**Required**. Sets the [`DocumentRoot`][] location, from which Apache serves files.
+**Required**. Sets the [`DocumentRoot`][] location, from which Apache serves files. Can be explicitly set to undef if additional_includes is defined.
 
 ##### `docroot_group`
 
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -146,6 +146,12 @@ define apache::vhost(
     validate_array($rewrites)
     validate_hash($rewrites[0])
   }
+  # Validate the docroot as a string if any of the following are true:
+  # - $manage_docroot is true
+  # - $additional_includes is undefined or empty
+  if $manage_docroot or ! $additional_includes or empty($additional_includes) {
+    validate_string($docroot)
+  }
 
   # Input validation begins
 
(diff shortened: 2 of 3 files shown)


igalic:
sounds sensible.
do you think we need tests for this?
do you think we can test this?

roidelapluie:
I have no idea. this is just a rebase of #1153 . @robbat2 what's your opinion?

README.md
```diff
@@ -1823,7 +1823,7 @@ Sets the list of resources to look for when a client requests an index of the di
 
 ##### `docroot`
 
-**Required**. Sets the [`DocumentRoot`][] location, from which Apache serves files.
```
igalic: i'm still unclear in how far this is no longer **Required**, since in vhost.pp, we still have this:

```puppet
 define apache::vhost(
   $docroot,
   $manage_docroot              = true,
  …
```

robbat2: I was explicitly passing undef for $docroot. Maybe keep it commented as a required param, but document that it can be explicitly undef.

roidelapluie: fixed

igalic: @robbat2 this is interesting, and surprising.
`undef` usually erases the value, taking on whatever is default, i didn't expect this outcome… cc / @hlindberg

igalic: @roidelapluie somehow, i was hoping you'd fix it by making it **not** required anymore ;)

roidelapluie: In Puppet 4, I was expecting Optional[String] $test to be the same as $test = undef. But it is not. So you have to use Optional[String] $test = undef. It means that undef is a thing.

igalic: `undef` can be tricky to pass in hiera (it's not, it's `!!null`, but still ;)

hlindberg:
@igalic not sure what you wanted me to comment on, or if there was a question. What was said elsewhere is true `Optional[T] $var` is not the same as `Optional[T] $var = undef` since the former specifies a required parameter that may be set explicitly to `undef`. It is however difficult to do so, since the rules for setting values take an explicit `undef` to mean "use the default". It should be possible to set an explicit undef via data binding. (Yes, it is complex, but now in 4.x it is at least consistent).

igalic:
@roidelapluie ping

roidelapluie:
closing this, letting @robbat2 reopening if needed

yakatz:
I opened a similar pull request, but it requires passing false for the docroot value, so avoids the odd undef value.

https://github.com/puppetlabs/puppetlabs-apache/pull/1222

PR #1184, opened 2015-08-16
The purge_vhost_configs parameter is actually called purge_vhost_dir

mpdude:
See https://github.com/puppetlabs/puppetlabs-apache/commit/2dbb1ec540e71fb2f920891d213d88774faf5c14#diff-60ae41fd0a31977447947f59940ee9a4R31


mhaskel:
Good catch, thanks @mpdude!

PR #1185, opened 2015-08-18
Add ListenBacklog for mod worker (MODULES-2432)

mwhahaha:
This change adds the ListenBacklog option for the worker module.


mhaskel:
Thanks @mwhahaha !

PR #1191, opened 2015-08-26
Don't qualified call to defined resource type

PierreR:
According to best practices:
(https://docs.puppetlabs.com/puppet/latest/reference/lang_defined_types.html#resource-uniqueness)

See these accepted PRs:
- https://github.com/theforeman/puppet-foreman/pull/346
- https://github.com/jfryman/puppet-nginx/pull/666

### DIFF
--- a/manifests/mod/rewrite.pp
+++ b/manifests/mod/rewrite.pp
@@ -1,4 +1,4 @@
 class apache::mod::rewrite {
   include ::apache::params
-  ::apache::mod { 'rewrite': }
+  apache::mod { 'rewrite': }
 }


DavidS:
Hi Pierre,

This change is no improvement to the module, as removing the full qualification only leads to ambiguous lookups for this define, leading to more work for the puppet master when compiling and potential errors in the worst edge cases.

Regards, David

CC @mmoll , @3flex

PierreR:
@DavidS  That's not what the puppet documentation is saying:

https://docs.puppetlabs.com/puppet/3.8/reference/lang_namespaces.html#syntax

> This is ugly and should be unnecessary, but is occasionally required due to an outstanding design issue. However, you can avoid this issue by enabling the future parser in Puppet 3.8. See below for details.

DavidS:
Since we need to stay backwards compatible with pre-future-parser versions, this change still can only go in after we've dropped support for those. That will take a while.

PierreR:
@DavidS The proposed change is backwards compatible. It might ease the transition to 4.x, that's all.

DavidS:
The following demonstrates the problem with puppet 3.8.1 both with and without the future parser. In both cases the unqualified reference finds the "wrong" define.

```
$ cat tmp/fqdr.pp
define apache::mod() { notify { "apache::mod": loglevel => warning } }
define apache::mod::rewrite::apache::mod() { notify { "apache::mod::rewrite::apache::mod": loglevel => warning } }

class apache::mod::rewrite {
::apache::mod { 'rewrite': }
apache::mod { 'rewrite': }
}

include apache::mod::rewrite
$ puppet apply tmp/fqdr.pp
Notice: Compiled catalog for davids-macbook-pro.local in environment production in 0.01 seconds
Warning: apache::mod::rewrite::apache::mod
Warning: /Stage[main]/Apache::Mod::Rewrite/Apache::Mod::Rewrite::Apache::Mod[rewrite]/Notify[apache::mod::rewrite::apache::mod]/message: defined 'message' as 'apache::mod::rewrite::apache::mod'
Warning: apache::mod
Warning: /Stage[main]/Apache::Mod::Rewrite/Apache::Mod[rewrite]/Notify[apache::mod]/message: defined 'message' as 'apache::mod'
Notice: Finished catalog run in 0.02 seconds
$ puppet apply --parser=future tmp/fqdr.pp
Notice: Compiled catalog for davids-macbook-pro.local in environment production in 0.51 seconds
Warning: apache::mod::rewrite::apache::mod
Warning: /Stage[main]/Apache::Mod::Rewrite/Apache::Mod::Rewrite::Apache::Mod[rewrite]/Notify[apache::mod::rewrite::apache::mod]/message: defined 'message' as 'apache::mod::rewrite::apache::mod'
Warning: apache::mod
Warning: /Stage[main]/Apache::Mod::Rewrite/Apache::Mod[rewrite]/Notify[apache::mod]/message: defined 'message' as 'apache::mod'
Notice: Finished catalog run in 0.05 seconds
$
```

Granted the chances of someone having a `apache::mod::rewrite::apache::mod` define around are very slim, but it is a good pattern to use overall, as there are often other cases where collisions are much more frequent.

DavidS:
I also teste this now with puppet 4.2.1 and this is probably what you're referring to:

```
$ puppet apply tmp/fqdr.pp
Error: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Apache::Mod[rewrite] is already declared in file /Users/david/git/puppetlabs-stdlib/tmp/fqdr.pp:5; cannot redeclare at /Users/david/git/puppetlabs-stdlib/tmp/fqdr.pp:6 at /Users/david/git/puppetlabs-stdlib/tmp/fqdr.pp:6:1 on node davids-macbook-pro.local
$
```

It doesn't change the fact though, that we'll need to support puppet 3 for quite a while.

PierreR:
@DavidS Thanks for your example. It is really appreciated.  Somehow I would prefer to slam the one who creates `apache::mod::rewrite::apache::mod` instead ;-) but of course this is your choice as a library author. 

The initial reason for me to submit the PR was to be compliant with [`language-puppet`](https://github.com/bartavelle/language-puppet) that I use heavily to catch all sort of bugs in my manifests. It would refuse such construct (with defined resources). I still have to test this but normally the latest commits in `language-puppet` should have fixed it (and be more forgiven).

Thanks again for the open discussion.

DavidS:
Given that that seems to be a re-implementation of the puppet parser (?) it would seem to be appropriate to follow the reference implementation's behaviour instead of adapting existing inputs to language-puppet's brokenness...

PierreR:
It does much more than parsing (check requirements, undefined variables, file existences, show template interpretation, ...) and try to push to "best practices". You would be surprised how much bugs this tool can catch.

PR #1192, opened 2015-08-26
Geoip package

stumped2:
Currently, including `apache::mod::geoip` fails on at least Ubuntu 14.04 because a required lib is not installed (or is not being installed), and there's no way to tell it what supporting packages it might need.

This adds the ability to install any need package(s) when enabling mod_geoip.


mhaskel:
Hi @stumped2 ,

If this is failing on Ubuntu 14.04, I think rather than adding a new parameter, the `apache::params` class should be updated to have the required package. See https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/params.pp#L188-L207. You'll probably need to do something to make sure the change only applies for trusty, but I think that's the right place for the change to be happening. Thanks!

stumped2:
oh, scratch this. It looks to actually be fixed in master, but not in the latest tag (1.6.0).

*grumble*grumble*

PR #1193, opened 2015-08-27
Catch that mod_authz_default has been removed in Apache 2.4

mpdude:
Having it in the server config will prevent the server from starting as the module file cannot be found.

### DIFF
--- a/manifests/mod/authz_default.pp
+++ b/manifests/mod/authz_default.pp
@@ -1,3 +1,9 @@
-class apache::mod::authz_default {
-  ::apache::mod { 'authz_default': }
+class apache::mod::authz_default(
+    $apache_version = $::apache::apache_version
+) {
+  if versioncmp($apache_version, '2.4') >= 0 {
+    warning('apache::mod::authz_default has been removed in Apache 2.4')
+  } else {
+    ::apache::mod { 'authz_default': }
+  }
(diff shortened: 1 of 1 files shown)


manifests/mod/authz_default.pp
```diff
@@ -1,3 +1,9 @@
-class apache::mod::authz_default {
-  ::apache::mod { 'authz_default': }
+class apache::mod::authz_default(
+    $apache_version = $::apache::apache_version
+) {
+  if versioncmp($apache_version, '2.4') >= 0 {
```
hunner: Could you remove the parameter and just reference $::apache::apache_version directly? I don't think there is any reason for anyone to pass the version in, is there?

hunner:
Could you squash this into one commit? Thanks.

manifests/mod/authz_default.pp
mpdude: I've seen this in many other modules, for example https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/worker.pp#L65, and simply assumed that there must be a good reason that I have too little Puppet knowledge to understand... So what?

mpdude:
Squashed.

mpdude:
*ping* @hunner

PR #1194, opened 2015-08-31
Add an option to configure PassengerLogFile

igalic:
PassengerLogFile (available since 5.0.5) allows us to configure a
separate file where (all) applications running under passenger will send
their stdout/stderr to.

This is helpful in reducing noise when trying to parse httpd's error log


DavidS:
Hi!!

please add a note to README.passenger.md for this.

Cheers, David

igalic:
hi @DavidS,

the mod_passenger documentation in README.passenger.md seems to be more outdated than the one in README.md… maybe we should just delete the former, and update the latter?

DavidS:
@igalic, you are definitely more qualified to judge that than me.

ekohl:
:+1: looks good.

HelenCampbell:
Thanks @igalic !

PR #1196, opened 2015-09-02
Removed enforced .conf extension on custom_config.pp

MattLud:
The current custom_config automatically updates the file to use .conf
rather than rely on the name provided. Any configs/vhosts/etc. using
this have to predict this behavior rather than rely on what's provided
in the name of the config.

This is not exposed or configurable so removing it should help make
things easier from a configuration aspect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1632,7 +1632,7 @@ Specifies an [array][] of [options](http://httpd.apache.org/docs/current/mod/mod
 
 #### Define: `apache::custom_config`
 
-Adds a custom configuration file to the Apache server's `conf.d` directory. If the file is invalid and this define's `$verify_config` parameter is 'true', Puppet throws an error during a Puppet run.
+Adds a custom configuration file to the Apache server's `conf.d` directory. If the file is invalid and this define's `$verify_config` parameter is 'true', Puppet throws an error during a Puppet run. Uses extension from name and does not change it .conf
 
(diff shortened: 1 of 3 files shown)


Seakip18:
I can also do something similar to priority where ".conf" is used by default if not set to false.

Seakip18:
Here's a branch with the optional extension parameter - https://github.com/puppetlabs/puppetlabs-apache/commit/456cf9335e0e0ba9fbadfd8afed648817f3d4a85

bmjen:
Thanks @MattLud. I'm unsure as to what the goal would be. 

A) Is it to clear up confusion as to not produce a file named "filename.conf.conf"? 

B) Or is there a desire to use a custom extension like "filename.ext"?

I would say that A) the need to clear up confusion is important, but I would think the most common use-case would still involve using the ".conf" extension.

MattLud:
@bmjen  - The goal would be to allow control of the file extension. This would allow you to avoid using the implicit .conf. where you end up with x.conf.conf accidentally. 

I think it would be better though to use a default of .conf and keep it compatible with existing code which may rely on adding .conf implicitly. I'll close this PR and open one on the other branch which uses the default behavior if that's ok.

bmjen:
@MattLud sounds like a good plan! Thanks!

PR #1198, opened 2015-09-07
Deflate "application/json" by default

leopoiroux:
Hello,
How can I customize the "types" of "deflate " while using " default_mods " ?
I need to deflate "application/json".

JSON is a common text resource types on the web which should be served with HTTP compression:
https://zoompf.com/blog/2012/02/lose-the-wait-http-compression

Thks


tphoney:
If we are changing the default behaviour, could you update the readme as well.

leopoiroux:
Done. Thank you ;)

PR #1200, opened 2015-09-14
Fix typo of MPM_PREFORK for FreeBSD package install

edmundcraske:
Was unsetting 'MPM_PERFORK', so 'MPM_PREFORK' would not have been
unset correctly.


bmjen:
:+1: Thanks for the catch @edmundcraske!

PR #1201, opened 2015-09-16
Removed enforced .conf extension on custom_config.pp

MattLud:
The custom_config resource forcibly renamed the file to include .conf as
the extension. The behavior was not documented and did not reflect the
name of the resource in this case. Any configs including those would
need to remember to include .config even if the base file name did not
have it.

By adding the parameter and defaulting it to '.conf', this commit allows
backward compatibility while also allowing custom config extensions in
the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1668,6 +1668,10 @@ This parameter is only used if the [`verify_config`][] parameter's value is 'tru
 
 Specifies whether to validate the configuration file before notifying the Apache service. Valid options: Boolean. Default: `true`.
 
+##### `default_ext`
+
+Specifies what extension to append on end of custom config file. Set to '' to keep original extension. Default: `.conf`.
+
 #### Define: `apache::fastcgi::server`
 
 Defines one or more external FastCGI servers to handle specific file types. Use this define with [`mod_fastcgi`][FastCGI].
--- a/manifests/custom_config.pp
+++ b/manifests/custom_config.pp
@@ -7,6 +7,7 @@ define apache::custom_config (
   $source         = undef,
   $verify_command = $::apache::params::verify_command,
   $verify_config  = true,
+  $default_ext    = '.conf',
 ) {
 
   if $content and $source {
@@ -28,10 +29,11 @@ define apache::custom_config (
   } else {
     $priority_prefix = ''
   }
-
+  
+  
   ## Apache include does not always work with spaces in the filename
   $filename_middle = regsubst($name, ' ', '_', 'G')
-  $filename = "${priority_prefix}${filename_middle}.conf"
+  $filename = "${priority_prefix}${filename_middle}${default_ext}"
 
   if ! $verify_config or $ensure == 'absent' {
     $notifies = Class['Apache::Service']


HelenCampbell:
Hi,
it would be great if there were tests for this new functionality. I would like to hear from community members to decide if this is something they would want added.

tphoney:
This re-implementation of #1196, it is still not a use-case that does fit most use cases for users.
why change the extension of the config file ?

MattLud:
It keeps the ".conf" extension the same by default but exposes it for modification in the event that the user doesn't want the ".conf" forcibly applied to their file. 

The current setup would require that conf files are named without an extension or end up with "foo.conf.conf".

hunner:
What is the effective use case of this though? "I don't like .conf on my file names" is not a use case as it does not affect the outcome and follows convention. Also apache.conf includes *.conf files, so other extensions would not be included.

MattLud:
Use case:
Multiple vhost definitions use rewrite.common as an additional_include and that file needs to be in place before apache starts with those vhosts - otherwise it will fail. The vhosts currently expect it to also be placed in conf.d with the apache configs. 

This change allows custom_config to put the file in the right place, name it properly and handle notifying the appropriate service. Otherwise, it'd place it as rewrite.common.conf and apache would try to parse it during the service start and likely fail. 

The alternative to this change is to place these common files manually and duplicate (unnecessarily, I believe) the logic of custom_config all due to the hard coded .conf.

manifests/custom_config.pp
```diff
@@ -28,10 +29,11 @@
   } else {
     $priority_prefix = ''
   }
-
+  
+
```
hunner: Extra whitespace (newline and two trailing spaces)

manifests/custom_config.pp
```diff
@@ -7,6 +7,7 @@
   $source         = undef,
   $verify_command = $::apache::params::verify_command,
   $verify_config  = true,
+  $default_ext    = '.conf',
```
hunner: Could you call this `$file_extension` instead?

README.md
```diff
@@ -1668,6 +1668,10 @@ This parameter is only used if the [`verify_config`][] parameter's value is 'tru
 
 Specifies whether to validate the configuration file before notifying the Apache service. Valid options: Boolean. Default: `true`.
 
+##### `default_ext`
+
+Specifies what extension to append on end of custom config file. Set to '' to keep original extension. Default: `.conf`.
```
hunner: I think the phrase "Set to '' to keep original extension" is odd. Under your use case, you could pass '.common' here instead of having it be part of `$name` so I don't think `''` would be something I would recommend.

README.md
```diff
@@ -1668,6 +1668,10 @@ This parameter is only used if the [`verify_config`][] parameter's value is 'tru
 
 Specifies whether to validate the configuration file before notifying the Apache service. Valid options: Boolean. Default: `true`.
 
+##### `default_ext`
```
hunner: This should be inserted in alphabetic order (with `file_extension` rename)

karmix:
I see a problem (not directly related with your PR) with generating the file name from the resource name in custom_config.  It makes it difficult to generate files with the same name, but differ in confdir, prefix, or extension.  For example, the straight forward approach to creating myvhost.conf and myvhost.common (using different extensions) would not work because you cannot have two resources named 'myvhost'.  Things would get worse if you needed to put similar files in different directories.

Would something like c8abd1fd60c6c6e7bfb0844885add85a585f5aca, which allows you to override the filename so it is no longer tied to the resource name, address your issue?

README.md
MattLud: It would require that your original source file be named with no extension, which seems weird. You might not know if "rewrite" is a conf or common extension.

bmjen:
Resolved in #1270

PR #1206, opened 2015-09-22
Fix erroneous example in README.md

aduitsis:
Fix error regarding the shib_request_settings setting for a vhost directory. Also, fix value of said attribute to a hash instead of a string.


igalic:
thanks @aduitsis!

PR #1207, opened 2015-09-22
MODULES-2613 Small style cleanup to templates/vhost/_additional_inclu…

snarlistic:
…des.erb

The file templates/vhost/_additional_includes.erb can use a small cleanup (IMHO).

Without this commit the includes are not nicely indented in line with the comment and the rest of the configuration file. This commit fixes that.


tphoney:
Thanks for the work

PR #1199, opened 2015-09-11
Fix mod_authz_ldap to work on RHEL7 and mod_ldap

simonoded:
Hi I fixed the load.erb to have condition on the mod_ldap for RedHat 7 and apache 2.4

PR #1209, opened 2015-09-24
move the RHEL7 authnz to the authnz template

simonoded:
Hi per our discussion, I moved the setting for authnz_ldap to the authnz template
Regards

### DIFF
--- a/templates/mod/authnz_ldap.conf.erb
+++ b/templates/mod/authnz_ldap.conf.erb
@@ -3,3 +3,6 @@ LDAPVerifyServerCert On
 <% else -%>
 LDAPVerifyServerCert Off
 <% end -%>
+<% if scope.lookupvar('::apache::version::distrelease') == '7' -%>
+LoadModule ldap_module modules/mod_ldap.so
+<% end -%>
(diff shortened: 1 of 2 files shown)


igalic:
@simonoded if memory serves me right what we actually discussed in #1199 is that you should use the specialized class apache::mod::authnz_ldap, that class will handle the inclusion of mod_ldap, in the correct order, on its own

simonoded:
Hi Igor,

I did move the additional setting to the authnz_ldap.conf.erb 

Which is called by the apache::mod::authnz_ldap module.

 

Regards

 

 

 

Oded Simon

M: 0410 881 316

E: simonoded@gmail.com

ekohl:
There is already a solution for this. Looking at #1199 you now have:
```puppet
class { 'apache': }
class { 'apache::mod::ssl': }
apache::mod { 'authnz_ldap': }  # <-- this here
class { '::apache::mod::php': }
```

If you use the class apache::mod::authz_ldap then it should already work:

```puppet
class { 'apache': }
class { 'apache::mod::ssl': }
class { 'apache::mod::authz_ldap': } # <-- this here
class { '::apache::mod::php': }
```

PR #1212, opened 2015-09-28
Allow to overwrite ssl mutext configuration for mod ssl

nabam:


tphoney:
On initial inspection this would require a readme update, to explain the new attribute

manifests/mod/ssl.pp
```diff
@@ -10,33 +10,39 @@
...
-    'debian': {
-      if versioncmp($apache_version, '2.4') >= 0 {
-        $ssl_mutex = 'default'
-      } elsif $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease == '10.04' {
-        $ssl_mutex = 'file:/var/run/apache2/ssl_mutex'
-      } else {
-        $ssl_mutex = "file:\${APACHE_RUN_DIR}/ssl_mutex"
+
+  if $ssl_mutex {
+    $ssl_mutex_real = $ssl_mutex
```
igalic: why not just use [pick()](https://github.com/puppetlabs/puppetlabs-stdlib#pick) instead of this giant if/else?

DavidS:
Hi @nabam , this change looks helpful, but it would be great to use [pick()](https://github.com/puppetlabs/puppetlabs-stdlib#pick) instead of the `if` block, add a short note to the README explaining the new property and have a unit test at the end of https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/ssl_spec.rb ascertaining that the value passed in is landing in the file.

nabam:
@DavidS sorry, had no time to work on it, I'll fix that soon

bmjen:
Hi @nabam, can you reopen this PR when you get a chance to address the feedback provided? Thanks!

PR #1214, opened 2015-09-30
(MODULES-2649) Allow SetOutputFilter to be set on a directory.

traylenator:
e.g

```puppet
apache::vhost{'abc.example.org':
  directories => [
    { path              => '/path/to/filtered',
      set_output_filter => 'puppetdb-filter',
    },
  ],
}
```

* https://tickets.puppetlabs.com/browse/MODULES-2649
* http://httpd.apache.org/docs/current/mod/core.html#setoutputfilter

### DIFF
--- a/README.md
+++ b/README.md
@@ -2884,6 +2884,21 @@ Sets a `SetHandler` directive as per the [Apache Core documentation](http://http
     }
 ~~~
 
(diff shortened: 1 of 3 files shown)


igalic:
thank you @traylenator!

PR #1216, opened 2015-09-30
Change SSLProtocol in apache::vhost to be space separated

bmfurtado:
https://tickets.puppetlabs.com/browse/MODULES-2655

### DIFF
--- a/templates/vhost/_ssl.erb
+++ b/templates/vhost/_ssl.erb
@@ -26,7 +26,7 @@
   SSLProxyEngine On
   <%- end -%>
   <%- if @ssl_protocol -%>
-  SSLProtocol             <%= @ssl_protocol.compact.join(' ') %>
+  SSLProtocol             <%= [@ssl_protocol].flatten.compact.join(' ') %>
   <%- end -%>
   <%- if @ssl_cipher -%>
   SSLCipherSuite          <%= @ssl_cipher %>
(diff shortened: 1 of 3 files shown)


templates/vhost/_ssl.erb
```diff
@@ -26,7 +26,7 @@
   SSLProxyEngine On
   <%- end -%>
   <%- if @ssl_protocol -%>
-  SSLProtocol             <%= @ssl_protocol %>
+  SSLProtocol             <%= @ssl_protocol.compact.join(' ') %>
```
igalic: i'd go as far as to make it
```ruby
[@ssl_protocol].flatten.compact.join(' ')
```

HelenCampbell:
Hello @bmfurtado ,
Well spotted on this! Thank you for your work and @igalic for your input to help get the tests passing :)
It would be awesome if you could put a test in place to cover the array input and the readme could also be updated to mention that a single space separated string also works. After that I'll be happy to merge this in.

bmfurtado:
@HelenCampbell I've added an acceptance test and updated the docs, is this ok?

bmfurtado:
Disclaimer: For reasons I haven't had time to investigate I can't run the tests and it doesn't look like Travis runs acceptance tests (for some reason?) so I'm not really sure this works :)

igalic:
Let's see :P

PR #1217, opened 2015-10-04
Add configure option for the include pattern in vhost_enable dir

mpdude:
Since I've switched to using this module, it happened to me quite some times that I (for whatever reason) had to manually edit one of the files in the `vhost_enable_dir` and that after a server restart the changes seemed not to have any effect.

It turned out that my favorite editor created `backup~` files and that those were read by Apache *after* the original file.

If I remember correctly, at least Ubuntu uses an `*.conf` include pattern for this directory by default to prevent such mishaps. 

Of course, we cannot simply change this due to BC, but at least we should give users the option to configure this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1126,7 +1126,7 @@ Changes your virtual host configuration files' location. Default: determined by 
 - **Gentoo**: `/etc/apache2/vhosts.d`
 - **Red Hat**: `/etc/httpd/conf.d`
 
-##### `vhost_enable_pattern`
+##### `vhost_include_pattern`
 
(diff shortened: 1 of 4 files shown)


DavidS:
Good catch there! Could you just please rename it to `vhost_include_pattern`?

mpdude:
Here you go.

Do you regularly tag new releases so we can roll this out into production?

DavidS:
This was merged in #1223, I took care to preserve attribution. Thanks for your contribution.

mpdude:
Thank you David.

DavidS:
ah, and to answer your question: there are semi-regular releases. Usually when enough changes/new features have accumulated.

PR #1221, opened 2015-10-08
Install all modules before adding custom configs

mpdude:
When setting up a new machine from scratch and adding some `::apache::custom_config`s, I see these:

```
Info: /Stage[main]/Profiles::Apache/Apache::Custom_config[my-permissions-generic]/File[apache_my-permissions-generic]: Scheduling refresh of Exec[service notify for my-permissions-generic]
Info: /Stage[main]/Profiles::Apache/Apache::Custom_config[my-permissions-generic]/File[apache_my-permissions-generic]: Scheduling refresh of Exec[remove my-permissions-generic if invalid]
Notice: /Stage[main]/Profiles::Apache/Apache::Custom_config[my-permissions-generic]/Exec[service notify for my-permissions-generic]/returns: AH00526: Syntax error on line 22 of /etc/apache2/sites-enabled/000-some.site
Notice: /Stage[main]/Profiles::Apache/Apache::Custom_config[my-permissions-generic]/Exec[service notify for my-permissions-generic]/returns: Invalid command 'RewriteEngine', perhaps misspelled or defined by a module not included in the server configuration
Notice: /Stage[main]/Profiles::Apache/Apache::Custom_config[my-permissions-generic]/Exec[service notify for my-permissions-generic]/returns: Action '-t' failed.
Notice: /Stage[main]/Profiles::Apache/Apache::Custom_config[my-permissions-generic]/Exec[service notify for my-permissions-generic]/returns: The Apache error log may have more information.
```

What happens is that I have already put some custom virtual host in place that requires some module (e.g. mod_rewrite). Then, a `custom_config` is applied and triggers the config test. That test fails because the module (required by the virtual host) is not yet in place.

### DIFF
--- a/manifests/custom_config.pp
+++ b/manifests/custom_config.pp
@@ -49,12 +49,13 @@ define apache::custom_config (
   }
 
   if $ensure == 'present' and $verify_config {
-    exec { "service notify for ${name}":
+    exec { "syntax verification for ${name}":
       command     => $verify_command,
       subscribe   => File["apache_${name}"],
       refreshonly => true,
       notify      => Class['Apache::Service'],
       before      => Exec["remove ${name} if invalid"],
+      require     => Anchor['::apache::modules_set_up']
     }
 
     exec { "remove ${name} if invalid":
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -381,4 +381,8 @@ class apache (
       manage_docroot  => $default_ssl_vhost,
     }
   }
+
+  # This anchor can be used as a reference point for things that need to happen *after*
+  # all modules have been put in place.
+  anchor { '::apache::modules_set_up': }
 }
--- a/manifests/mod.pp
+++ b/manifests/mod.pp
@@ -162,4 +162,6 @@ define apache::mod (
       }
     }
   }
+
+  Apache::Mod[$name] -> Anchor['::apache::modules_set_up']
 }
--- a/spec/defines/custom_config_spec.rb
+++ b/spec/defines/custom_config_spec.rb
@@ -26,7 +26,7 @@ describe 'apache::custom_config', :type => :define do
         'content' => '# Test',
       }
     end
-    it { is_expected.to contain_exec("service notify for rspec").with({
+    it { is_expected.to contain_exec("syntax verification for rspec").with({
       'refreshonly' => 'true',
       'subscribe'   => 'File[apache_rspec]',
       'command'     => '/usr/sbin/apachectl -t',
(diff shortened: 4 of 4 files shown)


DavidS:
There is no reason for having `Exec[service notify for my-permissions-generic]`. That only leads to multiple restarts of apache, when only one is needed. Instead notify `Class['apache::service']`, which should already have all the proper relationships set up.

mpdude:
@DavidS I don't do this. I just do

```
apache::custom_config { 'my-permissions-generic': ... }
```

and the module does the rest. The problem is that at the time it does this (and runs the config test, also a feature internal to the module) the Apache modules are not installed yet.

I don't think this issue should be closed yet, could you please re-open it?

DavidS:
Sorry, I missed that part. I'll have another look.

DavidS:
The main problem with your proposed solution would be that the collector `<||>` will realize all virtual resources matching (see https://docs.puppetlabs.com/puppet/latest/reference/lang_virtual.html#realizing-with-a-collector).

This will kill any configuration relying on virtual apache::mod resources. I've spelunked a bit through the code and I think the best bet is to try to connect the mod packages from https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod.pp#L70 directly to the verification exec (https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/custom_config.pp#L52), probably best done using a tag. If you are in there, please also rename that exec to `syntax verification for ...`, because - damn - that confused me.

mpdude:
Not sure I get you right. We should connect not only the package (https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod.pp#L70) as it might not even be installed. We need something to refer to the ::apache::mod as a whole.

Any could you please give me an example how you would use a tag here?

mpdude:
@DavidS What about the current solution with a shared anchor? Do you think this might work?

DavidS:
This looks like a very nice idea! If you could just squash/rebase it into a single commit, please?

mpdude:
Done.

It also passed a quick check and provisioned a machine correctly. But of course, that might also be resource ordering luck :smile: .

DavidS:
If jenkins goes green, I'll merge it. You are in the best position to judge whether it solves your problem. I like the change and I'd be stupid to try to second guess you :-)

mpdude:
Then, as a German, I'll go and port hmlb/phpunit-vw to Puppet.

DavidS:
:-D

mpdude:
Too bad. I hate those tests that fail when you change a natural language string printed to the user.

DavidS:
I've yet to find a better way to ascertain that the *right* error is thrown :-/

DavidS:
Danke, daß Du Dir die Mühe gemacht hast das ordentlich umzusetzen!

mpdude:
Vielen Dank für Deine schnelle Reaktion und den Merge :)

PR #1228, opened 2015-10-16
RewriteLock support

wickedOne:
as addition to the [RewriteMap](https://github.com/puppetlabs/puppetlabs-apache/pull/900) feature i've added the option to define a [RewriteLock](http://httpd.apache.org/docs/2.2/mod/mod_rewrite.html#RewriteLock) for apache version <= 2.2 which is considered best practice when using a RewriteMap of type [prg](https://httpd.apache.org/docs/2.2/rewrite/rewritemap.html#prg)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -55,7 +55,7 @@ class apache (
   $conf_template          = $::apache::params::conf_template,
   $servername             = $::apache::params::servername,
   $pidfile                = $::apache::params::pidfile,
-  $rewrite_lock           = false,
+  $rewrite_lock           = undef,
   $manage_user            = true,
   $manage_group           = true,
   $user                   = $::apache::params::user,
@@ -297,6 +297,10 @@ class apache (
       default   => false
     }
 
+    if $rewrite_lock {
+      validate_absolute_path($rewrite_lock)
+    }
+
     # Template uses:
     # - $pidfile
     # - $user
@@ -318,6 +322,7 @@ class apache (
     # - $server_tokens
     # - $server_signature
     # - $trace_enable
+    # - $rewrite_lock
     file { "${::apache::conf_dir}/${::apache::params::conf_file}":
       ensure  => file,
       content => template($conf_template),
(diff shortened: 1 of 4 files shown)


igalic:
Thanks for catching this oversight! would you mind also adding a test and a doc entry?

wickedOne:
yes will do

wickedOne:
done…

manifests/init.pp
```diff
@@ -55,6 +55,7 @@
   $conf_template          = $::apache::params::conf_template,
   $servername             = $::apache::params::servername,
   $pidfile                = $::apache::params::pidfile,
+  $rewrite_lock           = false,
```
igalic: let's make this `undef` by default, since we're dealing with a `String` (path) here.

manifests/init.pp
```diff
@@ -317,6 +318,7 @@
     # - $server_tokens
     # - $server_signature
     # - $trace_enable
+    # - $rewrite_lock
```
igalic: we should [validate this variable](https://github.com/puppetlabs/puppetlabs-stdlib#validate_absolute_path),
```puppet
if $rewrite_lock {
  validate_absolute_path($rewrite_lock)
}
```

spec/classes/apache_spec.rb
```diff
@@ -388,6 +388,37 @@
         it { is_expected.to contain_file("/etc/httpd/conf/httpd.conf").with_content %r{^IncludeOptional "/etc/httpd/conf\.d/\*\.conf"$} }
       end
 
+      context "with Apache version < 2.4" do
+        let :params do
+          {
+            :apache_version => '2.2',
+            :rewrite_lock => '/etc/httpd/RewriteLock'
```
igalic: shouldn't this be somewhere where httpd has write acces?
(i sure hope you're not giving it write access to its own config ;)

igalic:
looks very good otherwise, thanks @wickedOne!

wickedOne:
thanks for pointing all that out @igalic 
made changes accordingly and ammended them to my last commit.

moved the lockfile to location ``/var/lock/subsys/rewrite-lock`` in ``spec/classes/apache_spec.rb`` as i wasn't able to find something of a best practice for this specific lockfile and as that's where the other lockfiles are stored… hope that's ok with you

igalic:
\o/ @wickedOne!

wickedOne:
thanx @igalic for the super quick merge! \,,/.

PR #1229, opened 2015-10-20
#2544 Allow multiple IP addresses per vhost

Benedikt1992:
To make a vhost reachable over 2 IP addresses we need to configure 2
similar vhosts which differ in the IP address. This change allows to use an array
of IPs.

Ticket: https://tickets.puppetlabs.com/browse/MODULES-2544

Since Release 0.9.0 we have a dependency on stdlib >= 2.4.x [CHANGELOG](https://github.com/puppetlabs/puppetlabs-apache/blob/master/CHANGELOG.md#backwards-incompatible-changes-1)
With the changes in this pull requests we need stdlib >= 4.0.0 since `suffix()`  is not available in Release 2.4.x [stdlib CHANGELOG](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/CHANGELOG.md#2013-03-26---richard-soderberg-rsoderbergmozillacom---400),  [Code](https://github.com/puppetlabs/puppetlabs-apache/pull/1229/files#diff-c918e8884948c0fae251b33071374150R343)

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -339,8 +339,8 @@ define apache::vhost(
 
   if $ip {
     if $port {
-      $listen_addr_port = "${ip}:${port}"
-      $nvh_addr_port = "${ip}:${port}"
+      $listen_addr_port = suffix(any2array($ip),":${port}")
+      $nvh_addr_port = suffix(any2array($ip),":${port}")
     } else {
       $listen_addr_port = undef
       $nvh_addr_port = $ip
@@ -364,13 +364,13 @@ define apache::vhost(
     if $ip and defined(Apache::Listen["${port}"]) {
       fail("Apache::Vhost[${name}]: Mixing IP and non-IP Listen directives is not possible; check the add_listen parameter of the apache::vhost define to disable this")
     }
-    if ! defined(Apache::Listen["${listen_addr_port}"]) and $listen_addr_port and $ensure == 'present' {
-      ::apache::listen { "${listen_addr_port}": }
+    if $listen_addr_port and $ensure == 'present' {
+      ensure_resource('apache::listen', $listen_addr_port)
     }
   }
   if ! $ip_based {
-    if ! defined(Apache::Namevirtualhost[$nvh_addr_port]) and $ensure == 'present' and (versioncmp($apache_version, '2.4') < 0) {
-      ::apache::namevirtualhost { $nvh_addr_port: }
+    if $ensure == 'present' and (versioncmp($apache_version, '2.4') < 0) {
+      ensure_resource('apache::namevirtualhost', $nvh_addr_port)
     }
   }
 
--- a/spec/acceptance/vhost_spec.rb
+++ b/spec/acceptance/vhost_spec.rb
@@ -190,6 +190,59 @@ describe 'apache::vhost define', :unless => UNSUPPORTED_PLATFORMS.include?(fact(
     end
   end
 
+  context 'new vhost with multiple IP addresses on port 80' do
+    it 'should configure one apache vhost with 2 ip addresses' do
+      pp = <<-EOS
+        class { 'apache':
+          default_vhost => false,
+        }
+        apache::vhost { 'example.com':
+          port     => '80',
+          ip       => ['127.0.0.1','::1'],
+          ip_based => true,
+          docroot  => '/var/www/html',
+        }
+        host { 'ipv4.example.com': ip => '127.0.0.1', }
+        host { 'ipv6.example.com': ip => '::1', }
+        file { '/var/www/html/index.html':
+          ensure  => file,
+          content => "Hello from vhost\\n",
+        }
+      EOS
+      apply_manifest(pp, :catch_failures => true)
+    end
+
+    describe service($service_name) do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+
+    describe file("#{$vhost_dir}/25-example.com.conf") do
+      it { is_expected.to contain '<VirtualHost 127.0.0.1:80 ::1:80>' }
+      it { is_expected.to contain "ServerName example.com" }
+    end
+
+    describe file($ports_file) do
+      it { is_expected.to be_file }
+      it { is_expected.to contain 'Listen 127.0.0.1:80' }
+      it { is_expected.to contain 'Listen ::1:80' }
+      it { is_expected.not_to contain 'NameVirtualHost 127.0.0.1:80' }
+      it { is_expected.not_to contain 'NameVirtualHost ::1:80' }
+    end
+
+    it 'should answer to ipv4.example.com' do
+      shell("/usr/bin/curl ipv4.example.com:80", {:acceptable_exit_codes => 0}) do |r|
+        expect(r.stdout).to eq("Hello from vhost\n")
+      end
+    end
+
+    it 'should answer to ipv6.example.com' do
+      shell("/usr/bin/curl ipv6.example.com:80", {:acceptable_exit_codes => 0}) do |r|
+        expect(r.stdout).to eq("Hello from vhost\n")
+      end
+    end
+  end
+
   context 'apache_directories' do
     describe 'readme example, adapted' do
       it 'should configure a vhost with Files' do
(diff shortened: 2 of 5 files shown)


manifests/vhost.pp
```diff
@@ -338,23 +338,24 @@
   }
 
   if $ip {
+    $_ip = any2array($ip)
     if $port {
-      $listen_addr_port = "${ip}:${port}"
-      $nvh_addr_port = "${ip}:${port}"
+      $listen_addr_port = parseyaml(inline_template('<% res="" %><% @_ip.each { |a_ip| res = res + "#{a_ip}:#{@port}," } %><%= res.split(",").to_yaml -%>'))
```
igalic: this feels dirty

manifests/vhost.pp
```diff
@@ -338,23 +338,24 @@
...
+      $nvh_addr_port = $listen_addr_port
```
igalic: while i know that this works, it's not really seen very often…

manifests/vhost.pp
```diff
@@ -338,23 +338,24 @@
...
+      $nvh_addr_port = $_ip
       if ! $servername and ! $ip_based {
         fail("Apache::Vhost[${name}]: must pass 'ip' and/or 'port' parameters for name-based vhosts")
       }
     }
   } else {
     if $port {
-      $listen_addr_port = $port
-      $nvh_addr_port = "${vhost_name}:${port}"
+      $listen_addr_port =any2array($port)
```
igalic: this needs a space after =

igalic:
anyway, i think what you really want is… something entirely different:

if we're talking about name-based VirtualHosts then:

```irc
<fajita> http://httpd.apache.org/docs/current/mod/core.html#namevirtualhost. (Deprecated in 2.4) You must specify only one NameVirtualHost directive, hopefully with an IP address  and port or a *:80. It says on which interfaces should Apache listen to name based vhosts requests, nothing more.
```

or are you talking about 2.2 / ip-based virtualhosts?

manifests/vhost.pp
timogoebel: Let's call this an interesting approach.. @igalic : Would you suggest using a custom function for this? Nevertheless, this could be done a lot easier with future_parser enabled...

igalic: we'll have a real hard time porting this module to 4.x because it's pretty much one of the most used ones out there… i don't really have a strategy there…

Benedikt1992:
@igalic as far as i see it does not matter if we have ip based or name-based vhosts. Let's say we have a webpage that should be available over IPv4 and IPv6 with vhosts. With the current apache Module we have to make 2 vhosts which just differ in the IP address. A better solution from my point of view is to make one ip based vhost which listens on two ip addresses. So instead of 

    <VirtualHost 192.168.1.1>
        DocumentRoot "/www/server"
        ServerName server.example.com
        ServerAlias server
        
        # Other directives here ...
    </VirtualHost>
    
    <VirtualHost 2001:DB8::1>
        DocumentRoot "/www/server"
        ServerName server.example.com
        ServerAlias server
    
        # Other directives here ...
    </VirtualHost>

we have 

    <VirtualHost 192.168.1.1 2001:DB8::1>
        DocumentRoot "/www/server"
        ServerName server.example.com
        ServerAlias server
    
        # Other directives here ...
    </VirtualHost>

With name-based vhosts and apache 2.2 we would also need  

    NameVirtualHost 192.168.1.1
    NameVirtualHost 2001:DB8::1

Apache 2.4 will automatically configure name-based vhosts if there is more than one `VirtualHost` definition with the same IP address

manifests/vhost.pp
Benedikt1992: @igalic Like @timogoebel said: Do you have another approach?

igalic:
@Benedikt1992 what i don't understand is why you can't just use `*:80`

Benedikt1992:
@igalic That works for name-based vhosts. With ip-based vhosts you have to declare every ip address explicitly. Otherwise it wouldn't be a ip-based vhost.

igalic:
yes, but you don't need NameVirtualHost for each of those.
the point of `NameVirtualHost x`is to tell httpd (<2.4) that following this point, match all hosts with the pattern `x` as name-based virtualhosts.

Please read these two https://httpd.apache.org/docs/2.2/vhosts/name-based.html
https://httpd.apache.org/docs/2.2/vhosts/ip-based.html

Benedikt1992:
NameVirtualHost is only set if `ip_based = false` and only once per IP address/port combination (because of `ensure_resource()`) [Code](https://github.com/Benedikt1992/puppetlabs-apache/commit/c662f43a1297c4068c181346dd6eb54b01f2bb03#diff-c918e8884948c0fae251b33071374150R372)

Basically the behaviour has not changed. It is just able to handle an array of IP address/port combinations.

Benedikt1992:
Is someone still working on this pull request?

igalic:
@Benedikt1992 it was a long weekend.
let's take a fresh look at this.

manifests/vhost.pp
igalic: this hole line can be broken up into one call to [suffix()](https://github.com/puppetlabs/puppetlabs-stdlib#suffix):
```puppet
$listen_addr_port = suffix($_ip, $port)
```

Benedikt1992: That looks better. But this needs stdlib >= 4.2.x. If the Changelog is right this Requirement will jump from 2.4.x to 4.2.x...

Benedikt1992: @igalic I've changed that line to suffix(). Togeth

spec/acceptance/vhost_spec.rb
```diff
@@ -190,6 +190,59 @@ class { 'apache': }
...
+      it { is_expected.to contain '<VirtualHost 127.0.0.1:80 ::1:80>' }
+      it { is_expected.to contain "ServerName example.com" }
+    end
+
+    describe file($ports_file) do
+      it { is_expected.to be_file }
+      it { is_expected.to contain 'Listen 127.0.0.1:80' }
+      it { is_expected.to contain 'Listen ::1:80' }
+      it { is_expected.not_to contain 'NameVirtualHost 127.0.0.1:80' }
+      it { is_expected.not_to contain 'NameVirtualHost ::1:80' }
```
igalic: i'm still confused about this bit here ^

templates/vhost/_file_header.erb
```diff
@@ -3,7 +3,7 @@
 # Managed by Puppet
 # ************************************
 
-<VirtualHost <%= @nvh_addr_port %>>
+<VirtualHost <%= [@nvh_addr_port].flatten.join(' ') %>>
```
igalic: you might also need a `.compact` as well

spec/acceptance/vhost_spec.rb
Benedikt1992: This test case generates an ip-based vhost. Ip-based vhosts don't have a `NameVirtualHost` directive. Probably we could delete this two tests scince it is not directly related to the changes. But it is neither false.

templates/vhost/_file_header.erb
Benedikt1992: I'm not sure if it is even possible that there is a nil in that array but it won't break something...

spec/acceptance/vhost_spec.rb
igalic: what i mean, when we want to use ip-based vhosts, then the namevirtualhost stanzas will create just more confusion. they shouldn't be created at all.

igalic:
@Benedikt1992 aside when the nvh issue, could you please [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits)…

also, if you want to have extra magick internet points, you can also change the git commit author to something github recognizes, or else add this email into your github settings.

igalic:
thanks @Benedikt1992!

Benedikt1992:
@igalic Thank you for merging. It is not possible to change the author afterwards, isn't it?

I will keep squashing in my mind

igalic:
not in git.
not once it's merged.
(well, we too can git push -f, to master, but that's not nice)
it's now there to stay forever.

PR #1233, opened 2015-10-25
Added support for additional directory include.

yarikdot:
Scenario: We have some scripts that generate dynamically vhosts and other configuration. We love using this plugin, which handles whole basic configuration (including default vhost). But virtualhosts (except the default one) are managed by something else.

Also helps to fix this problem:
https://tickets.puppetlabs.com/browse/MODULES-1321?jql=project%20%3D%20MODULES%20AND%20resolution%20%3D%20Unresolved%20AND%20component%20%3D%20apache%20ORDER%20BY%20priority%20DESC


yarikdot:
Usage:

class { 'apache':
  additional_include_dir => '/etc/apache2/after-vhost-include',
  additional_include_pattern => '*.conf'
}

this will create after-vhost-include directory if doesn't exist (to prevent configcheck to fail). And updates httpd.conf to include files '*.conf' files

luto:
I'm not a maintainer, but I'm currently hitting the same problem. Since there is a [simmilar per-vhost option](https://forge.puppetlabs.com/puppetlabs/apache#additional_includes) already, I guess this should follow the pattern of supplying `include_dir` and `include_pattern` within the same setting, inside an array, like so:

```
class { 'apache':
    additional_includes => [ '/etc/apache2/after-vhost-include/*' ],
}
```

Also needs tests, probably.

yarikdot:
It was first thing I tried. Unfortunately /etc/apache2/after-vhosting-include directory must exists otherwise apache configtest will fail.

    root@hosting:/etc/apache2# apache2ctl configtest
    apache2: Syntax error on line 51 of /etc/apache2/apache2.conf: Could not open config directory /etc/apache2/asdf: No such file or directory
    Action 'configtest' failed.
    The Apache error log may have more information.
    root@hosting:/etc/apache2# head -n 51 /etc/apache2/apache2.conf | tail -n 1
    IncludeOptional "/etc/apache2/asdf/*/"
    root@hosting:/etc/apache2#

Therefore the directory must be created before configtest is executed.

ckaenzig:
@yarikdot, @luto, not a maintainer here either. Why don't you generate your dynamic files in the standard directories (`/etc/apache2/conf.d`, `/etc/apache2/sites-available`, etc.) and set the `purge_*` parameters to the module to avoid puppet deleting your files?

You could also use `apache::custom_config` to include any files you want.

yarikdot:
@ckaenzig Please read the first comment. We want puppet to maintain some virtualhost as they are the same on all servers. But, we also have the application which generates other vhosts. Therefore, we are looking for a way how to combine these two technologies.

jonnytpuppet:
Hi There

This PR needs documentation (Readme) as well as tests before we can consider to merge this.

However, I'm not sure if this use case is generic enough to be included in the module.

@hunner @bmjen Any thoughts on this?

Thanks

yarikdot:
ok. Once it is confirmed that this feature will be merged, I fix the PR.

hunner:
@yarikdot Is it possible to just use `apache::vhost::custom` and pass in your vhost files verbatem to the `content` property? This will correctly notify the service to reload and such as well https://github.com/puppetlabs/puppetlabs-apache#define-apachevhostcustom

yarikdot:
Yes it can. I can imagine using:

  file { '/etc/apache2/sites-cpanel/':
    ensure => 'directory'
  }

  apache::vhost::custom { 'include':
      content => 'include /etc/apache2/sites-cpanel/',
      priority => '99',
  }

Good point. @hunner what do you think?

bmjen:
@yarikdot that actually would work and be a much simpler solution. Do you mind if I close this PR?

PR #1235, opened 2015-10-28
(MODULES-2651) Default document root update for Ubuntu 14.04 and Debian 8

abednarik:
For security reasons Debian changes default docroot to /var/www/html. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=730372

### DIFF
--- a/README.md
+++ b/README.md
@@ -907,7 +907,7 @@ class { 'apache':
 
 Sets the default [`DocumentRoot`][] location. Default: Determined by your operating system.
 
-- **Debian**: `/var/www`
+- **Debian**: `/var/www/html`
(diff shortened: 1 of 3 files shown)


janfabry:
Should the test at https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/apache_spec.rb#L27 also be updated to test for the new location? Currently it has Debian 6 as a context, so maybe a new subcontext with Debian 8 should be created?

abednarik:
thanks for bringing that to my attention. Added 2 tests cases.

igalic:
thanks @abednarik!

PR #1236, opened 2015-10-28
(MODULES-2756) Adding include ::apache so mkdir exec works properly

damonconway:
This fixes a compile error where the require on Exec["mkdir ${::apache::mod_dir}"] was failing because it couldn't resolve ${::apache::mod_dir}.

### DIFF
--- a/manifests/mod/deflate.pp
+++ b/manifests/mod/deflate.pp
@@ -12,6 +12,7 @@ class apache::mod::deflate (
     'Ratio'  => 'ratio'
   }
 ) {
+  include ::apache
   ::apache::mod { 'deflate': }
 
   file { 'deflate.conf':


manifests/mod/deflate.pp
```diff
@@ -12,6 +12,7 @@
     'Ratio'  => 'ratio'
   }
 ) {
+  include ::apache
   ::apache::mod { 'deflate': }
```
igalic: i'm confused.
shouldn't that line ^ error out, and tell you to first include apache before including any modules?
see: https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod.pp#L11-L13

damonconway: I was confused as well. It did not error out there. The error I got is
documented in the ticket.

"Error: Could not retrieve catalog from remote server: Error 400 on SERVER:
Invalid relationship: File[deflate.conf] { require => Exec[mkdir ] },
because Exec[mkdir ] doesn't seem to be in the catalog"

Even though in my profile::apache I have "include ::apache" already, and my
"profile::apache::cfg::ontrack" uses $::apache::httpd_dir just fine and is
where I do a "include ::apache::mod::deflate".

I came to the include ::apache solution from MODULES-1491.

It seems as if puppet is not correctly resolving the catalog. The other
solution I can think of is to include ::apache in
profile::apache::cfg::ontrack and use anchors to make sure the apache class
loads before apache::mod::deflate. Seems like a lot of work for the end
user when this should just work.

Thanks,
Damon

damonconway: Sorry, I forgot to say that I attached the files for profile::apache and
profile::apache::cfg::ontrack to the last email.

Damon

igalic:
@damonconway i don't think github copies attatchments from emails into comments…

damonconway:
Ok, I'll attach them here in the web interface....or not. Github won't let me attach them. I've attached them to the ticket.

Thanks,
Damon

bmjen:
@damonconway I'm not sure why this is necessary. Typically when using/declaring an apache mod, the apache class is also declared, like this:

~~~
class { 'apache': }
class { 'apache::mod::deflate': }
~~~

It has never been a pattern in our mods to allow them to include or instantiate an apache class. Perhaps I'm not understanding the usecase?

@igalic any thoughts on the need or usecase described here?

DavidS:
mod_deflate is also managed by `apache::default_mods`, so that may muddy the waters here.

@damonconway @igalic  I'd be really grateful for revisiting this.

damonconway:
In the ticket, I attached my profile classes so you can see how I'm calling
it. In profile::apache, I call "include ::apache". From
profile::apache::cfg::ontrack, I call "include ::profile::mod::deflate". In
my role::web class, I call "include ::profile::apache". I include
profile::apache::cfg::ontrack in a classes array in hiera.

I don't think it's a good plan to include apache in apache::mod::deflate,
but that's the only way I got it to work. Putting it in
::profile::apache::cfg::ontrack did not work.

https://tickets.puppetlabs.com/browse/MODULES-2756

The require that's failing is 'require => Exec["mkdir ${::apache::mod_dir}"],'.
For some reason, it cannot resolve ::apache::mod_dir until I put "include
::apache" into ::apache::mod::deflate.

Please see https://tickets.puppetlabs.com/browse/MODULES-1491 for another
example of this problem.

Damon

tphoney:
I am actively looking into this.

tphoney:
So this seems to be endemic throughout the mod classes. I am working on adding tests to show this, and add the appropriate fixes. 

Thanks for pointing this out.

PR #1239, opened 2015-10-28
Support itk on redhat

edestecd:


edestecd:
I need to fix the spec test that expect unsupported os...

Does anyone have any feedback on my approach?
You probably want this pull against master?

igalic:
@edestecd why is this a patch against 1.5.x?

edestecd:
@igalic i'm being lazy.  That's the version I am currently stuck at ;)

bmjen:
@edestecd Can you resubmit this fix against master? I'm going to close this PR, as we likely will not do another release from the 1.5.x branch.

PR #1240, opened 2015-10-29
(MODULES-2440) Allow installing additional php-related packages

yakatz:


igalic:
@yakatz frankly, i think this is the wrong place to do that.
you should be using a php module to manage those.

yakatz:
I just made a patch for the open ticket because it overlaps with our needs.

abednarik:
See this PR https://github.com/puppetlabs/puppetlabs-apache/pull/1247

bmjen:
@yakatz Thanks for the contribution, but I'm going to have to agree with @igalic. These packages are better suited to be managed by a php module and falls outside of the scope of this Apache module.

PR #1245, opened 2015-10-31
(MODULES-2682) Update Apache Configuration to set FilesMatch instead of AddHandler

abednarik:
The issue with this is that the extension handling behaviour of apache is not well known by most php developers, and many php scripts are open to security issues if this configuration is used (most commonly these scripts handle upload forms which white list image extensions). For example foo.php.jpg will be handled by php.
Many distro's no longer use AddHandler in their default config http://bazaar.launchpad.net/~ubuntu-branches/ubuntu/vivid/php5/vivid/view/head:/debian/php5.conf. The php manual also recommends avoiding it http://php.net/manual/en/install.unix.apache2.php#example-20.

### DIFF
--- a/spec/classes/mod/php_spec.rb
+++ b/spec/classes/mod/php_spec.rb
@@ -88,7 +88,7 @@ describe 'apache::mod::php', :type => :class do
       let :params do
         { :extensions => ['.php','.php5']}
       end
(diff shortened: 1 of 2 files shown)


igalic:
@abednarik would you mind doing a proper [rebase](https://www.git-scm.com/book/en/v2/Git-Branching-Rebasing)?

PR #1247, opened 2015-11-01
(MODULES-2440) Specify additional php packages for installation

abednarik:
Added an optional param to install php extra modules. Documentation and Tests cases added.


abednarik:
Related to https://github.com/puppetlabs/puppetlabs-apache/pull/1247

PR #1254, opened 2015-11-09
Fix typo about dynamic AddHandler/AddType

olivierHa:
Hello, 

The last commit had introduced a new feature about adding dynamic AddHandler/AddType.
There is a little typo error. The "right" syntax should be :

AddType text/html .shtml

Regards

Olivier


igalic:
thank you very much @olivierHa!

olivierHa:
You're welcome

PR #1255, opened 2015-11-09
Add missing parameters in mod_auth_kerb

olivierHa:
Hello, 

According to the http://modauthkerb.sourceforge.net/configure.html webpage,
the 3  following parameters are missing.

- KrbVerifyKDC
- KrbServiceName
- KrbSaveCredentials

Regards

Olivier

### DIFF
--- a/README.md
+++ b/README.md
@@ -2074,6 +2074,18 @@ Location of the Kerberos V5 keytab file. Not set by default.
 
 Strips @REALM from username for further use. Not set by default.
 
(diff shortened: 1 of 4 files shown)


igalic:
@olivierHa thank you very much for this patch!
would you mind [squashing](https://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits) these three commits down to one?

olivierHa:
it should be better.

PR #1256, opened 2015-11-11
(MODULES-2811) Add missing helper lines to spec files

alexharv074:
See the Jira ticket for info on the motivation for this change.


igalic:
nice, thanks @alexharv074!

PR #1257, opened 2015-11-11
(MODULES-2812) Fix deprecation warning in service_spec.rb

alexharv074:
See Jira ticket for the motivation for this change.


igalic:
thanks @alexharv074!

PR #1258, opened 2015-11-11
(MODULES-2813) Fix deprecation warning in spec_helper.rb

alexharv074:
See Jira ticket for the motivation for this change.


igalic:
thank you @alexharv074!

PR #1260, opened 2015-11-11
Add support for changing mod_nss listen port (vol 2)

rexcze:
I added ability to mod_nss to be able to change listen port from default 8443 to for example 443, which is default https(SSL) port. It is quite useful for us.


rexcze:
Thanks!

PR #1264, opened 2015-11-13
updating the test to use ipv4 addresses

tphoney:


igalic:
why's that?

tphoney:
@igalic the ipv6 addresses were causing issues on older OSes in the acceptance tests. Ubuntu 10/12. The test that was added was just to check multiple ip's. not an ipv6 test

igalic:
@tphoney given that we've reached IPv4 exhaustion, the high council has decreed

> that Operating Systems without IPv6 support shall be removed from all test matrices

> that Operating Systems without IPv6 support shall be removed from all provisioning software

thank you for your attention,

:womans_hat:

tphoney:
@igalic i agree, there should be separate IPv6 tests.

PR #1267, opened 2015-11-18
Leave require directive unmanaged

robertvargason:
Instead to default require to 'Require all granted' if no 'require'
parameter is set for a directory, leave require unmanaged in apache 2.4
if needed. This is useful for complex authentication/authorization
requirements which are handled in a custom fragment.

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -978,5 +978,72 @@ describe 'apache::vhost', :type => :define do
       let :facts do default_facts end
       it { expect { is_expected.to compile }.to raise_error }
     end
+    context 'default of require all granted' do
+      let :params do
+        {
+          'docroot'         => '/var/www/foo',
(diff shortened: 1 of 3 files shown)


spec/defines/vhost_spec.rb
```diff
@@ -184,6 +184,15 @@
...
             },
+            {
+              'path'     => '/var/www/files_rag',
+              'provider' => 'files',
+            },
+            {
+              'path'     => '/var/www/files_nrag',
+              'provider' => 'files',
+              'require'  => 'unmanaged',
+            },
```
igalic: this seems needlessly duplicated, and is unused

igalic:
thanks @robertvargason!

ubellavance:
Thanks for this PR, I spent a lot of time searching to find how not to have the `'Require all granted'` in my Limit block. Would there be a better way to document it?

PR #1268, opened 2015-11-18
(MODULES-2834) Support SSLProxyCheckPeerCN and SSLProxyCheckPeerName …

traylenator:
…settings

Allows SSLProxyCheckPeerCN and SSLProxyCheckPeerName to be set on
an SSL enabled vhost.

```puppet
apache::vhost{'foo':
   ssl_proxy_check_peer_cn   => 'on',
   ssl_proxy_check_peer_name => 'on'
}
```

results in

```
SSLProxyCheckPeerCN  on
SSLProxyCheckPeerName on
```

apache configuration with in a vhost.

* https://tickets.puppetlabs.com/browse/MODULES-2834
* http://httpd.apache.org/docs/current/mod/mod_ssl.html#sslproxycheckpeercn
* http://httpd.apache.org/docs/current/mod/mod_ssl.html#sslproxycheckpeername

### DIFF
--- a/README.md
+++ b/README.md
@@ -3164,6 +3164,15 @@ Sets the [SSLProxyMachineCertificateFile](http://httpd.apache.org/docs/current/m
     }
 ~~~
 
+##### `ssl_proxy_check_peer_cn`
+ 
+Sets the [SSLProxyMachinePeerCN](http://httpd.apache.org/docs/current/mod/mod_ssl.html#sslproxycheckpeercn) directive, which specified whether the remote server certificate's CN field is compared against the hostname of the request URL .  Defaults to 'undef'.
+
+
+##### `ssl_proxy_check_peer_name`
+ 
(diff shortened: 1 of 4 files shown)


igalic:
i'd rather we used [bool2httpd](https://github.com/puppetlabs/puppetlabs-apache/blob/master/lib/puppet/parser/functions/bool2httpd.rb) here, but…

with the benefit of over a year of experience, i'm now seeing @hunner's points against this particular function being a pass-through…

perhaps what we need, is also a validate_httpd_bool(), which takes an optional argument à la validate_re().

README.md
```diff
@@ -3164,6 +3164,15 @@ Sets the [SSLProxyMachineCertificateFile](http://httpd.apache.org/docs/current/m
     }
 ~~~
 
+##### `ssl_proxy_check_peer_cn`
```
bmjen: Can you move this above the section `ssl_proxy_machine_cert` as to keep the params in alphabetical order?

bmjen:
I'm ok with this change provided the readme update. @igalic do you still have any objections? Your last comment was unclear.

sergehardy:
SSLProxyCheckPeerName was not a valid directive in Apache v2.2, so using this option leads to an error in that case.

http://httpd.apache.org/docs/2.2/mod/mod_ssl.html

PR #1270, opened 2015-11-19
add ability to overide file name generation in custom_config

karmix:
We discussed this a couple times in modules triage.  I proposed it in in comments in PR #1201, but haven't seen any response from the pull requester.  I'm putting it up so we can review it separately.


bmjen:
:+1: 

Resolves #1201

PR #1271, opened 2015-11-19
Add ::apache::vhost::custom

pabelanger:
Here we are adding a thin wrapper to apache::custom_config but setting
default specific to the vhost apache folder.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>

### DIFF
--- a/manifests/vhost/custom.pp
+++ b/manifests/vhost/custom.pp
@@ -1,0 +1,37 @@
+# See README.md for usage information
+define apache::vhost::custom(
+  $content,
+  $ensure = 'present',
+  $priority = '25',
+) {
+  include ::apache
+
+  ## Apache include does not always work with spaces in the filename
+  $filename = regsubst($name, ' ', '_', 'G')
+
+  ::apache::custom_config { $filename:
+    ensure   => $ensure,
+    confdir  => $::apache::vhost_dir,
+    content  => $content,
+    priority => $priority,
+  }
+
+  # NOTE(pabelanger): This code is duplicated in ::apache::vhost and needs to
+  # converted into something generic.
+  if $::apache::vhost_enable_dir {
+    $vhost_symlink_ensure = $ensure ? {
+      present => link,
+      default => $ensure,
+    }
+
+    file { "${priority}-${filename}.conf symlink":
+      ensure  => $vhost_symlink_ensure,
+      path    => "${::apache::vhost_enable_dir}/${priority}-${filename}.conf",
+      target  => "${::apache::vhost_dir}/${priority}-${filename}.conf",
+      owner   => 'root',
+      group   => $::apache::params::root_group,
+      mode    => '0644',
+      require => Apache::Custom_config[$filename],
+    }
+  }
+}
--- a/spec/defines/vhost_custom_spec.rb
+++ b/spec/defines/vhost_custom_spec.rb
@@ -1,0 +1,99 @@
+require 'spec_helper'
+
+describe 'apache::vhost::custom', :type => :define do
+  let :title do
+    'rspec.example.com'
+  end
+  let :default_params do
+    {
+      :content => 'foobar'
+    }
+  end
+  describe 'os-dependent items' do
+    context "on RedHat based systems" do
+      let :default_facts do
+        {
+          :osfamily               => 'RedHat',
+          :operatingsystemrelease => '6',
+          :operatingsystem        => 'RedHat',
+          :concat_basedir         => '/dne',
+          :id                     => 'root',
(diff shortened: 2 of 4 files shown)


manifests/vhost/custom.pp
```diff
@@ -0,0 +1,38 @@
...
+      present => link,
+      default => $ensure,
+    }
+
+    file { "${priority}-${filename}.conf symlink":
+      ensure  => $vhost_symlink_ensure,
+      path    => "${::apache::vhost_enable_dir}/${priority}-${filename}.conf",
+      target  => "${::apache::vhost_dir}/${priority}-${filename}.conf",
+      owner   => 'root',
+      group   => $::apache::params::root_group,
```
asasfu: This doesn't look like it supports the same root group override that the base apache class does.  Is that the plan?  Also what about the user for managing it?  Would it always be 'root' even if overridden in ::apache?

pabelanger: So, honestly I don't know the answer.  For the most part, this is copied verbatim from apache::vhost, so I would say the issue is there too.

Moving forward, I think this code should be broken out from both apache::vhost and apache::vhost::custom into a shared class.  Making bug fixes there too.

I didn't want to do that here, as to not make things more complicated.

asasfu: Good point.  After some reading, it was initially included as root/root and then expanded out to support different root_groups based on distros in this commit https://github.com/puppetlabs/puppetlabs-apache/commit/b87ce2d598e729729a5ef4f7b5c88230f9ba7b07
So it looks like vhost conf files are always owned by root as far as the apache module is concerned?  I have no idea if that's correct as far as Apache itself is concerned or not but it seems to be working so far.  So I guess par for course is the correct method to follow on this update.

nibalizer:
Unfortunately we weren't running apache in pcci so I don't have a baseline to compare this to. Also now that we are running it everything is failing, so that's not ideal.

nibalizer:
This code looks good to me +1

nibalizer:
I would like to see us have beaker tests for it but not critical.

pabelanger:
I can try and stab at breaker tests, but never done one before.  Also, I don't have the infra to really test it.

yrobla:
Initially looks good to me. I would like to see a common class for apache::vhost and this custom module in the future, to don't repeat logic.

manifests/vhost/custom.pp
```diff
@@ -0,0 +1,38 @@
...
+      mode    => '0644',
+      require => Apache::Custom_config[$filename],
+      notify  => Class['apache::service'],
```
bmjen: So.. I believe in the `apache::custom_config` class, there is a conditional that will set the notify if `$verify_config` is `true`. Do we want to adhere to that here?

Also corollary to that... since this is merely a wrapper around `apache::custom_config` should the notify be omitted here? So that you don't a) trigger an unwanted refresh or b) trigger a secondary refresh. Could there could be some idempotency issues with that as well?

spec/defines/vhost_custom_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+          :path                   => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin',
+          :is_pe                  => false,
+        }
+      end
+      let :params do default_params end
+      let :facts do default_facts end
+      it { is_expected.to contain_file("apache_rspec.example.com").with(
+        :ensure  => 'present',
+        :content => 'foobar',
+        :path   => '/etc/apache2/vhosts.d/25-rspec.example.com.conf',
```
bmjen: Rocket misalignment here.

spec/defines/vhost_custom_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+          :path                   => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+          :is_pe                  => false,
+        }
+      end
+      let :params do default_params end
+      let :facts do default_facts end
+      it { is_expected.to contain_file("apache_rspec.example.com").with(
+        :ensure  => 'present',
+        :content => 'foobar',
+        :path   => '/etc/apache2/sites-available/25-rspec.example.com.conf',
```
bmjen: Rocket misalignment here.

manifests/vhost/custom.pp
pabelanger: I'm not sure what we want to do here honestly.  I can drop the notify here, this was a copy and paste from vhost.pp.

PR #1274, opened 2015-11-25
(MODULES-2863) Set SSLProxy directives even if ssl is false

ckaenzig:
RewriteRules and ProxyPass directives can require SSLProxy*
configurations even if SSLEngine is not enabled.

This PR should be updated with options added in #1268 once
that one is merged.


igalic:
@ckaenzig ping!

ckaenzig:
@igalic PR rebased and updated. Thanks!

igalic:
wh00t!

PR #1275, opened 2015-11-26
Add request headers on a per directory basis.

grim8634:
Add request headers on a per directory basis, rather than just globally for the vhost.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2943,6 +2943,19 @@ apache::vhost { 'sample.example.net':
 
 `php_admin_value` sets the value of the directory, and `php_admin_flag` uses a boolean to configure the directory. Further information can be found [here](http://php.net/manual/en/configuration.changes.php).
 
+###### `request_headers`
+
+Adds lines for [RequestHeader](http://httpd.apache.org/docs/current/mod/mod_headers.html#requestheader) directives.
+
+~~~ puppet
+apache::vhost { 'sample.example.net':
+  docroot     => '/path/to/directory',
+  directories => {
+    path    => '/path/to/directory',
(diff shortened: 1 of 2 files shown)


igalic:
@grim8634 could you plz document this?
Also: do we need to include mod_header for this?

grim8634:
@igalic Updated docs thanks.
Yes, it requires mod_headers. There's no automatic checking it's loaded anywhere is there?

igalic:
@grim8634 here's how we do it at vhost level: https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/vhost.pp#L440-L445

DavidS:
Unless anybody has a better idea, I'd suggest using a inline_template and some ruby to check whether $directories has request headers and add a check like @igalic  suggested to include the mod_headers conditionally.

igalic:
@DavidS can you imagine what kind of face i'm making rn?

DavidS:
@igalic probably something involving rainbows and buckets?

jonnytpuppet:
@grim8634 I think your Readme is wrong. The property is called "headers" but in the template logic it's "request_headers". Am i missing something here?

Also, your tests need to pass before we can merge this.

Thanks!

bmjen:
Hi @grim8634 has there been any progress on addressing the feedback for this PR?

tphoney:
Hi @grim8634 , thanks for the work you have put into this pull request. However due to inactivity and and the comments not being addressed this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request.

PR #1277, opened 2015-11-26
(MODULES-2682) Update Apache Configuration to use FilesMatch instead …

DavidS:
…of AddHandler for PHP Files.

The issue with this is that the extension handling behaviour of apache is
not well known by most php developers, and many php scripts are open to
security issues if this configuration is used (most commonly these scripts
handle upload forms which white list image extensions). For example
foo.php.jpg will be handled by php.

Many distro's no longer use AddHandler in their default config:
  http://bazaar.launchpad.net/~ubuntu-branches/ubuntu/vivid/php5/vivid/view/head:/debian/php5.conf.

The PHP manual also recommends avoiding it:
  http://php.net/manual/en/install.unix.apache2.php#example-20

This is based on Alejandro Bednarik's <alejandro.bednarik@gmail.com> original
fix, I added proper regex escaping, and a changelog entry. All bugs are mine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## UNRELEASED
+### Summary
+
+TODO
+
+###
+
+#### Security
+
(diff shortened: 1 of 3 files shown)


abednarik:
Thanks @DavidS

PR #1279, opened 2015-11-27
(MODULES-2440) Allow additional php packages installation

abednarik:
Related to https://tickets.puppetlabs.com/browse/MODULES-2440


igalic:
i still don't agree with doing this through puppetlabs-apache, and think it should be done through a php module. Such as the approved [mayflower/php](https://forge.puppetlabs.com/mayflower/php)

HelenCampbell:
Hey @abednarik , Thank you for your contribution, however I agree with @igalic in that this would be best done through a php module itself. Closing PR.

PR #1283, opened 2015-12-01
changed typo

rexcze:
After changing variable nss_port to port in manifest file I did not changed port variable on "VirtualHost" line. Fixed


igalic:
how come this works in our tests?

rexcze:
Maybe because there are no test for mod_nss?

igalic:
oh, right… i even opened two bug on that: https://tickets.puppetlabs.com/browse/MODULES-2809 && https://tickets.puppetlabs.com/browse/MODULES-2808

igalic:
thank you @rexcze!

PR #1284, opened 2015-12-01
Fixes rspec syntax error in itk_spec wqtest.

bmjen:


DavidS:
This is not enough, as the spec_helper_acceptance logic to exclude ipv6 tests is the wrong way around and is currently ONLY running ipv6 tests

PR #1293, opened 2015-12-08
enable setting LimitRequestFieldSize globally as it does not actually…

KlavsKlavsen:
… work to increase it, inside a vhost. We tested with Kerberos keys (which is why we need to increase - as they can be upto 12000 bytes long.. so a default of 8190 doesn't work :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@
 [`KeepAliveTimeout`]: http://httpd.apache.org/docs/current/mod/core.html#keepalivetimeout
 [`keepalive` parameter]: #keepalive
 [`keepalive_timeout`]: #keepalive_timeout
+[`limitreqfieldsize`]: https://httpd.apache.org/docs/current/mod/core.html#limitrequestfieldsize
 
 [`lib`]: #lib
 [`lib_path`]: #lib_path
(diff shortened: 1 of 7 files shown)


igalic:
@KlavsKlavsen from [httpd's documentation](https://httpd.apache.org/docs/current/mod/core.html#limitrequestfieldsize)

>  When name-based virtual hosting is used, the value for this directive is taken from the default (first-listed) virtual host best matching the current IP address and port combination.

If your default (first listed) vhost doesn't have any settings, it'll use the Server's default setting, if the server hasn't changed it, it'll default to 8190.

at any rate, i agree, that this is more sensible to put in init.pp rather than vhost.pp

KlavsKlavsen:
@igalic Thank you. my parser did NOT quite catch that in the documentation ;) We just confirmed this however, by testing and found that it works when we set it in our default vhost (so we can avoid forking apache upstream module to set this - until puppetlabs-apache supports it as a general setting.

igalic:
@KlavsKlavsen cool

would you mind extending the documentation, and adding a test?
as well, as, perhaps, removing it from the vhost, since it can be really tricky to get right

my recommendation is to use the [github flow](https://guides.github.com/introduction/flow/index.html) workflow, if you want to make your life / contribution easier, with follow-up requests from me and my fellow colleagues ;)

KlavsKlavsen:
I usually create a branch - when I find I need to do other PR's.. but yeah.. I should probably start with doing that, just to be sure :)

I'm running "rake spec" now - to test my updated test procedure.. damn it takes time.. :)

igalic:
you can select only a few tests with rake rspec spec/classes/apache_spec.rb rspec spec/defines/vhost_spec.rb

KlavsKlavsen:
I've added the new test and removed the vhost option. the test succeeds:
1227 examples, 0 failures

igalic:
thank you @KlavsKlavsen
n.b.: this will need a minor(?) version bump.
Any comments from @puppetlabs here?

igalic:
thanks @KlavsKlavsen, thanks @tphoney for review

PR #1296, opened 2015-12-09
	Added support for modsecurity parameter SecPcreMatchLimit and SecPcr…

whotwagner:
…eMatchLimitRecursion

Set default PcreMatchLimit to 1000

fixed default variables for secpcrematchlimit(recursion) in params.pp

### DIFF
--- a/manifests/mod/security.pp
+++ b/manifests/mod/security.pp
@@ -3,6 +3,8 @@ class apache::mod::security (
   $activated_rules       = $::apache::params::modsec_default_rules,
   $modsec_dir            = $::apache::params::modsec_dir,
   $modsec_secruleengine  = $::apache::params::modsec_secruleengine,
+  $secpcrematchlimit = $::apache::params::secpcrematchlimit,
+  $secpcrematchlimitrecursion = $::apache::params::secpcrematchlimitrecursion,
   $allowed_methods       = 'GET HEAD POST OPTIONS',
   $content_types         = 'application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf',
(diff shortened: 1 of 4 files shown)


tphoney:
hi @whotwagner this change looks ok, but it will definitely need a readme update to explain the new parameter

whotwagner:
Hello, I changed the README.md and also set the default-values to modsecurity's default-values as it is described here https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SecPcreMatchLimit.

manifests/mod/security.pp
```diff
@@ -3,6 +3,8 @@
   $activated_rules       = $::apache::params::modsec_default_rules,
   $modsec_dir            = $::apache::params::modsec_dir,
   $modsec_secruleengine  = $::apache::params::modsec_secruleengine,
+  $modsec_secpcrematchlimit = $::apache::params::modsec_secpcrematchlimit,
+  $modsec_secpcrematchlimitrecursion = $::apache::params::modsec_secpcrematchlimitrecursion,
```
igalic: do we really need the `$modspec_` prefix within the mod::security module?

whotwagner: good question, the parameter-name is long enough without the prefix. I used the prefix because all other  parameter had this prefix too.  i thought it might be some kind of naming convention.

igalic: @whotwagner it seems to vary from module to module :S

whotwagner:
I changed(shortened) the variable-names and made a rebase again

tphoney:
Thanks for your work.

PR #1300, opened 2015-12-15
Skip fastcgi acceptance testing for Centos 7

pabelanger:
Since we don't have any packaging for it.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>


spec/acceptance/vhost_spec.rb
```diff
@@ -1326,7 +1326,8 @@ class { 'apache': }
   end
 
   # So what does this work on?
```
igalic: re that comment ^ what ever it works on, doesn't matter
you should be using mod_fcgid ;)

PR #1301, opened 2015-12-15
fix validation error when empty array is passed as rewrites parameter

timogoebel:
Manifest compilation fails when an empty array is passed as rewrites
parameter to vhost defined type.
This commit makes the code to compile nicely.


bmjen:
@timogoebel Thanks for the contribution! I'm ignoring the pcci failures b/c they are fastcgi related and those have been fixed in master.

PR #1302, opened 2015-12-15
Limit fastcgi testing to Debian osfamily

pabelanger:
Signed-off-by: Paul Belanger <pabelanger@redhat.com>

PR #1303, opened 2015-12-16
(maint) fixes the fastcgi test for debian and older ubuntu releases

bmjen:
For both Debian and older versions of Ubuntu, non-default repos are required to install the libapache2-mod-fastcgi package. I think the previous code worked on Ubuntu trusty, because in newer versions of Ubuntu, multiverse is enabled by default (according to the interwebs).

### DIFF
--- a/spec/acceptance/vhost_spec.rb
+++ b/spec/acceptance/vhost_spec.rb
@@ -1331,10 +1331,43 @@ describe 'apache::vhost define' do
     describe 'fastcgi' do
       it 'applies cleanly' do
         pp = <<-EOS
-          if $::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '10.04') >= 0 {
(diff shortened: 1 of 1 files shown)


bmjen:
@pabelanger @nibalizer this is to replace the fastcgi PR that didn't quite work from #1302

nibalizer:
:shipit:

bmjen:
Tested locally on Debian 6, 7, 8 and Ubuntu 10.04, 12.04, and 14.04 all have passed.

spec/acceptance/vhost_spec.rb
```diff
@@ -1348,7 +1379,7 @@ class { 'apache::mod::fastcgi': }
             fastcgi_dir    => '/tmp/fast',
           }
         EOS
-        apply_manifest(pp, :catch_failures => true)
+        apply_manifest(pp, :catch_failures => true, :acceptable_exit_codes => [0, 6])
```
DavidS: exit code 6 means changes AND failures. this doesn't sound right.

tphoney:
:+1:

PR #1305, opened 2015-12-17
EL7 uses conf.modules.d directory for modules.

jasonhancock:
This is briefly mentioned in [MODULES-1734](https://tickets.puppetlabs.com/browse/MODULES-1734), but EL7 uses the conf.modules.d directory for module configuration.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -62,7 +62,10 @@ class apache::params inherits ::apache::version {
     $server_root          = '/etc/httpd'
     $conf_dir             = "${httpd_dir}/conf"
     $confd_dir            = "${httpd_dir}/conf.d"
-    $mod_dir              = "${httpd_dir}/conf.d"
+    $mod_dir              = $::apache::version::distrelease ? {
+      '7'     => "${httpd_dir}/conf.modules.d",
+      default => "${httpd_dir}/conf.d",
+    }
     $mod_enable_dir       = undef
     $vhost_dir            = "${httpd_dir}/conf.d"
     $vhost_enable_dir     = undef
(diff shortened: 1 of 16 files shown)


manifests/params.pp
```diff
@@ -62,7 +62,10 @@
     $server_root          = '/etc/httpd'
     $conf_dir             = "${httpd_dir}/conf"
     $confd_dir            = "${httpd_dir}/conf.d"
-    $mod_dir              = "${httpd_dir}/conf.d"
+    $mod_dir              = $::apache::version::distrelease ? {
+      '7'     => "${httpd_dir}/conf.modules.d",
+      default => "${httpd_dir}/conf.d",
```
igalic: this should be using versioncmp()

bmjen: Agree, for the sake of consistency and maintainability.

bmjen:
@jasonhancock Are you able to address the feedback left by @igalic and also squash the commits?

manifests/params.pp
jasonhancock: If you want to talk consistency, then look further down in the file (line 101) where it's using the same syntax I'm using.

bmjen:
@jasonhancock you're right. Apache in general needs some cleanup. Do you mind squashing? I'll investigate the pcci failures.

bmjen:
@jasonhancock it appears the acceptance test failures on centos7 are related to this change.

jasonhancock:
@bmjen commits squashed and my test failures fixed. I'm not sure I understand the Ubuntu 14.04 fastcgi failure. Looks like it's not finding the libapache2-mod-fastcgi package, but I was unable to repoduce the issue locally on a 14.04 vagrant VM.

bmjen:
@jasonhancock Thanks! Ignore the fastcgi errors those are fixed in master.

PR #1306, opened 2015-12-17
Put headers and request headers before proxy

quixoten:
I have scenario where I need the request headers template to be rendered before the proxy template so I can remove an Authorization header before it goes across the proxy, e.g., 

```
## Request header rules
## as per http://httpd.apache.org/docs/2.2/mod/mod_headers.html#requestheader
RequestHeader unset Authorization

## Proxy rules
...
```

When the `RequestHeader unset` directive is after the proxy section, the header is sent to the proxy.

This change allows the order parameter of the request headers template to be changed without affecting current behavior.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -677,6 +677,26 @@ define apache::vhost(
   }
 
   # Template uses:
+  # - $headers
+  if $headers and ! empty($headers) {
+    concat::fragment { "${name}-header":
+      target  => "${priority_real}${filename}.conf",
+      order   => 140,
+      content => template('apache/vhost/_header.erb'),
+    }
+  }
+
+  # Template uses:
+  # - $request_headers
+  if $request_headers and ! empty($request_headers) {
+    concat::fragment { "${name}-requestheader":
+      target  => "${priority_real}${filename}.conf",
+      order   => 150,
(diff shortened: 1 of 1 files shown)


dteirney:
I think have run into this issue but never spent the time to figure out why. Is there any downside (other than an upgrade will change files) to having the request headers always before the proxy section?

The current API of a number to change it seems related to quite internal implementation details.

quixoten:
@dteirney, I don't know of any downsides to explicitly re-ordering it, but I'm not deeply familiar with apache so I started with the safest route. I have no issues with changing it though.

igalic:
Request/Headers should be earlier on in the vhost.

quixoten:
I re-ordered so that the headers and request headers templates are rendered just before the proxy template.

igalic:
@quixoten could you [squash](https://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits) your changes down to one?
(and from what i gather, a [rebase](https://www.git-scm.com/book/en/v2/Git-Branching-Rebasing) should guarantee that pcci will make trusty actually pass)

quixoten:
Squashed.

igalic:
merged! thanks @quixoten!

PR #1308, opened 2015-12-18
Add X-Forwarded-For into log_formats defaults

mpolenchuk:
In order to log the X-Forwarded-For client ip address or the real
client ip address add LogFormats default directive.

### DIFF
--- a/templates/httpd.conf.erb
+++ b/templates/httpd.conf.erb
@@ -77,6 +77,9 @@ LogFormat "%{Referer}i -> %U" referer
 <% unless @log_formats.has_key?('agent') -%>
 LogFormat "%{User-agent}i" agent
 <% end -%>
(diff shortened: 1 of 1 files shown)


HelenCampbell:
Hey @mpolenchuk , thank you for your contribution. However before I merge this all tests would need to have passed, and this change would also require a documentation update.

bmjen:
@mpolenchuk @HelenCampbell The pcci-trusty test failure can be ignored, I've already fixed the fastcgi problem in master. I'll merge as everything else looks ok.

antaflos:
@mpolenchuk shouldn't that be `%>s` instead of just `%s`? Both `common` and `combined` use `%>s` so `forwarded` should follow suit, no?

mpolenchuk:
@antaflos in terms of sameness yes. I've overlooked it. Thank you.

PR #1309, opened 2015-12-18
Support the mod_proxy ProxPassReverseCookieDomain directive  

occelebi:
Hey,

I've added ProxyPassReverseCookieDomain directive in mod_proxy.

Regards,
Cem

### DIFF
--- a/templates/vhost/_proxy.erb
+++ b/templates/vhost/_proxy.erb
@@ -20,7 +20,12 @@
   <Location <%= proxy['path']%>>
   <%- if not proxy['reverse_cookies'].nil? -%>
     <%- Array(proxy['reverse_cookies']).each do |reverse_cookies| -%>
-    ProxyPassReverseCookiePath <%= reverse_cookies['path'] %> <%= reverse_cookies['url'] %>
+      <%- if reverse_cookies['path'] -%>
+        ProxyPassReverseCookiePath <%= reverse_cookies['path'] %> <%= reverse_cookies['url'] %>
+      <%- end -%>
+      <%- if reverse_cookies['domain'] -%>
+        ProxyPassReverseCookieDomain <%= reverse_cookies['domain'] %> <%= reverse_cookies['url'] %>
+      <%- end -%>
     <%- end -%>
   <%- end -%>
   <%- if proxy['reverse_urls'].nil? -%>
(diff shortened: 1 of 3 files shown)


bmjen:
@occelebi Thanks for the contribution! This PR looks good. It needs a squash, but also.. how come you reviewed the spec test? It would be great if you got that test working so we can validate that the template gets populated correctly.

occelebi:
@bmjen, I got it squashed. However I do not know how to write proper test for this case as I do not know ruby much.

templates/vhost/_proxy.erb
```diff
@@ -20,7 +20,12 @@
   <Location <%= proxy['path']%>>
   <%- if not proxy['reverse_cookies'].nil? -%>
     <%- Array(proxy['reverse_cookies']).each do |reverse_cookies| -%>
-    ProxyPassReverseCookiePath <%= reverse_cookies['path'] %> <%= reverse_cookies['url'] %>
+     <%- if reverse_cookies['path'] -%>
```
bmjen: This is misaligned, can you add a space here.

bmjen:
@occelebi It's pretty simple to add a test case to an existing test. All we're expecting is that if a parameter is set, that the code does what it's intended.

Basically you can add another entry here: https://github.com/puppetlabs/puppetlabs-apache/blob/cf84e97e6a75dfaa9195d98b0917bfd2168badd8/spec/defines/vhost_spec.rb#L226-L229 with the `domain` key.

~~~
              'reverse_cookies' => [
                {
                  'path' => '/a',
                  'url'    => 'http://backend-a/',
                },
                {
                  'domain' => 'foo',
                  'url'         => 'http://foo',
               }
             ],
~~~

Then you can add another line below here: https://github.com/puppetlabs/puppetlabs-apache/blob/cf84e97e6a75dfaa9195d98b0917bfd2168badd8/spec/defines/vhost_spec.rb#L432-L433 to match the string output you'd expect based on what you set the domain to.

~~~
      it { is_expected.to contain_concat__fragment('rspec.example.com-proxy').with_content(
              /ProxyPassReverseCookieDomain\s+foo\s+http:\/\/foo/) }
~~~

occelebi:
@bmjen Thank you for your guidance. I've pushed new version.

But I'm still in need of clarification for test of ProxyPassReverseCookiePath:

              'reverse_cookies' => [
                {
                  'path' => '/a',
                  'url'    => 'http://backend-a/',
                },

According to that I expected something like that:

       it { is_expected.to contain_concat__fragment('rspec.example.com-proxy').with_content(
              /ProxyPassReverseCookiePath\s+\/a\s+http:\/\/backend-a\//) }

bmjen:
Thanks @occelebi ! That test has a loose regex match and passes because it only matches up to the "http://", your suggestion would have been better.

PR #1311, opened 2015-12-18
SSLProxyVerify

occelebi:
I've added SSLProxyVerify directive in mod ssl.

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -154,6 +154,7 @@ describe 'apache::vhost', :type => :define do
           'ssl_verify_depth'            => '3',
           'ssl_options'                 => '+ExportCertData',
           'ssl_openssl_conf_cmd'        => 'DHParameters "foo.pem"',
+          'ssl_proxy_verify'            => 'require',
           'ssl_proxy_check_peer_cn'     => 'on',
           'ssl_proxy_check_peer_name'   => 'on',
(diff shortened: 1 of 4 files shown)


bmjen:
@occelebi can you please make the updates as suggested by @igalic in the applied labels? Feel free to ask if any assistance or explanation is needed.

occelebi:
@bmjen I've added validation test and docs.

spec/defines/vhost_spec.rb
```diff
@@ -154,6 +154,7 @@
           'ssl_verify_depth'            => '3',
           'ssl_options'                 => '+ExportCertData',
           'ssl_openssl_conf_cmd'        => 'DHParameters "foo.pem"',
+          'ssl_proxy_verify'            => 'require',
```
bmjen: Since the template checks for existence of `@ssl_proxy_verify` before adding the line to the config, should this default be to `undef`? By setting a default, this means that the `SSLProxyVerify` line will always be present in the default config. Is that what we want?

cc/ @igalic

occelebi: @bmjen What would you suggest ? I might remove it. It doesn't bother me at all.

bmjen: Oops sorry @occelebi . I misread the diff and didn't see this was in the test and not manifest. This is fine the way it is! Thanks!

PR #1318, opened 2016-01-11
Added support for more LDAP parameters

netman2k:
There's many options these are all related LDAP cache setting stuff.
There are LDAPSharedCacheSize,  LDAPCacheEntries, LDAPCacheTTL, LDAPOpCacheEntries and LDAPOpCacheTTL.

it would be appreciated, if it applied to the codes.


igalic:
hi @netman2k! thank you very much for this patch.
in order to merge it, we'll need documentation, and tests!

tphoney:
Merged #1352 
Thanks for the work you put in.

PR #1320, opened 2016-01-11
Update docs for predefined LogFormats

mpolenchuk:
Add missed predefined "forwarded" log format into docs.


igalic:
nice! thank you @mpolenchuk!

PR #1321, opened 2016-01-11
Fixing error in Amazon $operatingsystem comparison

ryno75:


manifests/params.pp
```diff
@@ -52,7 +52,7 @@
   } else {
     $verify_command = '/usr/sbin/apachectl -t'
   }
-  if $::osfamily == 'RedHat' or $::operatingsystem == 'amazon' {
+  if $::osfamily == 'RedHat' or $::operatingsystem =~ /^[Aa]mazon$/ {
```
igalic: this is upsetting and saddening.

igalic:
how did we ever send someone on the moon?

i mean, thank you for catching and fixing this @ryno75!

PR #1322, opened 2016-01-12
MODULES-2958 : correct CustomLog syslog entry

BigAl:


igalic:
@BigAl would you mind squashing those two commits?

BigAl:
Squashing done

tphoney:
thanks for the work !

PR #1326, opened 2016-01-13
proxy: remove workaround for old broken clients

igalic:
this has been fixed for a long time
(https://bz.apache.org/bugzilla/show_bug.cgi?id=38864) and the
workaround documented by @niq http://www.apachetutor.org/admin/reverseproxies
hasn't been necessary for a long time.

### DIFF
--- a/templates/vhost/_proxy.erb
+++ b/templates/vhost/_proxy.erb
@@ -19,27 +19,26 @@
   <%- end -%>
   <%- if proxy['keywords'] %> <%= proxy['keywords'].join(' ') -%>
   <%- end %>
-  <Location <%= proxy['path']%>>
   <%- if not proxy['reverse_cookies'].nil? -%>
     <%- Array(proxy['reverse_cookies']).each do |reverse_cookies| -%>
       <%- if reverse_cookies['path'] -%>
-        ProxyPassReverseCookiePath <%= reverse_cookies['path'] %> <%= reverse_cookies['url'] %>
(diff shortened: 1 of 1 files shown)


igalic:
i'm impressed how this actually passes tests.

since i haven't touched spec/ at all.

oh well, i guess that's probably because there's no spec tests exercising this code-path at all…

@DavidS you got any opinions to spare?

DavidS:
https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/acceptance/vhost_spec.rb#L95 has some tests for this functionality, but the tests don't seem to be very inquisitive.

igalic:
works for me™
do you want me to add anything?

DavidS:
If you have time, it'd be great to have a test that actually talks through the proxy, otherwise just merge this.

igalic:
time is an extremely valuable resource.
i have had those tests in spec/acceptance.
it was axed.

tphoney:
Turns out the tests dont pass :(

igalic:
which ones?

tphoney:
@igalic i have a fix that i am testing, i will post it when ready. 
/spec/acceptance/vhost_spec.rb is failing on multiple OSes  in templates/vhost/_proxy.erb
we need a path associated with ProxyPassReverse 

 ProxyPassMatch /foo http://backend-foo/
  ProxyPassReverse http://backend-foo/
should become
 ProxyPassMatch /foo http://backend-foo/
  ProxyPassReverse /foo http://backend-foo/

tphoney:
@igalic @DavidS my proposed fix #1327

PR #1327, opened 2016-01-14
a path is needed for ProxyPassReverse

tphoney:


spec/acceptance/vhost_spec.rb
```diff
@@ -1013,7 +1013,7 @@ class { 'apache': service_ensure => stopped, }
 
     describe file("#{$vhost_dir}/25-test.server.conf") do
       it { is_expected.to be_file }
-      it { is_expected.to contain 'ProxyPass          / test2/' }
+      it { is_expected.to contain 'ProxyPass        / test2/' }
```
igalic: `%r{ProxyPass\s+ / test/}` ?

tphoney: i totally agree !
if we were to do this, it should be done for the whole file. so it is consistent

igalic: +1

PR #1328, opened 2016-01-14
Expose the directory that is assumed/supposed to be used for virtual hosts

mpdude:
Use case: I use to install virtual hosts in directories below `/var/www`. Some tools and cron jobs that need to traverse all the hosts could previously use `${::apache::docroot}` to find this directory.

However, fcc4d43f46ee2a0d835b7bbf1ba90387eabf3719 changed that setting (due to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=730372) and left no variable, setting or pointer to the `/var/www` value.

It may be debatable whether a read-only value like this new `$sites_root` makes sense. But both the above Debian bug report as well as the README.md examples describe virtual hosts at `/var/www/some.host.name`, so I think a generic reference to it comes in handy.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -136,14 +136,14 @@ class apache::params inherits ::apache::version {
     $mime_support_package = 'mailcap'
     $mime_types_config    = '/etc/mime.types'
     $sites_root           = '/var/www'
-    $docroot              = "$sites_root/html"
+    $docroot              = "${sites_root}/html"
     $alias_icons_path     = $::apache::version::distrelease ? {
       '7'     => '/usr/share/httpd/icons',
-      default => "$sites_root/icons",
+      default => "${sites_root}/icons",
     }
     $error_documents_path = $::apache::version::distrelease ? {
       '7'     => '/usr/share/httpd/error',
-      default => "$sites_root/error"
+      default => "${sites_root}/error"
     }
     if $::osfamily == 'RedHat' {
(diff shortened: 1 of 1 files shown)


jonnytpuppet:
Hi There

This is a pretty big change. It requires passing tests, readme updates, and additional tests before we can merge this.

Thanks

mpdude:
Sure, will work on that if you basically support the idea behind it?

jonnytpuppet:
@mpdude I was thinking about this more. For your use case, what stops you from overriding the docsroot property back to /var/www in your environment? Would that not solve your issue?

Thanks

mpdude:
The Debian bug (and potential security issue) made me aware that we need to make a distinction between the document root of the default (or any other virtual) host and the directory that is supposed/recommended/typically used to hold all the virtual hosts. I am not even sure if the latter is of any meaning for Apache itself.

igalic:
@mpdude what i generally do in my setups is:
```apache
<Directory />
  Set Permissions to lol nope
</Directory>

<Directory /srv/web>
  Set slightly less restrictive permissions
</Directory>

<Vhost x>
  Servername x
  <Directory /srv/web/x/htdocs>
   Set very specific permissions, if necessary
  </Directory>
</Vhost>
```

tphoney:
The workaround given by @igalic looks good and there has been no further activity. I am closing the PR. Thanks for the work and effort you have put into theis @mpdude . Cheers !

PR #1331, opened 2016-01-15
Fix fcgid.conf on Debian

sathieu:
Fixes #1178 regression on Debian: the conf symlink was missing.


jonnytpuppet:
@DavidS is this related to any apache problem you were seeing on debian?

DavidS:
@jonnytpuppet the debian problems I've been working on are related to service management and https://tickets.puppetlabs.com/browse/PUP-5016 . This looks more intricate.

spec/classes/mod/fcgid_spec.rb
```diff
@@ -102,8 +99,8 @@
     let :facts do
       {
         :osfamily                  => 'FreeBSD',
-        :operatingsystemrelease    => '10',
-        :operatingsystemmajrelease => '10',
+        :operatingsystemrelease    => '9',
+        :operatingsystemmajrelease => '9',
```
sethlyons: how come you changed this from 10 to 9?

manifests/mod/fcgid.pp
```diff
@@ -1,17 +1,24 @@
 class apache::mod::fcgid(
   $options = {},
 ) {
+  if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '7' {
```
sethlyons: this is needed on freebsd as well.

spec/classes/mod/fcgid_spec.rb
```diff
@@ -38,15 +36,13 @@
         :id                        => 'root',
         :kernel                    => 'Linux',
         :path                      => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
-        :is_pe                     => false,
+        :is_pe                  => false,
       }
```
igalic: please undo this change

manifests/mod/fcgid.pp
sathieu: Was this your original goal? Because #1178 doesn't talk about FreeBSD.

If you goal is just FreeBSD, add it to the condition instead of breaking every non-Redhat distro.

spec/classes/mod/fcgid_spec.rb
sathieu: I try to fix just one think in this PR. Mixing things together leads to regressions.

But I will add a commit for this indentation fix.

spec/classes/mod/fcgid_spec.rb
sathieu: From the revert of your commit that didn't explain the reason of this change.

I can apply it again, once I know what you canted to fix.

sethlyons:
once the revert is successful, i'm happy to change this to add freebsd to the redhat condition.  that being said, it was completely reasonable to implement this as a default instead of just for RH and freebsd, since [it is required to load mod_unixd before mod_fcgid](http://httpd.apache.org/mod_fcgid/mod/mod_fcgid.html).

sathieu:
I think this new OK fixes all the remarks:
- It now works again on Debian
- @sethlyons I uses unixd_fcgid on FreeBSD
- @bmjen it's rebased and squashed

bmjen:
Thanks @sathieu !

PR #1333, opened 2016-01-15
add paramter to set config file permissions

timogoebel:
This PR adds an parameter `file_mode` to specify the desired permissions mode the config files are created with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1160,6 +1160,10 @@ Controls how Apache handles `TRACE` requests (per [RFC 2616][]) via the [`TraceE
 
 Controls whether the systemd module should be installed on Centos 7 servers, this is especially useful if using custom built rpms. This can either be 'true' or 'false, defaults to 'true'.
 
+##### `file_mode`
+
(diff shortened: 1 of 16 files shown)


jonnytpuppet:
Hi There

The tests will need to be fixed before we can merge this.

Also, can you please state a business case where the default permissions are not acceptable?

Thanks

timogoebel:
I think that the test failures are not related and already broken in master.

A business case for the parameter is quite common: one may need to pass hardening requirements from pci-dss, cis, ...
I can add this to  the docs if you want...

timogoebel:
@jonnytpuppet: I just had another look at the acceptance test failures. Apache can not start because of
```
Invalid command 'FastCgiExternalServer', perhaps misspelled or defined by a module not included in the server configuration
```
This is definitely not related to the changes in this PR.

jonnytpuppet:
@timogoebel You are indeed correct, these failures are unrelated to your changes. @tphoney is working on apache internally to see if we can fix the currently build, then we can look at merging this.

Thanks for your patience

PR #1335, opened 2016-01-18
fix vhosts listen to wildcard ip

timogoebel:
```
apache::vhost { 'test':
  ip => '*',
}
```

currently fails. This PR fixes the behavior.

### DIFF
--- a/lib/puppet/parser/functions/enclose_ipv6.rb
+++ b/lib/puppet/parser/functions/enclose_ipv6.rb
@@ -28,16 +28,18 @@ Takes an array of ip addresses and encloses the ipv6 addresses with square brack
     result = []
 
     input.each do |val|
-      begin
-        ip = IPAddr.new(val)
-      rescue *rescuable_exceptions
-        raise(Puppet::ParseError, "enclose_ipv6(): Wrong argument "+
-          "given #{val} is not an ip address.")
+      unless val == '*'
+        begin
+          ip = IPAddr.new(val)
+        rescue *rescuable_exceptions
+          raise(Puppet::ParseError, "enclose_ipv6(): Wrong argument "+
+                "given #{val} is not an ip address.")
+        end
+        val = "[#{ip.to_s}]" if ip.ipv6?
       end
-      val = "[#{ip.to_s}]" if ip.ipv6?
-      result = [result,val]
+      result << val
     end
 
-    return result.flatten.compact
+    return result.uniq
   end
 end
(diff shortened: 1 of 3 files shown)


lib/puppet/parser/functions/enclose_ipv6.rb
```diff
@@ -28,16 +28,18 @@ module Puppet::Parser::Functions
...
+                "given #{val} is not an ip address.")
+        end
+        val = "[#{ip.to_s}]" if ip.ipv6?
       end
-      val = "[#{ip.to_s}]" if ip.ipv6?
-      result = [result,val]
+      result << val
     end
 
-    return result.flatten.compact
```
jonnytpuppet: It would be appreciated if you could explain why it is ok that this can be removed, and how it related to wildcard vhosts, thanks!

spec/unit/puppet/parser/functions/enclose_ipv6_spec.rb
```diff
@@ -49,8 +53,8 @@
   end
 
   it "should embrace ipv6 adresses within an array of ip addresses" do
-    result = scope.function_enclose_ipv6([['127.0.0.1','fe80::1','[fe80::1]']])
-    expect(result).to(eq(['127.0.0.1','[fe80::1]','[fe80::1]']))
+    result = scope.function_enclose_ipv6([['127.0.0.1','fe80::1','[fe80::2]']])
+    expect(result).to(eq(['127.0.0.1','[fe80::1]','[fe80::2]']))
   end
```
jonnytpuppet: It would be appreciated if these changes could be explained in the context of wildcard vhosts. Thanks!

lib/puppet/parser/functions/enclose_ipv6.rb
timogoebel: It's because I changed ```result = [result,val]``` to ```result << val``` which makes more sense. ```input``` is already flattened in line 27, so we do not need to flatten it once again.

spec/unit/puppet/parser/functions/enclose_ipv6_spec.rb
timogoebel: The tests broke, because I added ```return result.uniq```. An apache vHost can never listen on a single ip twice, can it?

igalic: pretty sure it'll actually cause an error

PR #1337, opened 2016-01-19
MODULES-2990: Gentoo - fix module includes in portage::makeconf

derdanne:
https://tickets.puppetlabs.com/browse/MODULES-2990

update module handling for gentoo when used apache 2.4
add module authnz_external as external package in gentoo

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -273,9 +273,24 @@ class apache (
         $scriptalias          = '/var/www/localhost/cgi-bin'
         $access_log_file      = 'access.log'
 
-        ::portage::makeconf { 'apache2_modules':
-          content => $default_mods,
+        if is_array($default_mods) {
+          if versioncmp($apache_version, '2.4') >= 0 {
+            if defined('apache::mod::ssl') {
+              ::portage::makeconf { 'apache2_modules':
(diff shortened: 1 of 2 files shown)


jonnytpuppet:
Hi There

Thanks for this PR.

Unfortunately, Gentoo is not a supported or compatible operating system for this module. To bring a module up to a "compatible" level for a particular OS, it would need a reference in the Readme, have some level of testing, as well as be included in the metadata. Also, all other features would need to work with the OS, or at least the "main" features.

All of these things would need to be done before we can merge this PR.

Thanks

derdanne:
i added the support last year, and it is in the README ;)

jonnytpuppet:
Hi There

Happy to merge this given you are maintaining the unofficial gentoo support. Apologies that I didn't notice this. Perhaps an update to the Readme to state the unofficial support for gentoo would be cool :)

Thanks

JT

derdanne:
Hi,

In the README is written: 

This module also provides functions for other distributions and operating systems, such as FreeBSD, Gentoo, and Amazon Linux, but is not formally tested on them and are subject to regressions.

I think this should be sufficient, what do you think?

jonnytpuppet:
@derdanne I missed that too! I was looking in the limitations section which I don't think makes a reference to Gentoo?

Thanks

PR #1339, opened 2016-01-20
add parameter root_directory_options

timogoebel:
This PR adds the parameter root_directory_options to the apache class in order to change the options in httpd.conf:
```
<Directory />
  Options FollowSymLinks
   AllowOverride None
 </Directory>
```

This is necessary for hardening of apache, e.g. it is done in the hardening.io project.


bmjen:
Thanks @timogoebel. This new PR looks good! Can you please add some documentation to the README for the new parameter?

timogoebel:
@bmjen : Thanks for reviewing. I added  the missing docs (which I just forgot to git add :-( ).

PR #1341, opened 2016-01-20
Allow 'auth_require' to take an array.

gerases:
This will allow stanzas like this:

  <LocationMatch "CONDITION">
    Require user user1
    Require group group1
  </LocationMatch>

In addition, if auth_require is set to an empty string, it should not be inserted. This is a valid case because the require directives may be already defined in a parent location.

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -62,15 +62,19 @@
     Require <%= req %>
         <%- end -%>
       <%- end -%>
-      <%- if directory['auth_require'] -%>
+      <%- if directory['auth_require'] && directory['auth_require'] != '' -%>
+        <%- Array(directory['auth_require']).each do |auth_req| -%>
     Require <%= directory['auth_require'] %>
+        <%- end -%>
       <%- end -%>
       <%- if !(directory['require'] && directory['require'] != '') && directory['require'] !~ /unmanaged/i && !(directory['auth_require']) -%>
     Require all granted
       <%- end -%>
     <%- else -%>
-      <%- if directory['auth_require'] -%>
+      <%- if directory['auth_require'] && directory['auth_require'] != '' -%>
+        <%- Array(directory['auth_require']).each do |auth_req| -%>
     Require <%= directory['auth_require'] %>
+        <%- end -%>
       <%- end -%>
       <%- if directory['order'] and directory['order'] != '' -%>
     Order <%= Array(directory['order']).join(',') %>


templates/vhost/_directories.erb
```diff
@@ -62,15 +62,19 @@
     Require <%= req %>
         <%- end -%>
       <%- end -%>
-      <%- if directory['auth_require'] -%>
+      <%- if directory['auth_require'] && directory['auth_require'] != '' -%>
+        <%- Array(directory['auth_require']).each do |auth_req| -%>
     Require <%= directory['auth_require'] %>
```
hunner: This should be `Require <%= auth_req %>`

templates/vhost/_directories.erb
```diff
@@ -62,15 +62,19 @@
...
+        <%- end -%>
       <%- end -%>
       <%- if !(directory['require'] && directory['require'] != '') && directory['require'] !~ /unmanaged/i && !(directory['auth_require']) -%>
     Require all granted
       <%- end -%>
     <%- else -%>
-      <%- if directory['auth_require'] -%>
+      <%- if directory['auth_require'] && directory['auth_require'] != '' -%>
+        <%- Array(directory['auth_require']).each do |auth_req| -%>
     Require <%= directory['auth_require'] %>
```
hunner: Ditto

gerases:
I'm sorry, I'm not quite following

hunner:
Your erb code boils down to this:
```erb
<% Array(directory['auth_require']).each do |auth_req| -%>
     Require <%= directory['auth_require'] %>
```
But this means that the string passed to the `puts` is `directory['auth_require']` and not `auth_req` as you appear to intend, so if you pass an array of values to auth_require then you'll get all of them smushed together and printed multiple times.

Also, you appear to be missing the `end` for the `.each`

What you really want is something like this:
```erb
<% Array(directory['auth_require']).each do |auth_req| -%>
     Require <%= auth_req %>
<% end -%>
```

hunner:
Also if you could update the readme and add a unit test and perhaps an acceptance test to verify this behavior, that would be great. Thanks!

gerases:
I see my confusion now. If you look at the diff, you will see that the line

`Require <%= directory['auth_require'] %>`

Is not part of my change. It's been some time now and I don't remember how I submitted my pull request. It was one of my first ones too :)

Anyway, the complete code is this:

```
      <%- if directory['auth_require'] && directory['auth_require'] != '' -%>
        <%- Array(directory['auth_require']).each do |auth_req| -%>
    Require <%= auth_req %>
        <%- end -%>
      <%- end -%>
```

And I can verify that it does work.

I've never written an acceptance test but I'll give it a shot.

hunner:
@gerases Ah, I see that you did indeed have an `<%- end -%>` already.

You submitted the PR using the web interface on github which created a branch called `patch-1` on your fork of the module. If you clone your fork, you should be able to check out that branch and push any changes to update this pull request.

You are correct in saying that the `Require <%= directory['auth_require'] %>` is not changed by you, however it is on the inner part of the loop and so is affected by your change, so they do need to be changed.

For acceptance tests, you should be able to just update an existing test like turning https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/acceptance/vhost_spec.rb#L464 into an array of auth_require values and verifying that it works by copy/pasting the curl line https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/acceptance/vhost_spec.rb#L525 and updating for each user in the array.

jonnytpuppet:
@gerases Hi there, have you had a chance to look at this yet? Thanks!

gerases:
No :(, I haven't had a chance. But it's on my list of things to do. Unless
somebody wants to do it before me?

igalic:
we need a new label "help wanted"

gerases:
I'll do it within 3 weeks. The process of testing under puppet is rather convoluted. Specifically setting up the test environment itself. I need to find a chunk of time to muddle through the ruby rspec obscurity.

gerases:
Also, I found another issue with the module and its readiness for apache 2.4. I don't know what the fix should be but...

The additional_includes directive adds the relative content at the BOTTOM of the `vhost` definition due to the priority number in the `concat` configuration in `vhost.pp`. That's a problem for certain verified cases involving authentication in 2.4 where it seems important to place an include BEFORE a container (Directory, Location, LocationMatch, etc). As a quick fix, I've swapped the priorities in my local vhost.pp like so:

```  if $_directories and ! empty($_directories) {
    concat::fragment { "${name}-directories":
      order   => 70,
      content => template('apache/vhost/_directories.erb'),
    }
  }

  # Template uses:
  # - $additional_includes
  if $additional_includes and ! empty($additional_includes) {
    concat::fragment { "${name}-additional_includes":
      order   => 60,
      content => template('apache/vhost/_additional_includes.erb'),
    }
  }
```

I'm not sure what would be a better solution. Any suggestions?

Thanks

gerases:
Hi,

I've finally got around to writing the tests for my change and I do have a question. It seems like the `require` option as well as `auth_require` insert the `Require` directive. The former takes and array while the latter expects a string. It's not clear from the docs what the difference between them is.

I can of course just use the `require` variant and then this change request will go away completely.

Suggestions?

Thanks,
 Sergei

hunner:
Even if you push the fixes for the code without having fixes to verify the changes, that would be helpful too.

gerases:
I'm sorry, I'm confused by that statement in light of my last message :)

I was saying that I wasn't sure if there were any problems left.

tphoney:
Hi @gerases , thanks for the work you have put into this pull request. However due to inactivity and  needing rebased this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request. Feel free to reach out to @hunner or me if you need some further help.

PR #1345, opened 2016-01-28
Preserve /etc/httpd/conf.d/passenger.conf on EL7

domcleal:
mod_passenger provides a passenger.conf containing PassengerRoot to
ensure the service starts up correctly.  On EL7, this is installed at
/etc/httpd/conf.d/passenger.conf and has been protected from purging
since 005bf61.

The change in 2a608592 to move LoadModule config files to
/etc/httpd/conf.modules.d/ on EL7 also affected the path of the
protected passenger.conf.  This change reverts the path change to match
the packages.

In summary, prior to 2a608592 (release 1.7.0):

- /etc/httpd/conf.d/passenger.conf with PassengerRoot (packaged)
- /etc/httpd/conf.d/passenger_extra.conf with apache::mod::passenger settings
- /etc/httpd/conf.d/zpassenger.load with LoadModule

In 2a608592 (release 1.8.0):

- /etc/httpd/conf.modules.d/passenger_extra.conf with apache::mod::passenger settings
- /etc/httpd/conf.modules.d/zpassenger.load with LoadModule
- /etc/httpd/conf.d/passenger.conf is purged

With this commit:

- /etc/httpd/conf.modules.d/passenger_extra.conf with apache::mod::passenger settings
- /etc/httpd/conf.modules.d/zpassenger.load with LoadModule
- /etc/httpd/conf.d/passenger.conf with PassengerRoot (packaged)

---

I targeted this at the 1.8.x branch, as it's a regression in 1.8.0 - is that OK?  I assume it'll get merged up to master and possibly put into a patch release this way.

### DIFF
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -35,7 +35,7 @@ class apache::mod::passenger (
   # Managed by the package, but declare it to avoid purging
   if $passenger_conf_package_file {
     file { 'passenger_package.conf':
-      path => "${::apache::mod_dir}/${passenger_conf_package_file}",
+      path => "${::apache::confd_dir}/${passenger_conf_package_file}",
     }
   }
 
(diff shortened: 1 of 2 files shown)


timogoebel:
@bmjen : Any chance, that this commit gets into master as well?

PR #1346, opened 2016-01-29
Allow value of $ssl_mutex to be defined outside this module

lapthorn:
I'd like to override the value of ssl_mutex (https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/ssl.pp#L16-L41) and supply a value from an external source such as hiera.

e.g `apache::mod::ssl::mutex: 'posixsem'`

Why? http://blather.michaelwlucas.com/archives/678

Setting `$ssl_mutex = 'default'` is breaking our Apache install, but I can't re-assign the variable.

I can do a hiera lookup using an intermediate variable like:
``` puppet
  if $mutex != undef {
        $ssl_mutex = $mutex
  }
  else {
    ... current module behaviour....
  }
```

Is there a neater way to do this?  I'm more than happy to work on this module, as that's what I'm doing already to get it working in our installation.

### DIFF
--- a/manifests/mod/ssl.pp
+++ b/manifests/mod/ssl.pp
@@ -9,34 +9,39 @@ class apache::mod::ssl (
   $ssl_pass_phrase_dialog  = 'builtin',
   $ssl_random_seed_bytes   = '512',
   $ssl_sessioncachetimeout = '300',
+  $mutex                   = undef,
   $apache_version          = $::apache::apache_version,
   $package_name            = undef,
 ) {
 
-  case $::osfamily {
-    'debian': {
-      if versioncmp($apache_version, '2.4') >= 0 {
-        $ssl_mutex = 'default'
-      } elsif $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease == '10.04' {
-        $ssl_mutex = 'file:/var/run/apache2/ssl_mutex'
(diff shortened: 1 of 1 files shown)


bmjen:
Thanks for the contribution @lapthorn . 

Some feedback. I definitely see the need for this option being configurable. However, typically our convention for doing this would include exposing the parameter named as its option name `ssl_mutex`. I'd recommend exposing the parameter as described, and then setting the OS-specific defaults in `manifests/params.pp`. That way in the `apache:mod:ssl` class you can expose the `ssl_mutex` parameter with a default of `$::apache::params::ssl_mutex`. Similar to how it is done with some of the passenger mod parameters here: https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/passenger.pp#L12-L14

In addition to that, you'll also need to update the README documentation to include this new parameter as well as spec tests.

lapthorn:
@bmjen Thanks for the comment, that's what I was hoping you'd say!  It follows the fairly common and well defined params.pp pattern.  I'll get on with it - but not sure how long I'll be ${DAYJOB} and all that....

bmjen:
Thanks @lapthorn! Please feel free to ask questions. We certainly appreciate your contribution!

PR #1347, opened 2016-01-29
Add apache_version fact

jyaworski:


bmjen:
Thanks @jyaworski !

lib/facter/apache_version.rb
```diff
@@ -0,0 +1,8 @@
+Facter.add(:apache_version) do
+  setcode do
+    if Facter::Util::Resolution.which('apachectl')
+      apache_version = Facter::Util::Resolution.exec('apachectl -v 2>&1')
```
igalic: on older versions of Debuntu, this will not work, because they renamed their apachectl to apache2ctl

jyaworski: How old is 'older'? 7?

hunner: It may not work, but then it will just be undef so same ol' same ol'.

igalic: 2.2

PR #1349, opened 2016-01-29
MODULES-3020 - quote php_values when the value is a string

jlambert121:
Ensure php_values with a string type as a value get quoted.

### DIFF
--- a/spec/acceptance/mod_php_spec.rb
+++ b/spec/acceptance/mod_php_spec.rb
@@ -80,7 +80,7 @@ describe 'apache::mod::php class' do
 
     describe file("#{$vhost_dir}/25-php.example.com.conf") do
       it { is_expected.to contain "  php_flag display_errors on" }
-      it { is_expected.to contain "  php_value include_path .:/usr/share/pear:/usr/bin/php" }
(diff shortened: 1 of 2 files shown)


tphoney:
Thanks for the code submission.
It looks like the new changes break the acceptance tests, can you please have a look. 
The linked modules ticket is https://tickets.puppetlabs.com/browse/MODULES-3020

bmjen:
The test failure:

```
1) apache::mod::php class custom extensions, php_flag, php_value, php_admin_flag, and php_admin_value File "/etc/apache2/sites-enabled/25-php.example.com.conf" should contain "  php_value include_path .:/usr/share/pear:/usr/bin/php"
     Failure/Error: it { is_expected.to contain "  php_value include_path .:/usr/share/pear:/usr/bin/php" }
       expected File "/etc/apache2/sites-enabled/25-php.example.com.conf" to contain "  php_value include_path .:/usr/share/pear:/usr/bin/php"
     # ./spec/acceptance/mod_php_spec.rb:83:in `block (4 levels) in <top (required)>'
```

@jlambert121 it appears that the tests need to be updated to expect the quotes.

PR #1227, opened 2015-10-15
ThreadLimit needs to be above MaxClients or it is ignored. https://bz…

mattkenn4545:
….apache.org/bugzilla/show_bug.cgi?id=46113

This should also be forwardported but we are currently stuck at 1.2

PR #1351, opened 2016-02-01
ThreadLimit needs to be above MaxClients or it is ignored. https://bz…

tphoney:
….apache.org/bugzilla/show_bug.cgi?id=46113


tphoney:
This is a rebase of #1227

PR #1352, opened 2016-02-01
Ldap parameters

tphoney:


tphoney:
This is based on the work in #1318

PR #1355, opened 2016-02-03
allow status code on redirect match to be optional and not a requirement

BigAl:
as per apache documentation
  http://httpd.apache.org/docs/2.2/mod/mod_alias.html#redirectmatch


bmjen:
Thanks for the contribution @BigAl !

PR #1357, opened 2016-02-04
Use valid parameter name

emahags:
According to this document, uppercase letters are not allowed:
https://docs.puppetlabs.com/puppet/latest/reference/lang_reserved.html#parameters

### DIFF
--- a/manifests/mod/authnz_ldap.pp
+++ b/manifests/mod/authnz_ldap.pp
@@ -1,14 +1,22 @@
 class apache::mod::authnz_ldap (
-  $verifyServerCert = true,
+  $verify_server_cert = true,
+  $verifyServerCert   = undef,
 ) {
   include ::apache
   include '::apache::mod::ldap'
   ::apache::mod { 'authnz_ldap': }
 
-  validate_bool($verifyServerCert)
+  if $verifyServerCert {
+    warning('Class[\'apache::mod::authnz_ldap\'] parameter verifyServerCert is deprecated in favor of verify_server_cert')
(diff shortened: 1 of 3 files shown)


bmjen:
Thanks for the contribution @emahags . As much as I want to merge this PR as it is, I can't because it would make the mod backwards incompatible for people that rely on setting the parameter. 

A possible workaround for this problem is to phase out the old parameter with something like this:

```
class apache::mod::authnz_ldap (
  $verify_server_cert = true, 
  $verifyServerCert   = undef,
   
  ) {
    include '::apache::mod::ldap'
    ::apache::mod { 'authnz_ldap': }
  
   if $verifyServerCert {
     warning('Class[\'apache::mod::authnz_ldap\'] parameter verifyServerCert is deprecated in favor of verify_server_cert')
     $_verify_server_cert = $verifyServerCert
   } else {
     $_verify_server_cert = $verify_server_cert
   }
...
```

In something like this, the template would also need updating to use `$_verify_server_cert`.

/cc @igalic for thoughts.

tphoney:
@emahags is there anything we can do to help you with this ?

emahags:
@tphoney Ah sorry, I was waiting for an answer from "/cc @igalic for thoughts.". If the comment from @bmjen is how you want things done I'll go ahead and modify my pull request.

igalic:
:+1: what @bmjen said

hunner:
Modifying this PR would be perfect :). Thanks!

emahags:
@bmjen : Fixed backwards compatability
@tphoney : Rebased

PR #1360, opened 2016-02-05
Fix in custom fact "apache_version" for RHEL.

rvincentatprokarmadotcom:
The custom fact defined by lib/facter/apache_version.rb runs
"apachectl -v" and applies the following regular expression:

```
^Server version: Apache\/([\w\.]+) \(([\w]+)\)
```

On RHEL 7.1, running apachectl -v produces the following output:

```
Server version: Apache/2.4.6 (Red Hat Enterprise Linux)
```

The regex fails to match the output because it does not allow
for whitespace inside the parentheses.  The following modified
regex matches properly:

```
^Server version: Apache\/([\w\.]+) \(([\w ]+)\)
```

### DIFF
--- a/lib/facter/apache_version.rb
+++ b/lib/facter/apache_version.rb
@@ -2,7 +2,7 @@ Facter.add(:apache_version) do
   setcode do
     if Facter::Util::Resolution.which('apachectl')
       apache_version = Facter::Util::Resolution.exec('apachectl -v 2>&1')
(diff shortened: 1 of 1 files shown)


rvincentatprokarmadotcom:
For easy googling, the error shown on puppet runs without this patch is as follows:
```
Error: Facter: error while resolving custom fact "apache_version": undefined method `[]' for nil:NilClass
```

PR #1361, opened 2016-02-05
changed rpaf Configuration Directives: RPAF -> RPAF_

gvdb1967:
mod_rpaf: I changed the names of the different configuration directives as described on https://github.com/gnif/mod_rpaf
modified files:
spec/classes/mod/rpaf_spec.rb
templates/mod/rpaf.conf.erb

os: rhel/centos 6
- httpd-2.2
- mod_rpaf

### DIFF
--- a/spec/classes/mod/rpaf_spec.rb
+++ b/spec/classes/mod/rpaf_spec.rb
@@ -26,25 +26,25 @@ describe 'apache::mod::rpaf', :type => :class do
     it { is_expected.to contain_file('rpaf.conf').with({
       'path' => '/etc/apache2/mods-available/rpaf.conf',
     }) }
-    it { is_expected.to contain_file('rpaf.conf').with_content(/^RPAFenable On$/) }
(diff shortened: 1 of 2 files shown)


gvdb1967:
bundle exec rake spec
Finished in 10 minutes 45 seconds (files took 2.01 seconds to load)
1263 examples, 0 failures

antaflos:
This was a very bad call and breaks Apache/mod_rpaf for those of us that use Apache 2.2 on Ubuntu 12.04, where the configuration directives are, as per the documentation of `libapache2-mod-rpaf`, like this (from /usr/share/doc/libapache2-mod-rpaf/README):

```
RPAFenable On
# Enable reverse proxy add forward
RPAFproxy_ips 127.0.0.1 10.0.0.1
# which ips are forwarding requests to us
RPAFsethostname On
# let rpaf update vhost settings 
# allows to have the same hostnames as in the "real"
# configuration for the forwarding Apache
RPAFheader X-Forwarded-For
# Allows you to change which header mod_rpaf looks
# for when trying to find the ip the that is forwarding
# our requests
```

PR #1364, opened 2016-02-07
support Ubuntu xenial (16.04)

mmoll:
This should also go into 1.8.x


tphoney:
The change is good, Thanks !

PR #1368, opened 2016-02-12
Support socket communication option in apache::fastcgi::server

janschumann:
For local fcgi servers it makes sense to skip the TCP stack and communicate via socket.

This is also sugested in #MODULES-1405.

Instead of introducing a new parameter socket, validate functions are used to decide if we have a host or a socket in parameter `host`. 

It might be confusing to pass a socket path to a parameter named `host`, but changing the host parameter name to something like `host_or_socket` would introduce backwards incompatibility.

### DIFF
--- a/manifests/fastcgi/server.pp
+++ b/manifests/fastcgi/server.pp
@@ -11,6 +11,10 @@ define apache::fastcgi::server (
 
   Apache::Mod['fastcgi'] -> Apache::Fastcgi::Server[$title]
 
+  if is_absolute_path($host) {
+    $socket = $host
+  }
+
   file { "fastcgi-pool-${name}.conf":
     ensure  => present,
     path    => "${::apache::confd_dir}/fastcgi-pool-${name}.conf",
--- a/README.md
+++ b/README.md
@@ -3355,6 +3355,14 @@ apache::vhost { 'www':
 
 The hostname or IP address and TCP port number (1-65535) of the FastCGI server.
 
+It is also possible to pass a unix socket:
+
+``` puppet
+apache::fastcgi::server { 'php':
+  host        => '/var/run/fcgi.sock',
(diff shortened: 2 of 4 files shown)


manifests/fastcgi/server.pp
```diff
@@ -10,6 +10,14 @@
 
   Apache::Mod['fastcgi'] -> Apache::Fastcgi::Server[$title]
 
+  if is_absolute_path($host) {
+    $is_host = false
```
hunner: Perhaps this would be more accurately `$is_socket`?

hunner: Or even ditch the whole $is_whatever and go for:
```puppet
if is_absolute_path($host) {
  $socket = $host
} else {
  $split = split($host, ':')
  validate_ip_address($split[0])
  $_host = $host
}
```
Then
```erb
<%= if @_host then " -host" end %><%= if @socket then "-socket" end %>
```
which leaves room for other types. Are there types other than host or socket?

tphoney:
@janschumann is there anything else we can help with ?

janschumann:
@tphoney No thanks, but sorry for the delay. I am waiting until #1370 has been merged.

janschumann:
@hunner: afaik there are only tcp and socket connection options available ...

janschumann:
@hunner rebased and squashed

hunner:
@janschumann Thanks. Could you also mention the new functionality in https://github.com/puppetlabs/puppetlabs-apache#host-1 ?

janschumann:
@hunner docs added. I also removed the ip validation of the host parameter, as it may also be a host name :-)

hunner:
Good point.

janschumann:
@hunner thanks for merging!

PR #1370, opened 2016-02-12
support pass-header option in apache::fastcgi::server

janschumann:


templates/fastcgi/server.erb
```diff
@@ -1,3 +1,4 @@
-FastCGIExternalServer <%= @faux_path %> -idle-timeout <%= @timeout %> <%= if @flush then '-flush' end %> -host <%= @host %>
+FastCGIExternalServer <%= @faux_path %> -idle-timeout <%= @timeout %> <%= if @flush then '-flush' end %> -host <%= @host -%>
+<%- if @pass_header -%> -pass-header <%= @pass_header %><% end %>
```
hunner: Could you actually change these two `-%>\n<%-` to just `%> <%` so they're on the same line? It seems a little confusing to have them be on two lines if you want the output to be a single line, and flush is already setting precedent for single-line code. Thanks!

janschumann:
Yes it is a bin confusing. But isn´t the readability of the template also a criteria?

Please see #1368 .. This line might get even longer.

Thoughts?

hunner:
Ah yeah, if there are lots of options to pass then the other option is to pre-set all the variables in a `<% %>` block before the line to the full string like
```erb
<%
  timeout = " -idle-timeout #{@timeout}"
  host = " -host #{@host}"
  if @pass_header and ! @pass_header.empty?
    pass_header = " -pass-header #{@pass_header}"
  else
    pass_header = nil
  end
  #...
  options =  timeout + flush + host + pass_header
%>
FastCGIExternalServer <%= @faux_path %><%= options %>`
```

janschumann:
agreed. i´ll do that.

hunner:
Sorry for the late response on this. Could you document your new parameter in the readme as well?

janschumann:
Sure. 

While looking into the documentation of [mod_fastcgi](http://www.fastcgi.com/mod_fastcgi/docs/mod_fastcgi.html), I realized that there are some more options available, such as `appConnTimeout` and `user` etc.. Do you want me to implement them as well? In another pull-request?

hunner:
It's okay as-is; I always say to write for future extensibility, but present simplicity :).

:+1:

PR #1371, opened 2016-02-16
mod/ssl: Add option to configure SSL mutex

daenney:
This allows the end user to explicitly define or override what the `SSLMutex` or `Mutex` configuration for Apache will be as the platform default might not always be desirable.

Supersedes #1346
Closes #1346

### DIFF
--- a/README.md
+++ b/README.md
@@ -1665,6 +1665,11 @@ Installs [Apache SSL features][`mod_ssl`] and uses the `ssl.conf.erb` template t
 - `ssl_protocol`: Default: [ 'all', '-SSLv2', '-SSLv3' ].
 - `ssl_random_seed_bytes`: Valid options: A string. Default: '512'.
 - `ssl_sessioncachetimeout`: Valid options: A string. Default: '300'.
+- `ssl_mutex`: Default: Determined based on the OS. Valid options: See [mod_ssl][mod_ssl] documentation.
+  - RedHat/FreeBSD/Suse/Gentoo: 'default'
+  - Debian/Ubuntu + Apache >= 2.4: 'default'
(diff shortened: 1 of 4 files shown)


daenney:
@bmjen @igalic This change introduces a huge problem. Because we need to do a `versioncmp` for Debian all tests now fail as `apache_version` isn't set by any of the tests. What to do?

manifests/params.pp
```diff
@@ -212,6 +213,13 @@
     $default_ssl_cert    = '/etc/ssl/certs/ssl-cert-snakeoil.pem'
     $default_ssl_key     = '/etc/ssl/private/ssl-cert-snakeoil.key'
     $ssl_certs_dir       = '/etc/ssl/certs'
+    if versioncmp($apache_version, '2.4') >= 0 {
```
bmjen: This line is causing the lint failure:

`metadata-json-lint metadata.json
manifests/params.pp - WARNING: top-scope variable being used without an explicit namespace on line 216`

bmjen:
Thanks @daenney ! This looks good. Just need to fix the linting error.

daenney:
@bmjen But that's a problem. Because `apache_version` comes from `version.pp` which gets loaded and exposed by `init.pp` as `apache_version = $::apache::version::default`. But since `params.pp` gets loaded as part of `init.pp`, loading it in `params.pp` from `version.pp` might cause us to use a different version for the comparison there than anything the user might explicitly provide.

bmjen:
Yep. I agree. I'm thinking you're right @daenney, in going back to the implementation of #1346... but with the correct parameter name, docs, and tests.

daenney:
@bmjen Alright, updated.

bmjen:
:+1:

daenney:
:beers: :dancer: :dancers: :tada:

PR #1372, opened 2016-02-16
Update in init.pp to harden Apache

lasse-kristensen:
Disable ServerSignatur and TraceEnable also set ServerTokens to "Prod"
This is best practices to harden Apache .


hunner:
As this is a backwards-incompatible change, we can't merge this yet though it is correct.

Perhaps the manifest could get some code comments to document the new values to be used next time apache gets a major release (not in the near future as far as I know) and we document in the readme that these three params should have the specified values to harden apache?

jonnytpuppet:
@hunner what do you propose we do with this PR? It's a sensible additional, save for the fact that it's a breaking change. Thanks!

tphoney:
@lasse-kristensen, thanks for the work you have put into this pull request. However due to inactivity and it being backwards incompatible this pull request is being closed. If you want to re-submit this PR were it does not break previous versions, we will gladly look at the new pull request.

PR #1383, opened 2016-03-02
Remove SSLv3

ghoneycutt:


hunner:
@igalic So, this sounds nice; do you think this would be "backwards incompatible change" or "security bugfix" or what?

roidelapluie:
Security Bugfix. POODLE.

roidelapluie:
However we could give people to ability to re-add the sslv3 in libnss with a parameter?

igalic:
pretty sure we've deprecated sslv3 a couple of years ago: https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslprotocol

ghoneycutt:
Thanks!

hunner:
[RFC-7525](https://tools.ietf.org/html/rfc7525) would not even allow SSLv3 as a fallback, so yeah not allowing this to be turned on is a good idea.

PR #1384, opened 2016-03-02
Add JkMount/JkUnmount directives to vhost

smoeding:
`mod_jk` is one of the possibilities to use Apache as a frontend to Tomcat. The module understands `JkMount` and `JkUnMount` directives to configure whether an URL should be handled by Apache or by Tomcat. This patch enhances `apache::vhost` to allow configuration of these two directives for a virtual host.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2149,6 +2149,23 @@ apache::vhost { 'sample.example.net':
 }
 ```
 
+##### `jk_mounts`
+
+Sets up a virtual host with 'JkMount' and 'JkUnMount' directives to handle the paths for URL mapping between Tomcat and Apache. Default: undef.
+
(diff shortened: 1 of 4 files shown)


igalic:
what's wrong with mod_proxy_ajp?

smoeding:
There is probably nothing wrong with it. The [Tomcat Wiki](http://wiki.apache.org/tomcat/FAQ/Connectors) has some opinions on the pros and cons of both modules.

In our case `mod_jk` is working pretty good and we have not seen any reason to replace it with something else.

igalic:
imo, if you're on 2.4, you should just use mod_proxy_ajp

i for one have been using mod_proxy_http since tomcat 6…

smoeding:
We will check if  mod_proxy_* works for us. Thanks!

PR #1386, opened 2016-03-03
apache::balancer: Add a target parameter to write to a custom path

roidelapluie:
Thit commits implements a target parameter to the apache::balancer
definition to specify a different configuration path if needed (e.g with
a different suffix or in a different location).

### DIFF
--- a/manifests/balancer.pp
+++ b/manifests/balancer.pp
@@ -23,6 +23,10 @@
 # Hash, default empty. If given, each key-value pair will be used as a ProxySet
 # line in the configuration.
 #
+# [*target*]
+# String, default undef. If given, path to the file the balancer definition will
+# be written.
+#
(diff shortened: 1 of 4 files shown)


manifests/balancer.pp
```diff
@@ -41,21 +45,27 @@
 define apache::balancer (
   $proxy_set = {},
   $collect_exported = true,
+  $target = undef,
 ) {
   include ::apache::mod::proxy_balancer
 
-  $target = "${::apache::params::confd_dir}/balancer_${name}.conf"
+  if $target {
+    $real_target = $target
```
hunner: Could you use `$_target = $target` instead of `$real_target = $target` ? Thanks!

roidelapluie: done :-)

PR #1389, opened 2016-03-05
Add mellon_sp_metadata_file parameter for directory entries

jokajak:
This allows specifying the mellon_sp_metadata_file parameter


jokajak:
I don't believe the checks failed due to changes in this pull request.

PR #1392, opened 2016-03-08
Allow configuring mod_security's SecAuditLogParts

stig:
The default configuration for this includes "I" which is not always
always suitable, e.g. if you cannot tolerate POST parameters appearing
in your modsec_audit.log

You may want to omit `I` if mod_security is protecting a hypothetical
web service that accepts credit card data in a POST request, for
example.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -47,6 +47,8 @@ class apache::params inherits ::apache::version {
 
   $vhost_include_pattern = '*'
 
+  $modsec_audit_log_parts = 'ABIJDEFHZ'
+
   if $::operatingsystem == 'Ubuntu' and $::lsbdistrelease == '10.04' {
     $verify_command = '/usr/sbin/apache2ctl -t'
   } else {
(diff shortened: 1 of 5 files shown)


stig:
So many apologies about the many commits here. This is my first encounter with rspec, and I was initially not able to run the tests locally, so was coding blind. I believe it should now pass. I'm happy to squash the commits if desired. (Assuming this change is is accepted at all :-) )

stig:
It appears that the failure is unrelated to my change:

```
Finished in 24 minutes 12 seconds (files took 1 minute 29.64 seconds to load)
472 examples, 1 failure, 1 pending

Failed examples:

rspec ./spec/acceptance/vhost_spec.rb:1397 # apache::vhost define fastcgi applies cleanly
```

igalic:
@stig as far as i know, @hunner is trying to figure this out, as we speak

stig:
Should I be adding an entry to the README as well, to document this new option?

igalic:
@stig yes please!

manifests/params.pp
```diff
@@ -47,6 +47,8 @@
 
   $vhost_include_pattern = '*'
 
+  $modsec_audit_log_parts = 'ABIJDEFHZ'
```
hunner: According to [the docs](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SecAuditLogParts) the default is `ABCFHZ`. Why did you choose this instead? It actually says that `D` isn't even implemented yet.

templates/mod/security.conf.erb
```diff
@@ -50,7 +50,7 @@
     SecDebugLogLevel 0
     SecAuditEngine RelevantOnly
     SecAuditLogRelevantStatus "^(?:5|4(?!04))"
-    SecAuditLogParts ABIJDEFHZ
```
hunner: Oh, because this was already the default...

PR #1394, opened 2016-03-16
mod_ssl requires mod_mime for AddType directives

sathieu:


sathieu:
See https://github.com/puppetlabs/puppetlabs-apache/blob/master/templates/mod/ssl.conf.erb#L7

PR #1396, opened 2016-03-17
Move all ensure parameters from concat::fragment to concat

domcleal:
Deprecated in 1.1.x and has no effect in 2.x.  From commit 1919eb3, but
was reverted when temporarily removing 2.x support in f54393e.


bmjen:
Thanks @domcleal ! Nice to get rid of those pesky warnings.

igalic:
:shipit:

PR #1397, opened 2016-03-17
Fix MODULES-3158 (any string interpreted as SSLCompression on)

tpdownes:
1. Fix MODULES-3158 (any string interpreted as SSLCompression on)
1. Convert ssl_honorcipherorder to boolean, backport strings 'on' or 'off'
1. Follow lower case on/off as in Apache docs

### DIFF
--- a/templates/mod/ssl.conf.erb
+++ b/templates/mod/ssl.conf.erb
@@ -19,7 +19,11 @@
   SSLMutex <%= @_ssl_mutex %>
   <%- end -%>
   SSLCryptoDevice <%= @ssl_cryptodevice %>
-  SSLHonorCipherOrder <%= @ssl_honorcipherorder %>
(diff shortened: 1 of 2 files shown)


tpdownes:
The build test fails because it doesn't accept lower-case "(on|off)" even though Apache itself does. Submitting #1398.

templates/mod/ssl.conf.erb
```diff
@@ -19,7 +19,11 @@
   SSLMutex <%= @_ssl_mutex %>
   <%- end -%>
   SSLCryptoDevice <%= @ssl_cryptodevice %>
-  SSLHonorCipherOrder <%= @ssl_honorcipherorder %>
```
igalic: i think we have a function for that.

PR #1398, opened 2016-03-17
Fix MODULES-3158 (any string interpreted as SSLCompression on)

tpdownes:
1. Fix MODULES-3158 (any string interpreted as SSLCompression on)
1. Convert ssl_honorcipherorder to boolean, backport accepting 'on' or 'off'

### DIFF
--- a/README.md
+++ b/README.md
@@ -1660,7 +1660,7 @@ Installs [Apache SSL features][`mod_ssl`] and uses the `ssl.conf.erb` template t
 - `ssl_cipher`: Default: 'HIGH:MEDIUM:!aNULL:!MD5:!RC4'.
 - `ssl_compression`: Default: false.
 - `ssl_cryptodevice`: Default: 'builtin'.
-- `ssl_honorcipherorder`: Default: 'On'.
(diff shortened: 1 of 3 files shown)


tpdownes:
It is worth mentioning that #1397 failed only because the build tests do not accept "SSLCompression on" (lower-case, following docs). The documentation uses lower case and Apache accepts both lower and upper-case "(On|Off)". I would advise that the module output more closely match the documentation and that the build test more closely mimics either the documentation or Apache behavior.

But, in the interest of a single commit that passes the build test as written, I have provided this commit.

bmjen:
This looks good. But can you update the README to indicate that the parameter now allows valid values of `On, Off, true, false`?

tpdownes:
Documentation is for other people! I've updated the PR with what I think are necessary and useful updates to the relevant sections.

bmjen:
Thanks @tpdownes !

igalic:
there's a function for this ;)
https://github.com/puppetlabs/puppetlabs-apache/blob/master/lib/puppet/parser/functions/bool2httpd.rb

PR #1399, opened 2016-03-23
Added vhost option fastcgi_idle_timeout

michakrause:


igalic:
@michakrause this will need docs, and, as always, we'd like spec tests.

michakrause:
:-/ none of the fastcgi vhost options is documented so far, and the contained spec tests are of the same quality as the other fastcgi vhost spec tests.

michakrause:
Ok, I added some documentation.

PR #1400, opened 2016-03-25
Take igalic's suggestion to use bool2httpd

tpdownes:
In #1398 @igalic suggests replacing 2 sections of the SSL configuration Ruby template with a function designed to safely convert Booleans to their Apache On/Off equivalent. This PR implements the suggestion.

The PR should be safe because #1398 modifies the Puppet code to ensure that each variable is already a boolean with the intended default value.


tpdownes:
@igalic I have taken your suggestion but, to my eyes, ,the build appears to be failing because params.pp does not define `pidfile` for the Gentoo osfamily.

igalic:
@tpdownes ay
the build failure is un-related and needs to be fixed separately.

igalic:
thank y ou very much for your patch @tpdownes!

PR #1403, opened 2016-03-26
Configurability of SecDefaultAction for OWASP Core Rule Set

FlatKey:


manifests/mod/security.pp
```diff
@@ -10,6 +10,7 @@
   $content_types         = 'application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf',
   $restricted_extensions = '.asa/ .asax/ .ascx/ .axd/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .resources/ .resx/ .sql/ .sys/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/',
   $restricted_headers    = '/Proxy-Connection/ /Lock-Token/ /Content-Range/ /Translate/ /via/ /if/',
+  $secdefaultaction           = 'deny',
```
igalic: why is this unaligned?
and why does it have `sec` prefix, unlike the others? (we're already in the mod_security module

FlatKey: It's aligned with the longest existing variable name in this block (secpcrematchlimitrecursion, please look at the complete block). I've aligned the other variables in this block with my PR #1404 to fit with the style guide.
`sec` is not a prefix, `SecDefaultAction` is the affected directive in *templates/mod/security_crs.conf.erb* so I used the same name for the variable.

jonnytpuppet:
@FlatKey it would be appreciated if you could rebase this against current master as changes have gone in that should fix the failing unit tests (assuming nothing in this PR breaks them).

Thanks

FlatKey:
@jonnytpuppet rebased.

PR #1409, opened 2016-03-28
MODULES-3211: fix broken strict_variable tests

jlambert121:


witjoh:
Hi, 

I applied your PR locally, and it seems to compile ok.  One thought:
I think tremoving the  the rspec file mod/dev_spec.rb is a bad idea. Always better to make it work.  Adding following line makes rspec working again:
--- dev_spec.rb_orig	2016-03-29 13:54:35.825263541 +0000
+++ dev_spec.rb	2016-03-29 13:51:18.271206754 +0000
@@ -2,7 +2,10 @@
 
 describe 'apache::mod::dev', :type => :class do
   it_behaves_like "a mod class, without including apache"
-
+  let(:pre_condition) {[
+    'include apache'
+  ]}
+  
   [
     ['RedHat',  '6', 'Santiago', 'Linux'],
     ['Debian',  '6', 'squeeze', 'Linux'],

Also in the param.pp,  if google did gave me the right result for the default pidfile ... , 
$pidfile          = '/var/run/apache2.pid'

(someone familiar with gentoo should confirm this.)

Grts and thx for this PR.,. Did solve my rspec problems
Johan

jlambert121:
I considered this solution, but I try to keep my unit tests testing the class itself, not the effects of a class called by a class (that secondary class should be tested by it's own unit tests).  Tests just for the sake of quantity aren't always as effective.  That's the reason I decided to remove it, but I can put it back if needed.

Thanks for the Gentoo PID file.  I've updated with that (but can change it if needed).

igalic:
i like @jlambert121 solution better than trying to ~fake~ it in the spec tests.

PR #1411, opened 2016-03-28
Configurability of Collaborative Detection Blocking for OWASP Core Rule Set

FlatKey:


jonnytpuppet:
@FlatKey it would be appreciated if you could rebase this against current master as changes have gone in that should fix the failing unit tests (assuming nothing in this PR breaks them).

Thanks

FlatKey:
@jonnytpuppet rebased.

tphoney:
This needs rebased, from looking at your code you are changing the default behaviour. setvar:tx.anomaly_score_blocking=on 
Now the default behaviour is off. This would be a potentially breaking change for people using this module. What do you think @FlatKey

FlatKey:
@tphoney rebased. No, that's not correct, please see at the comment above the rule code block: `# If you want to use anomaly scoring mode, then uncomment this line.` and this rule is per default commented: `#SecAction \`. I uncommented this rule and set the affected variable *off* to get the same default behaviour but make it configurable.

tphoney:
Sorry i misread. Thanks for the PR :+1:

PR #1413, opened 2016-03-30
reverted rspec which was removed with PR 1409

witjoh:
Just readded the rspec which was removed with the pull request solving the strict_variables issue.

Removing rspec to solve things is in my opinion not a good solution.


witjoh:
duplicate #1414 which is merged.   
Can be closed

tphoney:
closing as per @witjoh

PR #1416, opened 2016-03-31
Fix in custom fact "apache_version" for OracleLinux.

Reamer:
The custom fact defined by lib/facter/apache_version.rb runs
"apachectl -v" and applies the following regular expression:

    ^Server version: Apache\/([\w\.]+) \(([\w ]+)\)

On OracleLinux 7.2, running apachectl -v produces the following output:

    Server version: Apache/2.4.6 ()
    Server built:   Nov 21 2015 05:34:59

The regex fails to match the output because it does not allow
nothing inside the parentheses.  The following modified
regex matches properly:

    ^Server version: Apache\/([\w\.]+) \(([\w ]*)\)

### DIFF
--- a/lib/facter/apache_version.rb
+++ b/lib/facter/apache_version.rb
@@ -2,7 +2,7 @@ Facter.add(:apache_version) do
   setcode do
     if Facter::Util::Resolution.which('apachectl')
       apache_version = Facter::Util::Resolution.exec('apachectl -v 2>&1')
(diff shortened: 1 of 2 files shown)


DavidS:
Thanks for your submission! Please add your test data to https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/unit/apache_version_spec.rb to avoid regressions on this in the future.

Reamer:
@DavidS tests added

DavidS:
Great!

PR #1420, opened 2016-04-01
Add parameter passanger_log_level

samuelb:
I added a parameter passenger_log_level to be able to configure the PassengerLogLevel option in passenger.conf.


DavidS:
Excellent :+1:

PR #1421, opened 2016-04-04
Allow package names to be specified for mod_proxy, mod_ldap, and mod_authnz_ldap

MG2R:
Simple change to allow users to change the package name on the aforementioned mods. This is based on the code for mod::ssl, which already had this feature.

I'm not (very) experienced in the subject of spec tests and could use some help if those need to be improved as well. The code works as expected in my setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@
 [`apache::mod::alias`]: #class-apachemodalias
 [`apache::mod::auth_cas`]: #class-apachemodauth_cas
 [`apache::mod::auth_mellon`]: #class-apachemodauth_mellon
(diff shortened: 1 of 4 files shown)


DavidS:
This is a quite trivial change, so additional spec coverage is not really required. What you should do, though, is adding the params to the README, so people can find them, and do not need to guess.

MG2R:
Updated the README. Sorry for the delay.

tphoney:
With the recent merges, you will need to rebase this. Many apologies.

MG2R:
No problem. Learned something new :)

PR #1423, opened 2016-04-07
Wrap mod_security directives in an IfModule

kimor79:


hunner:
Instead of casing this out (and thus potentially making it do nothing) could you cause apache::vhost to include apache::mod::security if the parameters that need it are passed? https://github.com/kimor79/puppetlabs-apache/blob/wrap_vhost_mod_security/manifests/vhost.pp#L988

kimor79:
My use case is I have a vhost which is added "everywhere" but mod_security is only enabled for certain roles. When mod_security is enabled I don't want it enabled for that vhost but that vhost shouldn't need to pull in mod_security just to disable it.

tphoney:
@hunner does this address your concern ?

PR #1427, opened 2016-04-14
Add simple <Limit> support + ProxyAddHeaders

costela:
The main commit in this PR introduces minimal support for the \<Limit\> block inside \<Location\> blocks.
To do this, the template code handling "Require" directives has been refactored to a template of its own to avoid code duplication. This new template in turn then gets included in the main _directories.erb template, both directly inside the \<Location\> block and also inside each \<Limit\> block.

Additionally, this PR also includes the relatively straightforward support for the ProxyAddHeaders directive.


bmjen:
@costela this looks good! Due to other PRs merged before, this will need a rebase. Thanks!

costela:
@bmjen done.

PR #1325, opened 2016-01-13
mod_proxy_wstunnel support

niklas:
This module allows Apache to proxy Websocket connections. The handling of the connections is supposed to be done by rewrite rules.

PR #1290, opened 2015-12-03
Added `apache::mod::proxy_wstunnel` class

netors:

PR #1428, opened 2016-04-14
Support mod_proxy_wstunnel for Apache 2.4+

tpdownes:
Includes basic documentation. Derived from previous pull requests #1290 and #1325.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1305,6 +1305,7 @@ The following Apache modules have supported classes, many of which allow for par
 * `proxy_balancer`
 * `proxy_html` (see [`apache::mod::proxy_html`][])
 * `proxy_http`
+* `proxy_wstunnel`
 * `python`
 * `reqtimeout`
 * `remoteip`\*
@@ -1635,6 +1636,9 @@ Default values depend on your operating system.
 ##### Class: `apache::mod::proxy_html`
 
 **Note**: There is no official package available for mod\_proxy\_html and thus it must be made available by means outside of the control of the apache module.
+
+##### Class: `apache::mod::proxy_wstunnel`
+Includes the mod\_proxy_\wstunnel module on Apache versions >= 2.4. For versions below, no action is taken.
 
 ##### Class: `apache::mod::reqtimeout`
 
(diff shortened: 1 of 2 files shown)


tpdownes:
The first version of this pull request had a couple minor problems. Please make sure you pull from 5eec44e

README.md
```diff
@@ -1636,6 +1637,9 @@ Default values depend on your operating system.
 
 **Note**: There is no official package available for mod\_proxy\_html and thus it must be made available by means outside of the control of the apache module.
 
+##### Class: `apache::mod::proxy_wstunnel`
+Includes the mod\_proxy_\wstunnel module on Apache versions >= 2.4.5. For versions below, no action is taken.
```
hunner: Technically `>= 2.4` not 2.4.5

tphoney:
@tpdownes Thanks for the submission, it would be great if you added a simple test case.

README.md
tpdownes: The module doesn't exist below 2.4.5 (unless backported) but `::apache::version` doesn't offer finer resolution. I fixed the README to reflect the actual behavior. A relatively quick scan doesn't reveal any OS-packaged versions that are 2.4.4 or below.

tpdownes:
It would probably take me a day or 2 to figure out how to mimic something like `spec/classes/mod/proxy_html_spec.rb`. It's not that I don't support the idea of unit tests, I'm just not a ruby coder and don't have much experience with the CI system.

tphoney:
@tpdownes do you need any help with the test case ?

tpdownes:
There's no way I'll be getting to this in the next month or two - total reorganization of the data center going on. Baby on way, etc. So, I guess the answer is yes.

tphoney:
@tpdownes good luck on the baby, and the data center. i will get a look at this next week.

tphoney:
@jonnytpuppet this should just need a simple test case, copying what is done for other apache plugins. It  looks like a quick win.

DavidS:
An equivalent class with tests was merged as #1465

PR #1430, opened 2016-04-17
MODULES-2890: PHP 7

sathieu:


sathieu:
I've updated the patch series to make it less disruptive to existing installations.

Wouter0100:
Seems to work here, thanks! :+1: Ubuntu 16.04.

SirUrban:
Is there any special reason why this merge request has not been accepted yet? - It works fine for us on Ubuntu 14.04. More and more people are going to use PHP 7 now and it should be a good idea to have that implemented soon.

Wouter0100:
@SirUrban And PHP5 isn't even available anymore in Ubuntu 16.04's repos. Only PHP7.0.

hunner:
Thanks a lot!

SirUrban:
Thank YOU for your quick reaction! :)

victorsmirnov:
When the changes are planned for the release?
It looks like read me file does not have newly added parameter `php_version` explained. I guess this should work out of the box on Ubuntu 16.04 but may be this parameter is required and should be set to `7.0`.

Thank you for your help!

Wouter0100:
Hmm, as I see it's planned for the next release @victorsmirnov (because it's in master). I don't know when master is being released tho.

victorsmirnov:
I see the commit for 1.10.0 preparations - https://github.com/puppetlabs/puppetlabs-apache/commit/cd3ff1394bc59a09c914dc522a2dc71d3bd5e492 Looks like it should be released shortly.

templates/mod/php.conf.erb
```diff
@@ -2,20 +2,12 @@
...
 # easy for developers to write dynamically generated webpages.
 #
-#<IfModule prefork.c>
-#  LoadModule php5_module modules/libphp5.so
-#</IfModule>
-#<IfModule worker.c>
-#  # Use of the "ZTS" build with worker is experimental, and no shared
-#  # modules are supported.
-#  LoadModule php5_module modules/libphp5-zts.so
-#</IfModule>
```
igalic: 2016: PHP officially (or at least automatically;) declared as thread-safe. :D

sathieu: ??
I'm not sure to understand what you say. This block remove lines that were already commented out. Not changing anything.

In puppetlabs-apache, mod_php still couldn't be installed with mpm_worker, only mpm_prefork and mpm_itk are allowed. This is also the default in Debian packages.

But this doesn't mean that php is not thread-safe. PHP is, but extensions are probably not. See http://php.net/manual/en/faq.installation.php#faq.installation.apache2.

PR #1432, opened 2016-04-19
(MODULES-3140) explicitly rely on hasrestart if no restart command is…

DavidS:
… passed

Apache always had a restart option for the init script and in some cases
does require special handling of the daemon when restarting, so this
prefers the init scripts' restart action over just stopping/starting it.


cmurphy:
:+1: very neat!

bmjen:
@DavidS Nice!

PR #1433, opened 2016-04-19
Expose verify_config in apache::vhost::custom

cmurphy:
In many cases verify_config confuses more than it helps and users would
like to be able to turn it off. Expose the
$apache::custom_config::verify_config parameter in
apache::vhost::custom so custom vhost users can also turn this off.


DavidS:
:+1:

PR #1434, opened 2016-04-20
Set default keepalive to On

sathieu:
According to docs "In some cases this has been shown to result in an almost 50% speedup in latency times for HTML documents with many images."

Ref: http://httpd.apache.org/docs/2.4/mod/core.html#keepalive

### DIFF
--- a/README.md
+++ b/README.md
@@ -973,7 +973,7 @@ Sets the Apache server's base configuration directory. This is useful for specia
 
 ##### `keepalive`
 
-Determines whether to enable persistent HTTP connections with the [`KeepAlive`][] directive. Valid options: 'Off', 'On'. Default: 'Off'.
+Determines whether to enable persistent HTTP connections with the [`KeepAlive`][] directive. Valid options: 'Off', 'On'. Default: 'On'.
 
 If 'On', use the [`keepalive_timeout`][] and [`max_keepalive_requests`][] parameters to set relevant options.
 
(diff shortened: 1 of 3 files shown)


sathieu:
This is a behavior change and should be marked as such in the release notes.

igalic:
O_o why was keepalive off to begin with?

antaflos:
> O_o why was keepalive off to begin with?

I was asking myself the same question a few weeks ago. Especially when using Apache as an SSL-terminating reverse proxy the `KeepAlive` option makes a very noticeable difference in latency. IMHO this should always have been `On` by default, with the option to disable it on a vhost basis. I'll submit a PR for the latter.

antaflos:
I have submitted PR #1447.

igalic:
I reckon the reason is that when this module started out 2.2 was still the default on most platforms.
we had ***very bad*** defaults, wrt performance, and sometimes even security, in 2.0 and 2.2

after 2.4 came out, most distros (and, i include this module in that umbrella) were put on a spot to support this strange new behaviour (secure, fast, incompatible ;)

tphoney:
@igalic @antaflos with the changes merged with #1447 is there any need for this pr ?

igalic:
@tphoney yes, this should be merged into 2.x

sathieu:
Opinion: I think reasonable default performance-wise and security-wise changes can go in 1.x.

This is like 932f0d0 to me.

igalic:
…except that with keepalive on (and the *wrong* keepalivetimeout setting [and the wrong mpm]) you're exposing yourself to resource starvation attacks

PR #1435, opened 2016-04-20
(maint) update apache::default_mods tests to work on debian 7

DavidS:
Debian 7 and Ubuntu 12.04's init scripts will refuse to restart the service
if the configuration is broken. So this commit adds an explicit
configuration step to shutdown the apache before the test.


gregohardy:
Cleaner stop +1

PR #1436, opened 2016-04-20
Support the newer mod_auth_cas config options

pcfens:
mod_auth_cas supports a number of [per-directory options](https://github.com/Jasig/mod_auth_cas/blob/master/README#L333-384)

A [number of the options](https://github.com/Jasig/mod_auth_cas/blob/master/README#L386-413) aren't supported in directory stanzas despite what the documentation suggests. The only way I was able to figure out what was supported was through trial and error. I've added the non-directory directives to the global and vhost configurations so that they're still available.


bmjen:
Thanks @pcfens, Overall this looks good. The new params need to be add to the README though. Thanks!

pcfens:
I've added the new parameters to the README, squashed, rebased, and pushed. Thanks for the feedback.

tphoney:
Thanks for the work !!! it is much appreciated !

PR #1440, opened 2016-04-22
Set DAV parameters in a directory block

smoeding:
Currently the module supports the `mod_dav` and `mod_dav_fs` modules and also allows setting the `DAVLockDB` parameter. But actually activating DAV for a directory only seems to be possible by providing a custom fragment.

This PR adds support for the `dav` parameter and two other DAV related parameters in a vhost directory block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2916,6 +2916,18 @@ ProxyStatus On',
 }
 ```
 
+###### `dav`
+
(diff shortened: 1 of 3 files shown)


jonnytpuppet:
Closing due to #1454 which is a rebased version of this PR

PR #1444, opened 2016-04-26
setting options-hash in proxy_pass or proxy_match leads to syntax errors in Apache

zivis:
Revert "MODULES-2956: Enable options within location block on proxy_match"

this commit removes the location directive in _proxy.erb
proxy: remove workaround for old broken clients
99add117df452d2c513a3a7c36d9fd0416e696b9

this commit sets options in the location directive in _proxy.erb
MODULES-2956: Enable options within location block on proxy_match
db5b0bfb09959622ea1c6cc299b58057b94c588f

setting options-hash in proxy_pass or proxy_match now leads to syntax errors
in Apache like "AuthBasicProvider not allowed here"

### DIFF
--- a/README.md
+++ b/README.md
@@ -2435,7 +2435,7 @@ Specifies the destination address of a [ProxyPass](https://httpd.apache.org/docs
 
 ##### `proxy_pass`
 
-Specifies an array of `path => URI` values for a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) configuration. Default: undef. Parameters and location options can optionally be added as an array.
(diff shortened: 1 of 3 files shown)


antaflos:
Just to chime in here, reverting db5b0bfb09959622ea1c6cc299b58057b94c588f definitely seems like the right thing to do. It apparently was a crutch to work around the suboptimal behaviour of `_proxy.erb`.

But now that `_proxy.erb` no longer implements that ancient workaround of putting `ProxyPassReverse` inside its own `Location` directive there is really neither need nor use in trying to define options that are only valid inside a `Location` or `Directory` context. The `options` functionality for `proxy_pass` can be scrapped entirely.

Any options such as `AuthBasicProvider`, `AuthType`, `Require`, ... are supported by the `apache::vhost::directories` parameter, where they belong.

igalic:
:+1: thanks, @antaflos for jumping in

PR #1445, opened 2016-04-28
mod_event: do not set parameters twice

timogoebel:
The Name of the Parameters MaxRequestWorkers and MaxConnectionsPerChild  for mod_event changed between versions as the README already shows (mixed up though).
The apache module should not set both the old and the new parameter name, but instead just one of them.

The old names are still supported so it's better (and safe) to set the old names than both.

See:
* https://httpd.apache.org/docs/current/en/mod/mpm_common.html#maxconnectionsperchild
* https://httpd.apache.org/docs/current/en/mod/mpm_common.html#maxrequestworkers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1366,7 +1366,7 @@ Installs and manages [`mod_mpm_event`][]. You can't include both `apache::mod::e
 **Parameters within `apache::mod::event`**:
 
 - `listenbacklog`: Sets the maximum length of the pending connections queue via the module's [`ListenBackLog`][] directive. Default: '511'.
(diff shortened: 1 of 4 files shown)


jonnytpuppet:
@hunner @bmjen Do you agree that this is a breaking change as the defaults are being removed?

Thanks!

timogoebel:
@jonnytpuppet : Thanks for taking a look. I don't think this breaks anything. It just prefers the "newer" parameter names.

jonnytpuppet:
@timogoebel The default values are being removed for these values, though?
Thanks

timogoebel:
@jonnytpuppet : Yes, but  $maxconnectionsperchild = $maxrequestsperchild = 0 ($maxconnectionsperchild was 0 before the change) --> nothing really happend, the parameter is just not set twice

PR #1447, opened 2016-04-29
Allow setting KeepAlive related options per vhost

antaflos:
Introduce the parameters vhost::keepalive, vhost::keepalive_timeout and
vhost::max_keepalive_requests, which are all undef by default, meaning
the server-wide KeepAlive options will be in effect. This way the
KeepAlive settings can be changed on a per-vhost basis.

Includes updated documentation and basic spec tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2206,6 +2206,24 @@ apache::vhost { 'sample.example.net':
 }
 ```
 
+##### `keepalive`
+
+Determines whether to enable persistent HTTP connections with the [`KeepAlive`][] directive for the virtual host. Valid options: 'Off', 'On' and `undef`. Default: `undef`, meaning the global, server-wide [`KeepAlive`][] setting is in effect.
+
+Use the `keepalive_timeout` and `max_keepalive_requests` parameters to set relevant options for the virtual host.
+
+##### `keepalive_timeout`
+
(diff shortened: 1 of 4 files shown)


igalic:
What about keepalive as server setting?

antaflos:
You mean https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/init.pp#L63-L65?

igalic:
what i meant is, if we should change the defaults here https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/params.pp#L141-L143 to Apache httpd's actual defaults.

antaflos:
I it was decided in #1434 that changing the KeepAlive default value is a backwards-incompatible change, no?

igalic:
right. thank you for the patch

mmoll:
It would have been nice if true/false would have been used instead of 'On'/'Off'. How likely would such a change get accepted?

antaflos:
@mmoll I have used 'on'/'off' to be in line with the way the server-wide KeepAlive parameters are handled. And I think there are quite a few other boolean parameters that require 'on'/'off' instead of true/false.

But there is the [bool2httpd function](https://github.com/puppetlabs/puppetlabs-apache/blob/master/lib/puppet/parser/functions/bool2httpd.rb) which, I realise, I should have used.

mmoll:
@antaflos I just stumbled over it, as I wanted to use these new params in theforeman-foreman, where we use booleans and a custom template fragment, it's not a big thing. :)

igalic:
i think by now we have a any2bool function in stdlib, so in the future, that could be used everywhere
i didn't quite see the light back then when I wrote that function, in anger ;)

PR #1450, opened 2016-05-06
Add class apache::vhosts to create apache::vhost resources

gerhardsam:
This way it is possible to create apache::vhost resources directly via the main class without declaring them in a separate role or profile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -741,6 +741,7 @@ apache::balancer { 'puppet01':
 - [**Public classes**](#public-classes)
     - [Class: apache](#class-apache)
     - [Class: apache::dev](#class-apachedev)
+    - [Class: apache::vhosts](#class-apachevhosts)
     - [Classes: apache::mod::\*](#classes-apachemodname)
 - [**Private classes**](#private-classes)
     - [Class: apache::confd::no_accf](#class-apacheconfdno_accf)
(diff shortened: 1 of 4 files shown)


jonnytpuppet:
@gerhardsam Thank you for your efforts, they are very much appreciated!

It would be appreciated if you could give a little context to why this feature would be beneficial to all users. What are the issues related to using the apache::vhost type to obtain the desired functionality?

Thanks

smoeding:
With this change you can use the `apache::vhosts` key in hiera to specify the vhosts to create. So in many cases it would be sufficient to declare the `apache` class and have all parameters in hiera. Currently an additional class is needed to define your vhosts from hiera (since a defined type does not use hiera for data binding: https://ask.puppet.com/question/1549/how-to-create-an-apache-vhost-in-hiera-when-using-puppetlabsapache/).

jonnytpuppet:
@smoeding What we normally do if you wish for heira compatibility, is to create a new .pp file and call create_resources().

I personally do not like the duplicated behaviour which could confuse users.

@hunner @bmjen any thoughts on this?

Thanks!

bmjen:
I agree with @jonnytpuppet here. I'd prefer to keep the current behavior and not overcomplicate the apache class. This is traditionally the model we use in other supported modules, and as mentioned by JT, has a working model for hiera.

gerhardsam:
@jonnytpuppet Thanks for the feedback and sorry for not answering earlier.
The context is the following: We use Satellite 6 (which acts ENC) which allows only to add classes but not defined types. Therefore it is like @smoeding depicted that I had to write an extra wrapper class to create the defined resource apache::vhost. Because I wanted to get rid of this necessity I wanted to mimic the pattern which was is used in the [puppet-nginx module](https://github.com/jfryman/puppet-nginx/blob/master/manifests/init.pp#L299-304).

@bmjen I do not think that is an overcomplication of the apache class but as an option I could add a class apache::vhosts which would create the apache:vhost resources instead.
Would that be a viable option?

jonnytpuppet:
@gerhardsam I honestly believe the best way forward is to create a wrapper class in a seperate .pp file that makes a call out to create_resources, then uses the results to call apache::vhost

An argument could be made that this should actually fall outside of the module altogether, and should be part of your own environment manifests, however we have president for having a wrapper class inside a module in a separate .pp file, which I'd be willing to merge :)

Thanks!

gerhardsam:
@jonnytpuppet I extracted the creation of the apache::vhost resources in the new class apache::vhosts. Does that conform to your expectation?

I am not sure yet, whether to create a [separate acceptance test](https://github.com/gerhardsam/puppetlabs-apache/blob/vhosts/spec/acceptance/vhosts_spec.rb) or whether to keep it in [spec/acceptance/vhost_spec.rb](https://github.com/gerhardsam/puppetlabs-apache/blob/vhosts/spec/acceptance/vhost_spec.rb)?

jonnytpuppet:
@gerhardsam You still have references to vhosts in the apache class. Shouldn't these be removed now?

Thanks

gerhardsam:
@jonnytpuppet I removed the references in the apache class, added docs and squashed the commits.
Would you mind reviewing the change? Thanks.

gerhardsam:
@jonnytpuppet Thanks for merging!

PR #1452, opened 2016-05-09
Only set SSLCompression when it is set to true.

buzzdeee:
The default for SSLCompression in Apache is 'Off' anyways, see:
https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcompression
therefore there is no real need to add that into the config file.

This will prevent problems on Apache versions that have
linked against SSL libraries that do not have SSLCompression
enabled. There, even SSLCompression Off will lead to:

H00526: Syntax error on line 14 of /etc/apache2/modules/ssl.conf:
Setting Compression mode unsupported; not implemented by the SSL library

and prevents Apache startup

Add some spec tests for that behaviour, depending on whether Apache
version >= 2.4 or not, since the template also differentiates.

### DIFF
--- a/templates/mod/ssl.conf.erb
+++ b/templates/mod/ssl.conf.erb
@@ -12,7 +12,9 @@
   SSLSessionCacheTimeout <%= @ssl_sessioncachetimeout %>
   <%- if scope.function_versioncmp([@_apache_version, '2.4']) >= 0 -%>
   Mutex <%= @_ssl_mutex %>
+    <%- if @ssl_compression -%>
   SSLCompression <%= scope.function_bool2httpd([@ssl_compression]) %>
+    <%- end -%>
   <%- else -%>
   SSLMutex <%= @_ssl_mutex %>
   <%- end -%>
--- a/spec/classes/mod/ssl_spec.rb
+++ b/spec/classes/mod/ssl_spec.rb
@@ -133,6 +133,46 @@ describe 'apache::mod::ssl', :type => :class do
       it { is_expected.to contain_file('ssl.conf').with_content(/^  SSLPassPhraseDialog builtin$/)}
     end
 
+    context "with Apache version < 2.4" do
+      let :params do
+        {
+          :apache_version => '2.2',
+        }
+      end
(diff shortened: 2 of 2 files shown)


buzzdeee:
rebased

templates/mod/ssl.conf.erb
```diff
@@ -12,7 +12,9 @@
   SSLSessionCacheTimeout <%= @ssl_sessioncachetimeout %>
   <%- if scope.function_versioncmp([@_apache_version, '2.4']) >= 0 -%>
   Mutex <%= @_ssl_mutex %>
+    <%- if @ssl_compression -%>
```
igalic: this checks for truthiness. if it evaluates to falsey, this option will not be set (to `off`)

buzzdeee: and that's exactly what I wanted, and wrote the spec tests for it ;)

I'm on OpenBSD*, and there Apache 2.4 from packages is built against libressl, which doesn't support SSLCompression at all. 

Therefore, when the module sets 
SSLCompression off

I get this error message on apache startup, and it refuses to start:

H00526: Syntax error on line 14 of /etc/apache2/modules/ssl.conf:
Setting Compression mode unsupported; not implemented by the SSL library

The default value for SSLCompression, if not given in the configuration file, is 'off', see:
https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcompression

The puppet module defaults the $ssl_compression parameter of the class to false, following the 
Apache default value, which is fine. But since it's the default value, it is not necessary to explicitly 
set it to 'off'. That said, for every Apache/SSL combination, that supports SSLCompression, nothing changes, for Apache/SSL combination, that doesn't support SSLCompression, things are not broken anymore (:
People that want to have SSLCompression, have to enable it explicitly in any case anyways, and likely know if their Apache/SSL combination supports it. For those the statement will be added to the config file as it was done before.

Does that make sense?

* yes, yes, OpenBSD is not (yet ;) supported by the module. But there may be other OS/Distros that run into the same issue...

PR #1458, opened 2016-05-16
Remove duplicate shib2 hash element

domcleal:
Prevents warning under Puppet 4:

     The key 'shib2' is declared more than once at modules/apache/manifests/params.pp:241:21


DavidS:
Thanks!

PR #1463, opened 2016-05-19
apache::balancer now respects apache::confd_dir

traylenator:
Previously with a configuration of

```
class{'apache':
  confd_dir => '/my/directory':
}
balancer{'myb':}
```

balencer would create a file incorrectly in the default location
*/etc/apache2/conf.d/balancer_myb.conf*

balencer now uses the value *apache::confd_dir*.

A test case has been added for this bug.

### DIFF
--- a/spec/defines/balancer_spec.rb
+++ b/spec/defines/balancer_spec.rb
@@ -1,8 +1,8 @@
 require 'spec_helper'
 
 describe 'apache::balancer', :type => :define do
-  let :pre_condition do
-    'include apache'
+  let :title do
+    'myapp'
   end
   let :facts do
     {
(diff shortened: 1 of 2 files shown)


spec/defines/balancer_spec.rb
```diff
@@ -1,33 +1,58 @@
...
-        :target => '/tmp/myapp.conf'
+        :osfamily               => 'Debian',
+        :operatingsystem        => 'Debian',
+        :operatingsystemrelease => '6',
+        :lsbdistcodename        => 'squeeze',
+        :id                     => 'root',
+        :concat_basedir         => '/dne',
+        :path                   => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+        :kernel                 => 'Linux',
+        :is_pe                  => false,
```
DavidS: These seem to be the same facts as above? Please put them into the outer `describe` block, and avoid the duplication.

DavidS:
Good catch, and thanks for the test!

spec/defines/balancer_spec.rb
traylenator: Yes - fixed and re-based.

DavidS:
Sweet, thanks!

PR #1468, opened 2016-05-24
Manage default root directory access rights

smoeding:
Currently the root directory setting in httpd.conf provides an open access policy. By default Apache has access to all directories on the host. In some cases it may be desirable to deny access by default and explicitly configure access rights to every resource.

This patch adds the boolean `root_directory_secured` parameter to the main class. Setting this parameter to `true` will add a `Deny from all` (Apache 2.2) or a `Require all denied` (Apache 2.4) directive to the Apache config for the root directory. Setting the parameter to `false` will leave out these directives. The parameter default is `false` to remain backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1193,6 +1193,10 @@ Sets the desired permissions mode for config files, in symbolic or numeric notat
 
 Array of the desired options for the / directory in httpd.conf. Defaults to 'FollowSymLinks'.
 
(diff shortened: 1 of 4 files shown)


DavidS:
The travisci failure is unrelated to this change. Thanks very much for your contribution.

PR #1470, opened 2016-06-03
allow for clean up of symlinked conf files

adamcousins:
allow to purge symlink conf file.

use case AWS linux with php55


adamcousins:
not working as expected

PR #1471, opened 2016-06-06
Add apache::mod::socache_shmcb so it can be included multiple times

mpdude:
I am not sure if this is the best way to address this: I already got `::apache::mod { 'socache_shmcb': }` in my Apache profile classes because I need it to support database-backed authentication.

Now I would like to include `apache::mod::ssl` as well which in turn also declares `::apache::mod { 'socache_shmcb': }`.

If there is a better (more "Puppet" way) of solving this, please let me know. Otherwise, I hope that this PR allows to simply `include apache::mod::socache_shmcb` and be set.

### DIFF
--- a/manifests/mod/socache_shmcb.pp
+++ b/manifests/mod/socache_shmcb.pp
@@ -1,0 +1,3 @@
+class apache::mod::socache_shmcb {
+    ::apache::mod { 'socache_shmcb': }
+}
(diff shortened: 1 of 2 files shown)


mpdude:
Wait, is that too complicated and would work without the `manifests/mod/socache_shmcb.pp` file as well?

PR #1474, opened 2016-06-08
Allow for setting SecAuditLogRelevantStatus

kimor79:


bmjen:
Thanks for the contribution @kimor79 ! Do you mind adding a simple unit test for this new parameter? Perhaps a line setting a custom Status regex here https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/security_spec.rb#L103 and validating that it is seeded into the conf file correctly. Thanks!

sathieu:
I've written #1487 with similar goal (it's more complete as is includes a README.

DavidS:
Closing to prefer rebased and unified #1494

PR #1477, opened 2016-06-13
variety of xenial fixes

tphoney:


igalic:
@tphoney whee thanks a lot for this!

tphoney:
@igalic that was not fun.

PR #1478, opened 2016-06-15
Fix PassengerRoot under Debian stretch

sathieu:


mmoll:
As all newer versions have locations.ini, it could make sense to double-reverse the logic here to be more future-safe.

sathieu:
@mmol: hello again ;-) Done.

tphoney:
woops, i made a bad click. apologies

tphoney:
@sathieu great work ! thanks for the change, i can forsee this being expanded out again, at some point. For now, it looks a lot better.

PR #1490, opened 2016-06-21
Allow to set SecAuditLog

sathieu:


manifests/vhost.pp
```diff
@@ -397,6 +405,21 @@
     }
   }
 
+  if $modsec_audit_log_file {
```
bmjen: @sathieu This logic is confusing. You should always check the "manage" boolean first and determine path variables after. Something like:

```
if $mod_audit_log {
  if $modsec_audit_log_file {
    $modsec_audit_log_destination = "${logroot}/${modsec_audit_log_file}"
  } else if $modsec_audit_log_pipe { 
    $modsec_audit_log_destination = $modsec_audit_log_pipe
  } else {
    if $ssl {
       $modsec_audit_log_destination = "${logroot}/${name}_security_ssl.log"
     } else {
       $modsec_audit_log_destination = "${logroot}/${name}_security.log"
     }
} else {
    $modsec_audit_log_destination = undef
}
```

This way you only set variables/paths if `$modsec_audit_log` is set to true.

manifests/vhost.pp
```diff
@@ -243,6 +247,10 @@
     fail("Apache::Vhost[${name}]: 'error_log_file' and 'error_log_pipe' cannot be defined at the same time")
   }
 
+  if $modsec_audit_log_file and $modsec_audit_log_pipe {
```
bmjen: This validation should be ignored if `$modsec_audit_log` is `false`.

sathieu: I kept the same logic as access_log and error_log (just above).

manifests/vhost.pp
sathieu: My intention was that you only need to set one attribute to enable logging.

I've changed the code a bit to improve this.

PR #1499, opened 2016-07-07
Auto load Apache::Mod[slotmem_shm] and Apache::Mod[lbmethod_byrequest…

sathieu:
…s] with proxy_balancer on 2.4


binford2k:
Can you update your spec test with something like https://github.com/puppetlabs/puppetlabs-apache/blob/master/spec/classes/mod/worker_spec.rb#L4-L6 so the variable is available to use?

hunner:
And could you document the new parameter in README.md similar to the other mods with parameters? Thanks!

sathieu:
@hunner @binford2k  Comments addressed.

HelenCampbell:
Thanks for your contribution!

PR #1500, opened 2016-07-08
Sync security_crs.conf with latest upstream

sathieu:
This effectively add the real_ip variable (and anomaly_score, sql_injection_score, xss_score=0, inbound_anomaly_score, outbound_anomaly_score).


sathieu:
From https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/b375fbf7701aa43a849d225f54a4c2c4f4fd2c8d/modsecurity_crs_10_setup.conf.example

PR #1503, opened 2016-07-11
mod_passenger: PassengerMaxInstancesPerApp option

ygt-davidstirling:
PassengerMaxInstancesPerApp has been available [since version Passenger 3.0] (https://www.phusionpassenger.com/library/config/apache/reference/#passengermaxinstancesperapp)


hunner:
Cool! Could you add docs and update the spec test for this? Thanks!

ygt-davidstirling:
@hunner: good point about the docs, I've updated the readme now.
I already added a test in  `spec/classes/mod/passenger_spec.rb` (in the same way as the `passenger_max_pool_size` param was tested) so is there further testing you think necessary?

PR #1510, opened 2016-08-04
Add some more passenger params

Reamer:
For passenger NodeJS Apps we need some more params
- passenger_nodejs (https://www.phusionpassenger.com/library/config/apache/reference/#passengernodejs) - Available with 4.0.24
- passenger_sticky_sessions (https://www.phusionpassenger.com/library/config/apache/reference/#passengerstickysessions) - Available with 4.0.45
- passenger_startup_file (https://www.phusionpassenger.com/library/config/apache/reference/#passengerstartupfile) - Passenger has a good default, but for Express app generator we must set this param. Available with 4.0.25

If you need anything, let me know

Greetings Reamer


Reamer:
Just rebased.
If you need anything let me know
Greetings Reamer

PR #1513, opened 2016-08-15
Allow the proxy_via setting to be configured

electrical:
Previously it was hardcoded, this patch makes it configurable.


hunner:
Thanks! Could you add documentation for this new parameter?

electrical:
@hunner done. Hope the test passes now.

electrical:
@hunner anything that i need to do on this?

PR #1515, opened 2016-08-21
Update trace option to Off

chris1984:
Requesting to change   $trace_enable = 'Off',

This has caused some issues with customers using Red Hat Satellite 6 and causing security scanner software to raise red flags. Here is the downstream BZ for reference. 

https://bugzilla.redhat.com/show_bug.cgi?id=1305782


CalvinHartwell:
+1 can confirm this is causing red flags with security scanners for Red Hat Satellite 6.1

PR #1522, opened 2016-09-01
Corrected typo/ Needed parens for definition

sandersonjobs:
Using class {'::apache::namevirtualhost'} presently returns saying the class is undefined.


smortex:
This PR can be closed: `apache::namevirtualhost` is not a class but a [defined type](https://docs.puppet.com/puppet/4.10/lang_defined_types.html) and is not usable as expressed by @ghost.

PR #1523, opened 2016-09-01
Adding Amazon Linux support

seefood:
One more fork in the repo definition.


seefood:
I stand corrected. this may not be enough. for some versions of Amazon Linux the RPM dependencies break for the passenger package and I can't use the official repo. I'm having to go down the gem road instead. I'll reopen when I have a better branch.

PR #1524, opened 2016-09-15
Allow no_proxy_uris to be used within proxy_pass

ChristophBerg:
Previously, proxy_pass and proxy_pass_match configuration blocks did not
support using no_proxy_uris and no_proxy_uris_match. These parameters
required the use of proxy_dest, but mixing proxy_pass and proxy_dest
results in undesired duplication of ProxyPass config lines.

Fix by allowing no_proxy_uris and no_proxy_uris_match to be used within
proxy_pass and proxy_pass_match configuration blocks.

This PR is similar to #1521 but addresses the problem better in my opinion.


bmjen:
Closes #1521

PR #1525, opened 2016-09-15
Bug - Port numbers must be quoted

blackknight36:
The default port numbers for apache must be quoted to avoid errors
when running a catalog in the puppet agent.  This fixes https://tickets.puppetlabs.com/browse/MODULES-3871

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -384,7 +384,7 @@ class apache (
 
     ::apache::vhost { 'default':
       ensure          => $default_vhost_ensure,
-      port            => 80,
+      port            => '80',
       docroot         => $docroot,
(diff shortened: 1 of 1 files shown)


JaminShanti:
This fix yet?  It's 2017.
0.9.0 still showing this error:
==> default: Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error creating type specialization of an Apache::Listen-Type, Cannot use Integer where a resource title String is expected at /etc/puppetlabs/puppet/environments/production/modules/apache/manifests/vhost.pp:252:33  at /etc/puppetlabs/puppet/environments/production/modules/apache/manifests/init.pp:237 on node localhost

PR #1526, opened 2016-09-21
[MODULES-3882] Don't write empty servername for vhost to template

JAORMX:
If an empty servername is given to a vhost, it shouldn't be written
to the configuration file. Else we'll end up with an invalid
configuration and httpd will fail to start. It is actually valid not
to have a Servername, and in this case, httpd will route only based
on the IP.

This resolves bug https://tickets.puppetlabs.com/browse/MODULES-3882


EmilienM:
:+1: Please rename the PR to "[MODULES-3882] ..."

Thanks!

PR #1529, opened 2016-10-03
Let SSLCertificateKeyFile be optional

westbywest:
The parameter SSLCertificateKeyFile is actually optional.  A private key doesn't need to be supplied when using wildcard SSL certs, for example.  This PR patches the module to permit specifying ssl_key => false for appache::vhost, to work around the otherwise hardcoded default of .../localhost.key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3608,7 +3608,7 @@ Sets the certificate revocation check level via the [SSLCARevocationCheck direct
 
 ##### `ssl_key`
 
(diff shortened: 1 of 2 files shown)


antaflos:
Why do you think that a wildcard certificate does not require a private key? Why would you think any server certificate would work without a private key? That makes no sense.

You may be confusing this with the possibility to provide the private key in the same file that contains the server certificate, i.e. `SSLCertificateFile`, but doing so is [highly discouraged](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslcertificatekeyfile):

> The private key may also be combined with the certificate in the file given by SSLCertificateFile, but this practice is highly discouraged. If it is used, the certificate files using such an embedded key must be configured after the certificates using a separate key file.

bmjen:
@westbywest Alot of tests are failing. And also, have you been able to respond to @antaflos about the need for this feature?

PR #1530, opened 2016-10-06
Remove lines that create noise/spam on ubuntu

jamesmollison:
Revert the "puts" lines that cause unnecessary output/spam on any puppet run on ubuntu/xenial 

user@puppetagent-01:~$ sudo facter -p
Matching apachectl 'Server version: Apache/2.4.18 (Ubuntu)
Server built:   2016-07-14T12:32:26'
aio_agent_version => 1.7.0
apache_version => 2.4.18
apt_has_updates => true
... (etc)

### DIFF
--- a/lib/facter/apache_version.rb
+++ b/lib/facter/apache_version.rb
@@ -2,11 +2,9 @@ Facter.add(:apache_version) do
   setcode do
     if Facter::Util::Resolution.which('apachectl')
       apache_version = Facter::Util::Resolution.exec('apachectl -v 2>&1')
-      puts "Matching apachectl '#{apache_version}'"
       %r{^Server version: Apache\/(\d+.\d+(.\d+)?)}.match(apache_version)[1]
     elsif Facter::Util::Resolution.which('apache2ctl')
       apache_version = Facter::Util::Resolution.exec('apache2ctl -v 2>&1')
-      puts "Matching apache2ctl '#{apache_version}'"
       %r{^Server version: Apache\/(\d+.\d+(.\d+)?)}.match(apache_version)[1]
     end
   end


lib/facter/apache_version.rb
```diff
@@ -2,11 +2,9 @@
   setcode do
     if Facter::Util::Resolution.which('apachectl')
       apache_version = Facter::Util::Resolution.exec('apachectl -v 2>&1')
-      puts "Matching apachectl '#{apache_version}'"
```
wilson208: Wouldn't it be a better solution to change this to:
`Puppet.debug("Matching apachectl '#{apache_version}'")`
or
`Puppet.info("Matching apachectl '#{apache_version}'")`

See here http://www.rubydoc.info/gems/puppet/Puppet/Util/Logging#debug-instance_method

lib/facter/apache_version.rb
```diff
@@ -2,11 +2,9 @@
...
       %r{^Server version: Apache\/(\d+.\d+(.\d+)?)}.match(apache_version)[1]
     elsif Facter::Util::Resolution.which('apache2ctl')
       apache_version = Facter::Util::Resolution.exec('apache2ctl -v 2>&1')
-      puts "Matching apache2ctl '#{apache_version}'"
```
wilson208: See above :)

wilson208:
@jamesmollison Have left comments, the solution I outline would mean that these only show in when debugging is enabled.

smortex:
This PR can be closed as f6cfa03d5bcd3d184841c614b1bfb8046a0c7d03 fixed this issue by using `Facter.debug`.

PR #1537, opened 2016-10-14
Notify Apache of new symlink

nielslaukens:
The notify was present in the regular vhost.pp manifest, but not in the
vhost/custom.pp manifest


igalic:
```
Facter::Util::Fact
  apache_version
    with value

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

that can't be right…

nielslaukens:
I've tried to trigger a rebuild (since I think the error is transient), but can't find the button to do it...

igalic:
@nielslaukens only members of an org can do that (if they have merge access?)

i've triggered it for you now

PR #1541, opened 2016-10-19
[MODULES-1491] mod_dir is empty when instanciating apache::mod::...

intelradoux:
Add "include ::apache" on class that needed that.


snahelou:
+1

PR #1544, opened 2016-10-21
Bugfix: don't use a different name because port is not specified

nielslaukens:
When the port is not specified, the name of the <VirtualHost> tag is taken to
be the resource name, instead of the vhost name.


nielslaukens:
For future reference. I fixed my problem by specifying a `port`.

I still believe this bugfix is needed, but the acceptance tests fail. So either I am not understanding the concept correctly, or the tests are wrong. Either way, I will no longer pursue this PR.

PR #1547, opened 2016-10-28
Allow user to specify alternative package and library names for shibboleth module

tpdownes:
If left `undef`, they remain the values set in `params.pp`. The context is that more up-to-date packages are typically available from external repositories such as from [Switch](https://www.switch.ch/aai/guides/sp/installation/) but they use different package names (that force the OS packages to uninstall) and potentially different library files. These parameters allow a generic way to override the defaults without affecting existing configurations. I have tried to mimic the parameter names used in other modules under `manifests/mod/*` for consistency.

### DIFF
--- a/manifests/mod/shib.pp
+++ b/manifests/mod/shib.pp
@@ -1,6 +1,8 @@
 class apache::mod::shib (
   $suppress_warning = false,
-  $mod_full_path = undef,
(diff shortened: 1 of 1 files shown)


wilson208:
Thanks for the contribution! 👍

PR #1548, opened 2016-11-01
Support parameter PassengerDataBufferDir

seefood:
Support parameter PassengerDataBufferDir on the passenger module manifest and template.


wilson208:
@seefood could you document this in the README.md and I will get this merged :)

seefood:
Should be better now...

PR #1549, opened 2016-11-01
Support Passenger repo on Amazon Linux

seefood:
No systemd on AmaLinux, so it should get the el6 repo.


manifests/mod/passenger.pp
```diff
@@ -56,17 +56,32 @@
...
-      gpgcheck      => '0',
-      gpgkey        => 'https://packagecloud.io/gpg.key',
-      repo_gpgcheck => '1',
-      sslcacert     => '/etc/pki/tls/certs/ca-bundle.crt',
-      sslverify     => '1',
-      before        => Apache::Mod['passenger'],
+    if $::operatingsystem == 'Amazon' {
+      yumrepo { 'passenger':
+        ensure        => 'present',
+        baseurl       => 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/6Server/$basearch',
```
wilson208: If I am not mistaken, isn't the only change in this PR hardcoding 6Server rather than using $releasever on Amazon? Would it be tidier to move the base_url out to a separate variable, rather than duplicate the entire yumrepo resource? e.g. 
`if $::operatingsystem == 'Amazon' {`
`  $passenger_base_url = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/6Server/$basearch'`
`} else {`
`  $passenger_base_url = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/$releasever/$basearch'`
`}`

wilson208:
Left some comments in the code @seefood, let me know your thoughts :)

seefood:
Sure, I didn't dare refactor stuff just to get my patch in, I don't mind doing this "the right way" :)

seefood:
OK, should be better now.

seefood:
Halleluya! Just as we're getting rid of the AmaLin imagesin favor of CentOS...

PR #1552, opened 2016-11-16
Fix vhost template

iamspido:
fix vhost template file causing unexpected tIDENTIFIER


eputnam:
close/reopen to rekick travis

eputnam:
thanks @iamspido !

PR #1554, opened 2016-11-17
(MODULES-4049) SLES support

eputnam:


manifests/mod/info.pp
```diff
@@ -6,7 +6,22 @@
...
   $_apache_version = pick($apache_version, $apache::apache_version)
-  apache::mod { 'info': }
+
+  if $::osfamily == 'Suse' {
+    if defined(Class['::apache::mod::worker']){
+      $suse_path = '/usr/lib64/apache2-worker'
+    } elsif defined(Class['::apache::mod::prefork']){
+      $suse_path = '/usr/lib64/apache2-prefork'
+    } else {
+      $suse_path = pick('/usr/lib64/apache2-prefork', '/usr/lib64/apache2-worker')
```
hunner: technically this will always pick the first one, since it's never undef.

manifests/mod/info.pp
```diff
@@ -6,7 +6,22 @@
 ){
   include ::apache
   $_apache_version = pick($apache_version, $apache::apache_version)
-  apache::mod { 'info': }
+
+  if $::osfamily == 'Suse' {
+    if defined(Class['::apache::mod::worker']){
```
hunner: Hmm. `defined()` is parse-order dependent, but most likely this will work.

manifests/mod/info.pp
hunner: Also, since prefork would be the default case then you don't need the `elsif` above.

manifests/mod/php.pp
```diff
@@ -53,13 +53,12 @@
...
 
   if $::operatingsystem == 'SLES' {
-    $suse_lib_path = $::apache::params::suse_lib_path
       ::apache::mod { $mod:
         package        => $_package_name,
         package_ensure => $package_ensure,
         lib            => 'mod_php5.so',
         id             => "php${_php_major}_module",
-        path           => "${suse_lib_path}/mod_php5.so",
+        path           => "${$::apache::lib_path}/mod_php5.so",
```
hunner: wat? This isn't from a special path?

manifests/mod/ssl.pp
```diff
@@ -67,10 +67,25 @@
...
 
-  ::apache::mod { 'ssl':
-    package => $package_name,
+  if $::osfamily == 'Suse' {
+    if defined(Class['::apache::mod::worker']){
+      $suse_path = '/usr/lib64/apache2-worker'
+    } elsif defined(Class['::apache::mod::prefork']){
+      $suse_path = '/usr/lib64/apache2-prefork'
+    } else {
+      $suse_path = pick('/usr/lib64/apache2-worker', '/usr/lib64/apache2-prefork')
```
hunner: ditto

manifests/mod/cgi.pp
```diff
@@ -1,11 +1,18 @@
 class apache::mod::cgi {
+  $_lib_path = $::apache::lib_path
```
bmjen: Where is this being used?

manifests/mod/ssl.pp
```diff
@@ -67,10 +67,25 @@
...
+    }
+    ::apache::mod { 'ssl':
+      package  => $package_name,
+      lib_path => $suse_path
```
hunner: technically this resource declaration doesn't need to be duplicated because `lib_path => $suse_path` when undef would be as if it wasn't passed at all

manifests/mod/security.pp
```diff
@@ -27,23 +27,14 @@
     default => "${secdefaultaction},log",
   }
 
-  if $::osfamily == 'FreeBSD' {
+  if $::osfamily == 'FreeBSD' or ($::osfamily == 'Suse' and $::operatingsystemrelease < '11') {
```
bmjen: This should be in a separate block with it's own unique failure msg.

manifests/mod/worker.pp
```diff
@@ -118,11 +118,18 @@
         }
       }
     }
-    'debian', 'freebsd', 'Suse': {
+    'debian', 'freebsd' : {
```
bmjen: get rid of this space. /nit

manifests/mod/worker.pp
```diff
@@ -118,11 +118,18 @@
...
       ::apache::mpm{ 'worker':
         apache_version => $apache_version,
       }
     }
+    'Suse' : {
```
bmjen: this space too. /nit

manifests/mod/worker.pp
```diff
@@ -118,11 +118,18 @@
...
+      ::apache::mpm{ 'worker':
```
bmjen: add a space here. /nit

manifests/mpm.pp
```diff
@@ -120,10 +137,8 @@
         }
       }
 
-      if versioncmp($apache_version, '2.4') < 0 {
```
bmjen: hmm.. this change wasn't SLES specific...are we changing debian behavior here on purpose?

spec/spec_helper_acceptance.rb
```diff
@@ -40,6 +40,12 @@
       shell 'apt-get install -y ssl-cert'
     end
 
+    if fact('operatingsystem') == 'SLES'
+      #generate certs for SLES
+      shell 'zypper install -y apache2'
```
bmjen: hmm... is this what we want to do here? Doesn't this circumvent the module installing apache using the platform provider?

hunner:
👍 with bmjen's comments

spec/spec_helper_acceptance.rb
eputnam: i wasn't sure what to do here. gensslcert is part of apache2 and seemed like the easiest route to uniform certificate creation across SLES versions. there is the whole circumventing the module thing. i could move gensslcert activity into only the tests that require it.

bmjen: 👍  on making it isolated to test environments. I think installing/managing packages like that fall under the scope of the users' sys administration.

manifests/mpm.pp
eputnam: it's difficult to see, but it is in the SLES block

bmjen: haha. doh. stupid github code folding...

PR #1555, opened 2016-11-22
Don't fail if first element of  is not an hash before flattening

sathieu:
The template is using [@rewrites].flatten.compact which maps to Puppet's delete_undef_values(flatten([]))

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -192,7 +192,8 @@ define apache::vhost(
   if $rewrites {
     validate_array($rewrites)
     unless empty($rewrites) {
-      validate_hash($rewrites[0])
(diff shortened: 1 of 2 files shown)


eputnam:
@sathieu thanks for the PR! In what situation would the first element not be a hash?

sathieu:
@eputnam : When the directives param is made from several variables (like [$a, $b,$c]), and some of them are undef, of an array

hunner:
Looks like `delete_undef_values` was added to stdlib in 4.2.0, so could you bump the lower bound in metadata.json for that? I think no one uses stdlib 2 or 3 any more that would also use a modern apache version, so we'll call it a bugfix.

PR #1556, opened 2016-11-22
Move mod_env in its own class and load it when required

sathieu:


sathieu:
The tests are probably failing for unrelated reason (connection error).

manifests/vhost.pp
```diff
@@ -539,11 +539,19 @@
...
+  $use_env_mod = $setenv and ! empty($setenv)
+  if ($use_env_mod) {
+    if ! defined(Class['apache::mod::env']) {
+      include ::apache::mod::env
+    }
+  }
   # Check if mod_setenvif is required and not yet loaded.
   # create an expression to simplify the conditional check
-  $use_setenv_mod = ($setenv and ! empty($setenv)) or ($setenvif and ! empty($setenvif)) or ($setenvifnocase and ! empty($setenvifnocase))
+  $use_setenvif_mod = ($setenvif and ! empty($setenvif)) or ($setenvifnocase and ! empty($setenvifnocase))
```
bmjen: @sathieu  This appears to be a change in behavior. Previously, the `use_setenv_if_mod` would only be `true` if `use_setenv_mod` were true. Now it is evaluated independently and below... can add the config fragment without the setting of `$setenv`.

Is this desired behavior?

spec/acceptance/default_mods_spec.rb
```diff
@@ -34,12 +34,18 @@ class { 'apache':
           default_mods => false,
         }
         apache::vhost { 'defaults.example.com':
-          docroot => '#{$doc_root}/defaults',
-          aliases => {
+          docroot     => '#{$doc_root}/defaults',
```
bmjen: indentation is wrong here.

bmjen:
Ping @sathieu. I'm hoping to cut a release today and would like to include this PR.

manifests/vhost.pp
sathieu: Previously mod_setenvif was loaded when only setenv was used, which is not needed at all.

The new behavior handle the modules independently (see [mod_env](http://httpd.apache.org/docs/2.4/mod/mod_env.html) and [mod_setenvif](http://httpd.apache.org/docs/2.4/mod/mod_setenvif.html)).

So yes: behavior desired.

spec/acceptance/default_mods_spec.rb
sathieu: Fixed

PR #1561, opened 2016-11-29
Add SSL stapling

tjikkun:
It is possible to set a global default stapling on/off
Per vhost this can be overridden.

Based on initial pull request https://github.com/puppetlabs/puppetlabs-apache/pull/907 by David Teirney <david.teirney@orionhealth.com>


ekohl:
Maybe mention that this is a 2.4-only feature? Would the readme suffice?

tjikkun:
@ekohl Good one, added to README

PR #1562, opened 2016-11-30
mod_proxy_balancer manager

sathieu:


sathieu:
I'm tempted to enable it by default. The current commit doesn't change default behavior.

PR #1565, opened 2016-12-02
custom facts shouldn't break structured facts

igalic:
Change the `puts` output to `Facter.debug` so that our structured facts remain side-effect free in normal runs, and can just be piped into tools like `jq`

### DIFF
--- a/lib/facter/apache_version.rb
+++ b/lib/facter/apache_version.rb
@@ -2,11 +2,11 @@ Facter.add(:apache_version) do
   setcode do
     if Facter::Util::Resolution.which('apachectl')
       apache_version = Facter::Util::Resolution.exec('apachectl -v 2>&1')
-      puts "Matching apachectl '#{apache_version}'"
+      Facter.debug "Matching apachectl '#{apache_version}'"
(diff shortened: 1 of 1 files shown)


igalic:
before:
```
igalic@app02 ~> sudo -H /opt/puppetlabs/bin/puppet facts | jq .values.networking.primary
parse error: Invalid numeric literal at line 1, column 9
Error: Broken pipe @ io_write - <STDOUT>
Error: Try 'puppet help facts find' for usage
```
because:
```
igalic@app02 ~> sudo -H /opt/puppetlabs/bin/puppet facts
Matching apachectl 'Server version: Apache/2.4.20 (Ubuntu)
Server built:   2016-04-20T00:00:00'
{
  "name": "networking.primary",
  "values": {
    "aio_agent_version": "1.8.1",
    "apache_version": "2.4.20",
    "apt_has_updates": false,
    "apt_reboot_required": false,
    "apt_update_last_success": 1480660628,
    "architecture": "amd64",
```

after:

```
igalic@app02 /e/p/c/e/p/m/apache ((7d7007e9…))> sudo -H /opt/puppetlabs/bin/puppet facts | jq .values.networking.primary
"eth0"
```

PR #1568, opened 2016-12-09
apache::mod::auth_kerb requires apache::mod::authn_core (on >=2.4)

sathieu:


sathieu:
auth_kerb has no use with AuthType/AuthName directives.

PR #1569, opened 2016-12-12
(MODULES-4213) Allow global rewrite rules inheritance in vhosts

EmersonPrado:
Apache module rewrite allows virtual hosts to inherit global rules
if their conf files contain the directive "**RewriteOptions Inherit**"
among rewrite rules

This change includes this directive depending on a boolean parameter
in `apache::vhost` class - **rewrite_inherit** - which defaults to false

### DIFF
--- a/README.md
+++ b/README.md
@@ -2868,6 +2868,35 @@ apache::vhost { 'site.name.fdqn':
 
 Refer to the [`mod_rewrite` documentation][`mod_rewrite`] for more details on what is possible with rewrite rules and conditions.
 
+##### `rewrite_inherit`
+
+Determines whether the virtual host inherits global rewrite rules. Default: false.
+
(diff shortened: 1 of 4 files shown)


bmjen:
Thanks @EmersonPrado ! Can you update the README and add a spec test for the new parameter?

EmersonPrado:
@bmjen I added a README and a spec test, but:
1. I tend to overkill at READMEs. Feel absolutely free to criticize.
1. I'm absolutely clueless on spec tests. I copied, pasted and edited what I think is right in the vhost defined type test. And I can't test it on my workstation, because it requires puppet-blacksmith, which is only installable by gem, and I'm behind a proxy. I just can't make it work. So I'll do some trial and error passes on this specs. Maybe I get something.

EmersonPrado:
@bmjen Gave up completely on spec testing a file content added by template. I don't even know if spec tests can do that. How can I back off to commit 7d977e7?

bmjen:
@EmersonPrado No Problem! I will just squash it all, and I'll add something simple to the test case in a separate PR.

PR #1572, opened 2016-12-13
[#puppethack] MODULES-2872 - clarified the use of "default_vhost => false"

genebean:


DavidS:
docs-only PR merged before travis finished.

PR #1581, opened 2016-12-22
Add FileETag

kuchosauronad0:
There should be a option to set the FileETag option in the global webserver configuration. If you need anything, let me know

o/


tphoney:
looks good, docs / tests everything 👍

PR #1585, opened 2017-01-02
[apache::mod::cgi] Fix: ordering constraint for mod_cgi

punycode:
`mod_cgi` is used on other MPM implementations than just `prefork`,
so we have to enforce resource ordering against these too.


eputnam:
@punycode thanks for the PR!

PR #1588, opened 2017-01-06
Add mod_{authnz_pam,intercept_form_submit,lookup_identity}

ekohl:
I added these modules based on what we have in https://github.com/theforeman/puppet-foreman/blob/25dda6012a476829265fb5b4d8c959be03cd4d3e/manifests/install.pp#L44-L46. It should be noted that from that I know the Red Hat packages must work. For Debian I only looked at the packages names and they should install.


ekohl:
Updated per discussion with @hunner

mmoll:
What's the status here?

PR #1591, opened 2017-01-13
AllowOverride is configured  by using allow_override

dal00:
Correcting the documentation to reflect the code.


wilson208:
@dal00 you are correcting the documentation for the apache::vhost defined type. However if you look [here](https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/vhost.pp#L45) the parameter actually used is `override` and not `allow_override`.

However, if you are specifying the `directories` parameter of `apache::vhost`, the `allow_override` param is used and the documentation looks correct for this.

If I have got this wrong feel free to correct me and re-open

Thanks

dal00:
No, you're right! :)

PR #1592, opened 2017-01-18
add parser function apache_pw_hash

aptituz:
This commit adds a new parser function apache_pw_hash, which
allows generating password hashes to be used in htpasswd files.


eputnam:
@aptituz thanks for this, it looks great. could you rebase so we can run against the latest travis config?

aptituz:
@eputnam Sure. There you go. :)

eputnam:
thanks @aptituz !

PR #1594, opened 2017-01-24
ensure csr_package is 'installed' instead of 'latest'

pyther:
mod/security.pp is the only module that ensures the package is `latest`. This change makes the behavior of `mod/security.pp` consistent.

Other modules ensure `present` or `installed`
- mod.pp = present
- mod/php.pp = present
- mod/mime.php = installed

### DIFF
--- a/manifests/mod/security.pp
+++ b/manifests/mod/security.pp
@@ -53,7 +53,7 @@ class apache::mod::security (
 
   if $crs_package  {
     package { $crs_package:
-      ensure => 'latest',
+      ensure => 'installed',
       before => [
         File[$::apache::confd_dir],
         File[$modsec_dir],


pyther:
In my environment, our system patching procedure is to create quarterly snapshots. When we roll over to the new repositories (update the yum config) we have other process that actually updates the machine.

- Rolled Over our repostories
- Puppet ran, updating mod_security_csr (because of `ensure => latest`)
- Apache failed to restart with the following configuration error
```
Jan 24 16:00:13 web-dev.example.com httpd[11964]: AH00526: Syntax error on line 97 of /etc/httpd/modsecurity.d/security_crs.conf:
Jan 24 16:00:13 web-dev.example.com httpd[11964]: ModSecurity: Found another rule with the same id
```
- Package installs `/etc/httpd/modsecurity.d/modsecurity_crs_10_config.conf`
- Run puppet again, puppet remove `/etc/httpd/modsecurity.d/modsecurity_crs_10_config.conf` and restart is successful

This broke our production web servers.

I think there are two issues here:
1. The package behavior (install vs latest) is inconsitent from the rest of the classes that install packages (this PR fixes this)
2. The package should be installed  before managing `$modsec_dir`

gdubicki:
I stumbled upon the `ModSecurity: Found another rule with the same id` issue too. But for  me even upgrade to current `1.11.0` version of this module didn't help. But it was probably because I started with `1.4.1` and solutions mentioned in https://tickets.puppetlabs.com/browse/MODULES-3744 were sensitive to existing files. Probably cleaning `/etc/httpd` or other config dirs would help.

tphoney:
Thanks for the change @pyther

PR #1595, opened 2017-01-30
Document mod_rewrite with the rest of modules

pedroadame:
The class `apache::mod::rewrite` is only documented in README as a note in the `rewrites` option for Virtualhosts. My changes make it more visible in place with the rest of mods.

Did it because took me an hour of in-depth lecture of README to find that.


HelenCampbell:
Merged - Thank you!

PR #1597, opened 2017-02-20
Errror : To configure a mix of SSL and unencrypted virtual hosts at the same domain,

vineeth777:
I have faced while configuring SSL and unencrypted virtual hosts at the same domain,

Initial manifest,

```
include apache
include apache::mod::rewrite

apache::vhost { 'site.mydomain.com non-ssl':
 port         => '80',
 docroot      => '/var/www/site',
  rewrite_rule    => '(.*) https://site.mydomain.com [R,L]',
  redirect_status => 'permanent',
  redirect_dest   => 'https://site.mydomain.com'
}

apache::vhost { 'site.mydomain.com ssl':
  port          => '443',
  ssl           => true,
  docroot       => '/var/www/site',
  docroot_owner => 'www-data',
  docroot_group => 'www-data',
}
```

**Apache Error,**

`AH00526: Syntax error on line 7 of /etc/apache2/sites-enabled/25-site.mydomain.com_non-ssl.conf‌​: ServerName takes one argument, The hostname and port of the server`

This might have thrown  with same vhostname ,

I edited the manifest and this worked for me,

```
apache::vhost { 'site.mydomain.com:80':
  servername      => 'site.mydomain.com',
  port            => '80',
  docroot         => '/var/www/site',
  rewrite_rule    => '(.*) https://site.mydomain.com [R,L]',
  redirect_status => 'permanent',
  redirect_dest   => 'https://site.mydomain.com'
}

apache::vhost { 'site.mydomain.com:443':
  servername    => 'site.mydomain.com',
  port          => '443',
  ssl           => true,
  docroot       => '/var/www/site',
  docroot_owner => 'www-data',
  docroot_group => 'www-data',
  rewrite_rule  => '(.*) https://site.mydomain.com [R,L]',
}
```

Am i missing something?


wilson208:
@vineeth777 This is the wrong place for raising issues, all issues should be raised through [here](https://tickets.puppetlabs.com).

However, I do not believe the error you have hit is an issue with the module, or apache. When you do not specify the `servername` of an `apache::vhost` it defaults to the name of the vhost you have defined, so in this case servername has defaulted to `site.mydomain.com non-ssl`.

Inside `25-site.mydomain.com_non-ssl.conf‌​` this will look like `ServerName "site.mydomain.com non-ssl"` and apache see's this as invalid, because servername should just be set to the domain/IP address and optionally a port.

By setting the servername parameter is the correct way to solve this issue. Any further issues please raise them through the above link.

Many Thanks.

PR #1599, opened 2017-03-03
The base tag also needs link rewriting

tobixen:
The base tag is defined at https://www.w3schools.com/tags/tag_base.asp.  Including this ProxyHTMLLinks line is recommended i.e. at http://httpd.apache.org/docs/current/mod/mod_proxy_html.html#proxyhtmllinks and other places.

In my situation we're serving a http-only backend from apache over https, some oddities remained until we managed to fix the base url (i.e. weirdly enough google analytics javascript was fetched over http, causing a security alert).

I fail to see situations where one would want to rewrite the given URL for all other links than the base tag.


wilson208:
This LGTM, 👍 

Thanks!

PR #1601, opened 2017-03-07
Modules-4500 Add optional "AdvertiseFrequency" directive in cluster.conf template

EmersonPrado:
If calling class defines `advertise_frequency`, and `server_advertise` is `true`, this change adds the directive `AdvertiseFrequency` in mod_cluster virtual host conf file template

### DIFF
--- a/README.md
+++ b/README.md
@@ -1606,6 +1606,7 @@ class { '::apache::mod::cluster':
 
 - `port`: mod_cluster listen port. Default: '6666'.
 - `server_advertise`: Whether the server should advertise. Default: true.
+- `advertise_frequency`: Interval between advertise messages in seconds[.miliseconds]. Default: 10.
 - `manager_allowed_network`: Network allowed to access the mod_cluster_manager. Default: '127.0.0.1'.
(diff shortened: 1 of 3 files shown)


tphoney:
Looks like a good change, it might be worth documenting this new parameter in the readme.

EmersonPrado:
@tphoney Good idea. Done.

tphoney:
in the docs you say the default value is 10, but it in code it is undef. its late here, so i might be code blind

wilson208:
@tphoney have looked at the docs [here](https://docs.jboss.org/mod_cluster/1.2.0/html_single/#d0e701) and AdvertiseFrequency defaults to 10ms when not defined. This PR only sets the AdvertiseFrequency in the template when the param is defined, therefore it defaults to 10ms when not defined.

This LGTM 👍

EmersonPrado:
@wilson208 and @tphoney : would it be overkill to push a new commit improving the README to explain the default is from Apache modcluster, not from the Puppet class?
(Just for pickiness: 10s, not 10ms)

PR #1602, opened 2017-03-09
Add param for AllowOverride in the userdir.conf template

dstepe:
#4516 Add param for AllowOverride values in userdir template

Add a param to the apache::mod::userdir class to accept an array of override value. Replace the hardcoded list in the userdir.conf.erb template with an array join, similar to the existing Options line. Use the previously hardcoded values as the default for the overrides param.

The AllowOverride directive values were previously hardcoded in the template.

https://tickets.puppetlabs.com/browse/MODULES-4516


wilson208:
@dstepe LGTM 👍

PR #1608, opened 2017-04-06
Added supplementary_groups to the user resource

chgarling:
As discussed in https://tickets.puppetlabs.com/browse/MODULES-4646


tphoney:
This looks good, it would be awesome if this new parameter was documented in the README

chgarling:
I added the section about supplementary_groups to the README. How long would it take for the changes to come upstream?

tphoney:
Awesome work !!!!
Many thanks for the feature and the readme.

PR #1612, opened 2017-04-26
Enable configuring CA file in ssl.conf

JAORMX:
Instead of having to specify a CA file for each vhost, it's possible to
just specify it globally.


JAORMX:
@hunner @EmilienM could you check this out?

EmilienM:
:+1: 
@JAORMX fyi I don't have perms to land it.

PR #1614, opened 2017-05-02
Add WSGIRestrictEmbedded to apache::mod::wsgi

dsavineau:
Because this option is not available on all platform, it's not set in
the wsgi mod configuration by default.


eputnam:
@dsavineau thanks for this!

PR #1615, opened 2017-05-03
Account for Suse osfamily for proxy_html module

cetanu:
Self explanatory


tphoney:
@cetanu thanks for the PR !!! :)

PR #1616, opened 2017-05-03
MODULES-4816 - new param for mod::security class

cedef:
- New param `$deploy_security_crs` set to true by default so the default
  behaviour is unchanged.
- new rspec example so that the parameter is spec-tested

### DIFF
--- a/manifests/mod/security.pp
+++ b/manifests/mod/security.pp
@@ -24,6 +24,7 @@ class apache::mod::security (
   $secrequestbodylimit         = '13107200',
   $secrequestbodynofileslimit  = '131072',
   $secrequestbodyinmemorylimit = '131072',
+  $manage_security_crs         = true,
 ) inherits ::apache::params {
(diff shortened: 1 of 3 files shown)


manifests/mod/security.pp
```diff
@@ -24,6 +24,7 @@
   $secrequestbodylimit         = '13107200',
   $secrequestbodynofileslimit  = '131072',
   $secrequestbodyinmemorylimit = '131072',
+  $deploy_security_crs         = true,
```
hunner: Could you please name this `manage_security_crs` to be more similar to other parameters in the module?

cedef:
Here you are :) I've changed the parameter name and added an entry into the README.md file.
Thank you for your feedback !
Regards,

cedef:
Any news on this one ?

eputnam:
@hunner please review changes

PR #1620, opened 2017-05-11
Add attributes owner and group to apache::custom_config

aklein-dex:
These attributes are used for the resource type `file`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,10 @@ matrix:
     script: bundle exec rake beaker
     services: docker
     sudo: required
-  - rvm: 2.3.1
+  - rvm: 2.4.0
     bundler_args: --without system_tests
-    env: PUPPET_GEM_VERSION="~> 4.0"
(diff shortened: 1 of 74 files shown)


eputnam:
@aklein-dex , thanks for the PR! could you please rebase? also, what motivated you to add these params?

aklein-dex:
if the file is owned by user bob UID 1005 on the Puppet server, then it will be owned by the user with UID 1005 on the node where the file is deployed. But on that node maybe UID 1005 is Max. That can lead to a security risk.
So it would be nice to be able to specify the owner (and group) of the file.

ekohl:
@aklein-dex it looks like you merged in master but it'd be better to rebase. Now there are a lot of commits which makes reviewing very hard. Other than that to me it makes sense to set the owner/group explicitly.

aklein-dex:
@ekohl yes sorry for that. Let me close it and create another one.

PR #1622, opened 2017-05-11
Support SSL certificate file containing the key file.

johnzimm:
I have a situation where it is desirable to keep the certificate and key in the same file.  Modifying the templates/vhost/_ssl.erb template to exclude the SSLCertificateKeyFile if the 'ssl_key' parameter is set to 'false' allows this to happen.  The README has been updated to reflect this as well.  The default values of the parameter have not been modified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,8 @@ apache::vhost { 'cert.example.com':
 }
 ```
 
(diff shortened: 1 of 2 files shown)


johnzimm:
Withdrawing this pull request.  I've found several incompatibilities with things other than browsers when the certificate is configured this way.  It's now desirable to change the situation instead.

PR #1623, opened 2017-05-12
Fix single quoted string

lordbink:
single_quote_string_with_variables:ERROR:single quoted string containing a variable found


tphoney:
many thanks !

PR #1625, opened 2017-05-22
(#MODULES-4805) Delete enclose_ipv6.rb

rswarts:
* Delete enclose_ipv6 function that is also present in (required) stdlib
* Delete test for enclose_ipv6
* Bump stdlib requirement to the tag in which enclose_ipv6 was introduced


rswarts:
oh, I made a dupe...

#1624

PR #1630, opened 2017-06-07
[Modules 4746] Creates class for managing Apache mod_jk connector

EmersonPrado:
Current state: includes `mod_jk` module, creates jk.load, jk.conf and worker properties files and adds some description in the README file.

Lacks (as far as my creativity goes): parameter validation, proper parsing of some parameters, worker map file management and tests.

### DIFF
--- a/manifests/mod/jk.pp
+++ b/manifests/mod/jk.pp
@@ -11,7 +11,8 @@
 class apache::mod::jk (
   $workers_file          = undef,
   $worker_property       = {},
-  $shm_file              = "${::apache::logroot}/jk-runtime-status",
+  $logroot               = undef,
+  $shm_file              = 'jk-runtime-status',
   $shm_size              = undef,
   $mount_file            = undef,
   $mount_file_reload     = undef,
@@ -21,7 +22,7 @@ class apache::mod::jk (
   $mount_copy            = undef,
   $worker_indicator      = undef,
   $watchdog_interval     = undef,
-  $log_file              = "${::apache::logroot}/mod_jk.log",
+  $log_file              = 'mod_jk.log',
   $log_level             = undef,
   $log_stamp_format      = undef,
   $request_log_format    = undef,
@@ -68,6 +69,16 @@ class apache::mod::jk (
     notify  => Class['apache::service'],
   }
 
+  # Shared memory and log paths
+  if $logroot == undef {
+    $shm_path = ${::apache::logroot}/${shm_file}
+    $log_path = ${::apache::logroot}/${log_file}
+  }
+  else {
+    $shm_path = ${logroot}/${shm_file}
+    $log_path = ${logroot}/${log_file}
+  }
+
   # Main config file
   file {'jk.conf':
     path    => "${::apache::mod_dir}/jk.conf",
@@ -78,6 +89,7 @@ class apache::mod::jk (
     ],
   }
 
+  }
   # Workers file
   if $workers_file != undef {
     $workers_path = $workers_file ? {
--- a/manifests/mod/jk.pp
+++ b/manifests/mod/jk.pp
@@ -10,13 +10,13 @@
 #
 class apache::mod::jk (
   $workers_file          = undef,
-  $worker_property       = undef,
+  $worker_property       = {},
   $shm_file              = undef,
   $shm_size              = undef,
   $mount_file            = undef,
   $mount_file_reload     = undef,
-  $mount                 = undef,
-  $un_mount              = undef,
+  $mount                 = {},
+  $un_mount              = {},
   $auto_alias            = undef,
   $mount_copy            = undef,
   $worker_indicator      = undef,
@@ -42,8 +42,8 @@ class apache::mod::jk (
   $remote_port_indicator = undef,
   $remote_user_indicator = undef,
   $auth_type_indicator   = undef,
-  $options               = undef,
-  $env_var               = undef,
+  $options               = [],
+  $env_var               = {},
   $strip_session         = undef,
 ){
 
--- a/manifests/mod/jk.pp
+++ b/manifests/mod/jk.pp
@@ -46,6 +46,7 @@ class apache::mod::jk (
   $env_var               = {},
   $strip_session         = undef,
   # Workers file content
+  # See comments in template mod/jk/workers.properties.erb
   $workers_file_content  = [],
 ){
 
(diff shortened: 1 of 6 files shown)


EmersonPrado:
Now the class manages workers map file. Parsing of some parameters of worker properties file was corrected.

EmersonPrado:
Now the main mod_jk config file (jk.conf) accepts "Location" sections, defined by a hash parameter. Also got some minor corrections, like parameter defaults and linting. Needs more local testing and, perhaps, some spec tests to make it real good.

eputnam:
@EmersonPrado is there anything specifically we can help with on this PR?

EmersonPrado:
@eputnam Yes: with the tests. I have way too little knowledge on rspec-puppet, so I'm getting lost. I'd really appreciate some help making decent tests which pass.

manifests/mod/jk.pp
```diff
@@ -0,0 +1,105 @@
...
+  # Location list
+  # See comments in template mod/jk.conf.erb
+  $location_list         = [],
+  # Workers file content
+  # See comments in template mod/jk/workers.properties.erb
+  $workers_file_content  = {},
+  # Mount file content
+  # See comments in template mod/jk/uriworkermap.properties.erb
+  $mount_file_content    = {},
+){
```
eputnam: there's not really a great pattern to open up access to the vars you reference in the parameters here, so i'll take a page from many of the other mods and say use inheritance here. 
```puppet
) inherits apache {
```

then you can remove the `include ::apache` below as well.

spec/classes/mod/jk_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it { is_expected.to compile.with_all_deps }
+    it { is_expected.to create_class('apache::mod::jk') }
+    it { is_expected.to contain_class('apache') }
+    it { is_expected.to contain_apache__mod('jk') }
+    it { is_expected.to contain_file('jk.conf').that_notifies('Class[apache::service]') }
+  end
+
+  context "with only required facts and no parameters" do
+
+    let :facts do
```
eputnam: parens here for :facts

manifests/mod/jk.pp
EmersonPrado: Does that mean that, by using inheritance in the class, the external variables (from ::apache) will be available in the spec test?

eputnam: it's going to work just like the `include` but with a guaranteed parse order. i'm not actually sure what you're looking for, did you have something specific in mind?

EmersonPrado: What I exactly need is the mod/jk spec test to access the variables from ::apache class. If I understood you correctly, inheritance would make sure the variables get included before needed in the spec test. Is that correct?

I'm avoiding inheritance here because it's considered an anti-pattern by Puppet itself - see block "Aside: When to Inherit" [here](https://docs.puppet.com/puppet/5.0/lang_classes.html#inheritance).

An alternative would be to declare the parameter as undef in the class signature, then assingn values depending on such parameter, like I did with workers_file and mount_file. So, by assigning something in the test in :parameters section, maybe the required variables will be there. Doesn't seem much clear (I'm in a hurry now), but feel free to ask.

Thanks!

manifests/mod/jk.pp
```diff
@@ -0,0 +1,105 @@
...
+# All parameters are optional. When undefined, some receive default values,
+# while others cause an optional directive to be absent
+#
+# For help on parameters, pls see official reference at:
+# https://tomcat.apache.org/connectors-doc/reference/apache.html
+#
+class apache::mod::jk (
+  $workers_file          = undef,
+  $worker_property       = {},
+  $shm_file              = "${::apache::logroot}/jk-runtime-status",
```
eputnam: in the case of these logroot dependencies, you could just set these to the file's name and then prepend the logroot path below. this would be a more predictable pattern for those who are wanting to override anyway.

manifests/mod/jk.pp
eputnam: i commented above, i think the undef pattern you mention would work best, then we can avoid inheritance altogether. as for mocking apache params, you can use "pre_conditions" in a compile block for one: https://github.com/puppetlabs/puppetlabs-apache/blob/40c1ad4793f00944d1acac36f32ff9be580b157d/spec/classes/mod/auth_kerb_spec.rb#L77-L105 is that maybe like what you're thinking off?

EmersonPrado: OK. I used the undef pattern for ::apache::logroot only, including a new parameter for it in mod::jk class. Then I changed the file parameters to only the name, as you suggested (so the user can override either filename or diretory at will). Finally, added the logroot parameter in the :params section in the spec test. Let's hope it works!

As for pre_conditions section, I still don't get quite well how it works. I'll try something later. But thanks for the hint!

eputnam:
@EmersonPrado,

for logroot, you don't need a $logroot parameter in your class, you should just be able to get it from ::apache, and it will never be `undef` so you don't need that check either. unless of course you're expecting the user to set a special logroot for the jk mod. does that make sense?

so:
1. get rid of the $logroot parameter (don't forget to remove it from the spec test)
2. you don't need to handle `undef` values for $apache::logroot, just set `shm_file` and `log_file`, like you are without the conditional:
```puppet
$shm_path = "${::apache::logroot}/${shm_file}"
$log_path = "${::apache::logroot}/${log_file}"
```

and for future reference, :params in the spec test expects a hash, so it might look like this instead:
```ruby
let (:params) do
      { :logroot => '/var/log/httpd' }
end
```

was any of that helpful?

EmersonPrado:
@eputnam Yes, I'm including the possibility for the user to override logroot directory. I don't know how much this will be used - the user would probably be safer overriding this in ::apache class instead - but I'm leaving this here for pickiness. Maybe it's something to think twice.
As for the `undef` test, unfortunately the spec test fails without that apparently unnecessary pattern. If I just have `$logroot = $::apache::logroot` in apache::mod::jk class signature, the spec test will complain there's no `$::apache::logroot`. This is kinda OK, since the spec test doesn't parse included classes, but it should ignore the external variable, since it's overridden with `let (:params) do { :logroot => '/var/log/apache2' } end`. But this is a problem somewhere else.

Your correction about spec :params section nailed it.

With last commits, all spec tests pass. I included a Debian context too. So I guess we're all set now.

Thanks a lot!

eputnam:
@EmersonPrado thanks this is great.

PR #1631, opened 2017-06-08
MODULES-3838 Pass mod_packages through init.pp to allow overrides

optiz0r:
When using the SCL packages on CentOS, package names have a prefix that needs to be overridden. With most other attributes in params.pp, this is possible by overriding variables passed to `Class['::apache']`, but in the case of `mod_packages` it's read directly from the params file on use. This is workaroundable by not using the `apache::mod::...` subclasses can calling `apache::mod { 'mod_name' }` directly, but there is a compounding problem.
The `apache::vhost::auth_kerb` parameter is used for for both including the `auth_kerb` modules (using class syntax rather than include, so cannot be duplicated), and for including config snippets in the generated vhost. It's not possible to make use of the config snippet whilst also needing to amend the package name.

This commit follows the existing style and passes `mod_packages` through `init.pp`, and updates all references of `apache::params::mod_packages` to `apache::mod_packages`. It's therefore possible to override the package names using something like:

```puppet
include ::apache::params
class { '::apache':
  mod_packages => merge($::apache::mod_params {
    'auth_kerb' => 'httpd24-mod_auth_kerb',
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1183,10 +1183,11 @@ Default: Depends on operating system.
 Allows the user to override default module package names.
 
 ```puppet
-class { '::apache':
-  mod_packages => merge($::apache::mod_params {
+include apache::params
+class { 'apache':
+  mod_packages => merge($::apache::params::mod_packages, {
     'auth_kerb' => 'httpd24-mod_auth_kerb',
-  }
+  })
 }
 ```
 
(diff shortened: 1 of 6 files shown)


optiz0r:
Rebase done. I don't have a lot of free cycles, and considering there is no documentation or tests for the `apache::params::mod_packages` in the first place, there's nothing to update, and the codebase is no worse off by any measure with this commit as-is than it was before.

PR #1633, opened 2017-06-13
When absolute path is specified for access_log_file/error_log_file, don't prepend logbase

ca-asm:
It might sometimes be desirable to have access and error logs in different folders. This isn't possible at the moment as the `access_log_file`/`error_log_file` is treated as a path relative to `logbase`. This pull request changes the behaviour so that if absolute path is detected, `logbase` is not prepended.


tphoney:
Excellent change, thanks for this @ca-asm

PR #1637, opened 2017-06-23
Change name of event conf to ensure values load.

lliss:
The apache event module configuration gets added into a file. This
file however contains duplicate configuration options and the original
values are stored in `mpm_event.conf`. Since `mpm_event.conf` sorts
alphabetically after `event.conf` the original settings in
`mpm_event.conf` will always be used. This change alters the name of
the file used to store the configuration so that it sorts after
`mpm_event.conf` and thus the new settings will take precedence.

### DIFF
--- a/manifests/mod/event.pp
+++ b/manifests/mod/event.pp
@@ -54,7 +54,7 @@ class apache::mod::event (
   case $::osfamily {
     'redhat': {
       if versioncmp($_apache_version, '2.4') >= 0 {
-        apache::mpm{ 'event':
+        apache::mpm{ 'override_event':
           apache_version => $_apache_version,
         }
       }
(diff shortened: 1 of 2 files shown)


lliss:
I rebased this against master so this is up to date but I don't have a good sense of how to get around the failing tests.

david22swan:
@lliss Have you made any progress regarding the failing tests?

david22swan:
@lliss Apologies but due to a lack movement on this PR I will be closing it. If you have the time later on to return to and finish your work we would be more than happy to review it once again.

lliss:
Feel free to close it. I don't use this module anymore.  However you should probably leave this open as an issue somewhere because it is a legitimate problem in the module and should be fixed.

PR #1639, opened 2017-07-03
Ensure that ProxyPreserveHost is set even when ProxyPass (etc) are not.

tpdownes:
Proxying can be enabled when the [P] flag is given to a RewriteRule, not only when ProxyPass is used. In that instance, we may want to be able to set ProxyPreserveHost.


eputnam:
@tpdownes thanks for the PR! could you please rebase on master so it will run the newer travis config?

tpdownes:
I believe that 56e5800 should do it.

PR #1643, opened 2017-07-06
Don't install proxy_html package in ubuntu xenial

amateo:
In Ubuntu Xenial, proxy_html module is include in apache standard
packages, so don't try to install it


eputnam:
@amateo thanks for your work here! Do you know if the same is true for Debian 9? If not, please adjust the conditional accordingly.

mmoll:
It's also included in apache2-bin on Debian 9:
```
$ cat /etc/debian_version
9.0
$ dpkg -S /usr/lib/apache2/modules/mod_proxy_html.so
apache2-bin: /usr/lib/apache2/modules/mod_proxy_html.so
```

eputnam:
ugh, i read the pr wrong. thanks @amateo @mmoll !

PR #1647, opened 2017-07-12
(MODULES-4445) remove deprecated parameter in authnz_ldap

eputnam:
the $verifyServerCert parameter in the authnz_ldap module class is
deprecated and this removes it and updates the template for the upcoming major release.


hunner:
We're doing a major release, so just delete deprecated parameters.

PR #1648, opened 2017-07-12
(MODULES-2808) documented mod_nss

8675309:
added entry in README for mod_nss module


hunner:
Could you also update the doc line on line 1423?

PR #1649, opened 2017-07-13
Custom default of parameters

imluyuan:
We use Haproxy and Apache on the same Server that Can not use the custom default port.


hunner:
I think there is already a parameter provider for turning off the default vhost if it does not work for you: https://github.com/puppetlabs/puppetlabs-apache/pull/1649/files#diff-60ae41fd0a31977447947f59940ee9a4L19

PR #1621, opened 2017-05-11
Migrate to puppet4 datatypes

bastelfreak:
goal of this is to get rid of all the legacy valida_* function calls, and not to provide perfect datatypes for all parameters.

PR #1653, opened 2017-07-24
Use puppet 4 types to avoid deprecation warnings

ekohl:
This completes the work https://github.com/puppetlabs/puppetlabs-apache/pull/1621 started. I think this covers all deprecation warnings.

Also makes puppet-lint a bit happier


eputnam:
thanks @ekohl !

PR #1661, opened 2017-07-28
(MODULES-5324) Adding documentation for overrides

pmcmaw:
Adding documentation for the overrides param for apache::mod::userdir.
Also removing ' to solve a formatting issue.


hunner:
([MODULES-5324](https://tickets.puppet.com/browse/MODULES-5324)) Adding documentation for overrides

this is testing for changelog generation.

PR #1662, opened 2017-07-28
use ERB comment to keep it out of the generated file

mmoll:


templates/httpd.conf.erb
```diff
@@ -11,7 +11,7 @@ KeepAlive <%= @keepalive %>
 MaxKeepAliveRequests <%= @max_keepalive_requests %>
 KeepAliveTimeout <%= @keepalive_timeout %>
 LimitRequestFieldSize <%= @limitreqfieldsize %>
-# Actually >= 2.4.24, but the minor version is not provided
+<%# Actually >= 2.4.24, but the minor version is not provided -%>
```
tphoney: Is the final '-' needed ?

mmoll: nope, just personal style. does it need to get removed?

tphoney:
Nope, its all good. i was just wondering if there was some ERB magic happening.

tphoney:
THANKS @mmoll  !!!

PR #1663, opened 2017-08-01
Expose loadfile_name option to mod::python class

traylenator:
This is a work around to a particular clash between these
two lines.

```puppet
class{'apache::mod::python':}
class{'collectd::plugin::python':}
```

Both classes create a file resource `file{'python.load':}`

This patch allows the resource name to be configured for
`apache::mod::python` setup.

There is an altenative more perfect solution to use a better
more unique name space by default e.g `file{"apache_${_loadfile_name}:` but this
would touch so many test files....

This change is fully backwards compatible.

* [collectd::plugin::python](https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/plugin.pp#L16)


tphoney:
This seems like a reasonable workaround for the issue. Should this be documented ?

traylenator:
Docs and rebased.

PR #1664, opened 2017-08-06
use Enum instead of validate_re for $sendfile

mmoll:


ekohl:
I don't see why the tests failed.

mmoll:
parallel_tests maybe? 😹

spec/classes/apache_spec.rb
```diff
@@ -636,7 +636,7 @@
         it "should fail" do
           expect do
             catalogue
-          end.to raise_error(Puppet::Error, /"foo" does not match/)
+          end.to raise_error(Puppet::PreformattedError, /Evaluation Error: Error while evaluating a Resource Statement, Class\[Apache\]: parameter 'sendfile' expects a match for Enum\['Off', 'On', 'off', 'on'\], got 'foo'/)
```
bastelfreak: is it still needed to test against those cases? In the Vox Pupuli namespace we dropped them

mmoll: I can delete this for sure, if wanted...

manifests/init.pp
```diff
@@ -38,7 +38,7 @@
   $purge_vhost_dir                                               = undef,
   $purge_vdir                                                    = false,
   $serveradmin                                                   = 'root@localhost',
-  $sendfile                                                      = 'On',
+  Enum['On', 'Off', 'on', 'off'] $sendfile                       = 'On',
```
TheMeier: why not map On/Off, on/off to bool in puppet?

hunner: Eh, overly complicated?

ekohl: For simplification you could add a type ```Apache::OnOff``` since this validation is used in a lot of places.

ekohl:
@mmoll if you rebase, do the tests pass then?

mmoll:
Tests are green now.

manifests/init.pp
mmoll: I can do that, if wanted.

However, that would be a separate PR, which also changes all occurrences of it.

ekohl: I think you're right they shouldn't be in this PR.

mmoll:
anything needed to get this in?

PR #1667, opened 2017-08-08
[Modules 5385] Include support for Apache 2.4 mod_authz_host directives in apache::mod::status

EmersonPrado:
In the current state, creates helper templates `mod/_allow` and `mod/_require` with, respectively, Apache 2.2 and 2.4 host authorization logic. These aim to be used by various templates in `mod/`. Then, template `mod/status.conf.erb` is modified to use one of these helper templates depending on Apache version.
The `_require` helper template implements "Require" directives from the parameter `requires` provided by the calling class. If this is `undef`, then uses variable `requires_defaults`, also provided by the calling class. In order not to break existing systems, if `requires == undef` but `allow_from` is defined, uses the latter and issue a deprecation warning.

`requires` parameter can be either of:
- `undef`
- '' or 'unmanaged' (No "Require" directive)
- String (`ip <list>`, `host <list>`, `not <something>`, etc.)
- Array of strings (each with same syntax as standalone string)
- Hash
  - Key `requires` - Array of strings (same rule as standalone array)
  - Key `enforce` (optional) - Either of "Any|All|None" for `<Require*></Require*>` block

Tested on RHEL 6 and 7 and Debian 7 and 8, but only in Puppet 3 (unfortunately, it's what I have here for the moment)
Test specs were not updated yet

### DIFF
--- a/manifests/mod/status.pp
+++ b/manifests/mod/status.pp
@@ -7,6 +7,20 @@
 # - $allow_from is an array of hosts, ip addresses, partial network numbers
 #   or networks in CIDR notation specifying what hosts can view the special
 #   /server-status URL.  Defaults to ['127.0.0.1', '::1'].
+#   > Creates Apache < 2.4 directive "Allow from"
+# - $requires is either a:
+#   - String with:
+#     - '' or 'unmanaged' - Host auth control done elsewhere
+#     - 'ip <List of IPs>' - Allowed IPs/ranges
+#     - 'host <List of names>' - Allowed names/domains
+#     - 'all [granted|denied]'
+#   - Array of strings with ip or host as above
+#   - Hash with following keys:
+#     - 'requires' - Value => Array as above
+#     - 'enforce' - Value => String 'Any', 'All' or 'None'
+#       This encloses "Require" directives in "<Require(Any|All|None)>" block
+#       Optional - If unspecified, "Require" directives follow current flow
+#   > Creates Apache >= 2.4 directives "Require"
 # - $extended_status track and display extended status information. Valid
 #   values are 'On' or 'Off'.  Defaults to 'On'.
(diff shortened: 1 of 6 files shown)


EmersonPrado:
Now with spec tests for all reasonable options for `requires` parameter:
- undef
- Empty string
- 'unmanaged'
- String with 'ip ...' or 'host ...'
- Array of strings
- Hash with 'requires' array
  - Without 'enforce' key
  - With 'enforce' key

After lots of corrections in spec test (since I'm a beginner in that still) and a minor correction in helper template, all tests pass.

I'd like to read other people opinions about the path I'm going. In the absence of issues with that, I intend to replicate this in the other affected classes.

manifests/mod/status.pp
```diff
@@ -26,7 +26,8 @@
 #  }
 #
 class apache::mod::status (
-  Array $allow_from                               = ['127.0.0.1','::1'],
+  Optional[Array] $allow_from                     = undef,
+  $requires                                       = undef,
```
ekohl: Could you check this with ```Optional[Variant[String, Array, Hash]]```?

EmersonPrado:
@ekohl Good idea about validation. Done!

EmersonPrado:
Since this PR has already many commits and changes, I'll limit it only to `apache::mod::status` class, then work the other affected classes in new branches and PRs. If there are no other suggestions or other needs about this, I'll just wait for the merge to copy the implementation to the other classes.

david22swan:
@EmersonPrado Could you please alter the readme to reflect your changes?
Also we would prefer you to squish your commits down into a single one.

EmersonPrado:
@david22swan Done. Also rebased to master. Got some failures, but I think they're unrelated.

david22swan:
Yeah, those fails are related to a separate issue.

PR #1669, opened 2017-08-15
Use Puppet 4 Pattern instead of validate_re

ekohl:


ekohl:
I guess https://github.com/puppetlabs/puppetlabs-apache/pull/1664 is a better solution.

PR #1671, opened 2017-08-16
[Modules 5492] - Include treatment for absolute, relative and pipe paths for JkLogFile and JkShmFile for class mod::jk

EmersonPrado:
- Prepends logdir in relative paths
- Uses unchanged absolute / pipe paths
- Uses defaults for unspecified paths

Previous version prepended logdir in any filename, resulting in incorrect paths for absolute and pipe paths.


eputnam:
@EmersonPrado, good to see you working on apache again! Would you mind adding some unit tests for this?

EmersonPrado:
@eputnam Sure! We're using this module extensively at work, so we're really willing to improve as much as we can.
Here are the docs. I'll work on the spec tests now.
Thanks!

EmersonPrado:
OK, I have to admit I know squat about spec tests. I keep hoping its documentation improves to the point I can really solve this kind of stuff. Gave up for today.

spec/classes/mod/jk_spec.rb
```diff
@@ -115,7 +115,7 @@
 
         it_behaves_like 'minimal resources'
         it do
-          is_expected.to contain_file("/etc/httpd/conf.d/jk.conf")
+          is_expected.to contain_file("#{:mod_dir}/jk.conf")
```
eputnam: this is a bit of a gotcha. `contain_file` is referencing the file resource in the catalog by its title. If you look in [jk.pp](https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/jk.pp#L83) you'll see that the config file has a hard-coded title of 'jk.conf' which is why this test wasn't working. I think what you're looking for is something more like this:
```ruby
is_expected.to contain_file('jk.conf').with({ :path => "#{mod_dir}/jk.conf" })
```

EmersonPrado: Ah, that should be it! I was using the path, when I should be using the resource title. I'll start over pushing very little at a time, so we can get somewhere. Thanks for the hint!

EmersonPrado:
@eputnam Now we've got updated docs and functional tests covering all situations envolved. Anything else comes to mind?
Regards,
Emerson

spec/classes/mod/jk_spec.rb
```diff
@@ -55,11 +59,74 @@
...
+    let (:mod_dir) { mod_dir }
+
+    it_behaves_like 'minimal resources', mod_dir
     it {
       verify_contents(catalogue, 'jk.conf', ['<IfModule jk_module>', '</IfModule>'])
     }
 
   end
 
+  shm_log_paths = {
```
eputnam: this should/can be a let statement as well:
```ruby
let (:shm_log_paths) do
  { ... }
end
```

EmersonPrado: rspec-puppet didn't like the change:
https://travis-ci.org/puppetlabs/puppetlabs-apache/jobs/268018580#L799
https://travis-ci.org/puppetlabs/puppetlabs-apache/jobs/268018581#L680
'`shm_log_paths` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).'

What is the correct way to reference local variables declared in `let` statements? Current form is [this](https://github.com/puppetlabs/puppetlabs-apache/pull/1671/files#diff-b5f561614a193ce787d91557fc9ed417R98)

eputnam:
oh i see what happened. i'm sorry, i guess i was wrong. what you can do is call `each` right on the hash instead of assigning it to a variable if you're only using it the one time.

PR #1672, opened 2017-08-17
(maint) Improve version matching fact

smortex:
* `/` not being a special char, there is no need to escape it (removed preceding `\`);
* `.` matching *any* char, escaping it is required so that it matches a dot char (added preceding `\`).


pmcmaw:
@smortex thank you for your contribution 👍

PR #1676, opened 2017-08-17
(maint) Account for Array[Hash] from ecdee22 in spec tests

hunner:
This previously was an array of arrays, but the recently merged PR doesn't allow that.

### DIFF
--- a/spec/acceptance/vhost_spec.rb
+++ b/spec/acceptance/vhost_spec.rb
@@ -503,8 +503,7 @@ describe 'apache::vhost define' do
           $_directories = [
             { 'path' => '/var/www/files', },
             { 'path' => '/foo/', 'provider' => 'location', 'directoryindex' => 'notindex.html', },
-            $_files_match_directory,
-          ]
+          ] + $_files_match_directory
 
           apache::vhost { 'files.example.net':
             docroot     => '/var/www/files',


ekohl:
It looks like the acceptance tests on Travis now fail.

eputnam:
is there any reason we couldn't just change $directories to include arrays in arrays? besides the gratuitous brackets.
```puppet
Optional[Variant[Hash, Array[Variant[Array,Hash]]]] $directories
```

ekohl:
> It looks like the acceptance tests on Travis now fail.

https://github.com/puppetlabs/puppetlabs-apache/blob/34e45869164b65ce37de20b83609e9afd6a7c84a/spec/acceptance/vhost_spec.rb#L495 needs to become an array as well for the tests to pass.

> is there any reason we couldn't just change $directories to include arrays in arrays? besides the gratuitous brackets.

Personally I'm not a big fan of the flatten pattern but since it's in the code users might rely on it and it could break the API for them. It means the current acceptance test is a good test!

eputnam:
PR testing systems test with Apache 2.4, the problem we're seeing is with 2.2 on redhat 6 where we need to be able to pass arrays of rules to `allow` and `deny` for each directory.

eputnam:
closing in favor of #1681

PR #1678, opened 2017-08-18
(MODULES-5458) Add docs for default vhost workaround

HAIL9000:
Since there has repeatedly been some confusion with respect to
overriding the `default` and `default_ssl` vhosts in apache, add
some documentation explaining what is and isn't possible and a
potential workaround. See MODULES-5193 for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3029,6 +3029,15 @@ The `apache::vhost` defined type uses `concat::fragment` to build the configurat
 
 For the custom fragment's `order` parameter, the `apache::vhost` defined type uses multiples of 10, so any `order` that isn't a multiple of 10 should work.
 
+> **Note:** When creating an `apache::vhost`, it cannot be named `default` or `default-ssl`, because vhosts with these titles are always managed by the module. This means that you cannot override `Apache::Vhost['default']`  or `Apache::Vhost['default-ssl]` resources. An optional workaround is to create a vhost named something else, such as `my default`, and ensure that the `default` and `default_ssl` vhosts are set to `false`:
+
+```
+class { 'apache':
+  default_vhost     => false
+  default_ssl_vhost => false,
+}
+```
+
 **Parameters**:
 
 ##### `access_log`


README.md
```diff
@@ -3029,6 +3029,15 @@ The `apache::vhost` defined type uses `concat::fragment` to build the configurat
 
 For the custom fragment's `order` parameter, the `apache::vhost` defined type uses multiples of 10, so any `order` that isn't a multiple of 10 should work.
 
+> **Note:** When creating an `apache::vhost` it cannot be named `default` or `default-ssl` as vhosts with these titles will always be managed by the module. This means that you cannot override the `default` or `default-ssl` vhosts. An optional workaround is to create a vhost named something else such as "my default" and ensure that the `default` and `default_ssl` vhosts are set to false:
```
jbondpdx: > **Note:** When creating an `apache::vhost`, it cannot be named `default` or `default-ssl`, because vhosts with these titles are always managed by the module. This means that you cannot override the `default` or `default-ssl` vhosts. An optional workaround is to create a vhost named something else, such as "my default", and ensure that the `default` and `default_ssl` vhosts are set to `false`:

HAIL9000:
@jbondpdx Updated! Thanks for the suggestions

README.md
```diff
@@ -3029,6 +3029,15 @@ The `apache::vhost` defined type uses `concat::fragment` to build the configurat
 
 For the custom fragment's `order` parameter, the `apache::vhost` defined type uses multiples of 10, so any `order` that isn't a multiple of 10 should work.
 
+> **Note:** When creating an `apache::vhost`, it cannot be named `default` or `default-ssl`, because vhosts with these titles are always managed by the module. This means that you cannot override the `default` or `default-ssl` vhosts. An optional workaround is to create a vhost named something else, such as `my default`, and ensure that the `default` and `default_ssl` vhosts are set to `false`:
```
hunner: I would probably take out "This means that you cannot override the `default` or `default-ssl` vhosts" as this is precisely what `default_vhost => false` is for. Maybe it should be "... `Apache::Vhost['default']` or `Apache::Vhost['default-ssl']` resources" to be more specific about what cannot be overridden?

HAIL9000:
@hunner Updated!

PR #1680, opened 2017-08-25
Setup SSL/TLS client auth without overly broad trusts for client certificates

epackorigan:
If `$ssl_certs_dir` defaults to a location that contains certificates, apache will trust implicitly all the certificates presented by the client for auth that were issues by *any* certificates in that location. Setting the default to the location where all root certificates is stored enabled anyone with a key and cert from any of the SSL certificate provider to connect.

Note: 
 * In Gentoo, `$ssl_certs_dir` was pointing to /etc/apache2/ssl, which may have been safe, or used for storing CA certs to check client auth. Gentoo users may need to set that value explicitly depending on how it was used in prior versions.

For more details, see [MODULE-5471]

### DIFF
--- a/spec/acceptance/apache_ssl_spec.rb
+++ b/spec/acceptance/apache_ssl_spec.rb
@@ -19,6 +19,7 @@ describe 'apache ssl' do
         }
       EOS
       apply_manifest(pp, :catch_failures => true)
+      apply_manifest(pp, :catch_changes => true)
     end
 
     describe file("#{$vhost_dir}/15-default-ssl.conf") do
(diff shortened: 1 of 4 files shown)


eputnam:
commented on [ticket](https://tickets.puppet.com/browse/MODULES-5471)

epackorigan:
I don't understand why the test is failing. Is it inheriting the defaults from the first test listed? How would i set it up to have a brand new setup to run into?

eputnam:
@epackorigan extra space happens to the best of us. I'd be happy to talk to you about running acceptance tests but maybe email or chat would be a better forum for that.

templates/vhost/_ssl.erb
```diff
@@ -19,23 +30,12 @@
   <%- if @ssl_crl -%>
   SSLCARevocationFile     "<%= @ssl_crl %>"
   <%- end -%>
+  <%- if @ssl_verify_depth -%>
+  SSLVerifyDepth          <%= @ssl_verify_depth %>
```
eputnam: what was the purpose of changing the order of these directives?

spec/acceptance/apache_ssl_spec.rb
```diff
@@ -91,4 +92,61 @@ class { 'apache':
...
+        apache::vhost { 'test_ssl_ca_only':
+          docroot              => '/tmp/test',
+          ssl                  => true,
+          ssl_cert             => '/tmp/ssl_cert',
+          ssl_key              => '/tmp/ssl_key',
+          ssl_ca               => '/tmp/ssl_ca',
+          ssl_verify_client    => 'test',
+        }
+      EOS
+      apply_manifest(pp, :catch_failures => true)
```
eputnam: in our acceptance tests, we use a pattern to ensure idempotency that applies the manifest twice; the second time ensures nothing has changed.
```puppet
apply_manifest(pp, :catch_failures => true)
apply_manifest(pp, :catch_changes => true)
```

ekohl: There's even a ```shared_examples "a idempotent resource"``` for this in the spec helper so you can simply use ```it_behaves_like "a idempotent resource"``` here instead.

templates/vhost/_ssl.erb
epackorigan: I was trying to regroup the directives that have no effect when client certs are not being looked at/needed... I just moved them all around into one block to make it easier to read (and missed that one the first time around).

spec/acceptance/apache_ssl_spec.rb
epackorigan: Added the idempotency tests to all 4 tests modified.

templates/vhost/_ssl.erb
epackorigan: it's not as clear as i'd like from the diff...
the set of directives that i moved in here are only effective is "ssl_verify_client" on line 19 is set. That "if" statement ends on line 39. That means that all the directives that are "ssl client auth specific" are only included in the config when they are actually needed/used.

This might be out of scope from the pull request perspective, but will make it clear that those are only to be used when the setup enforce verification of client certificates for authentication.

eputnam:
@epackorigan , please remove the lines you added to the change log. We definitely appreciate you including them but those will get taken care of when we release this patch and will go under different headings.

PR #1682, opened 2017-08-26
the userdir apache module is install by default and with this change is configure u…

c4m4:
…serdir.conf with <Directory /*/> and Multiviews is begin applied to all dirs

This reverts commit e37b99459b6376bcdbb5d231b77108b38f2e0805, reversing
changes made to 0c6999201600fe41fc2d4bbcf51123732786245c.


c4m4:
If I using apache module with this change, the apache MultiViews option is applied to all dirs, If I don't set any dir for the userdir module the module configuration that puppet generate is as follow:
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
cat /etc/apache2/mods-enabled/userdir.conf 
<IfModule mod_userdir.c>
  UserDir disabled root
  UserDir /home/*/public_html 

  <Directory "/*/"> # :(
    AllowOverride FileInfo AuthConfig Limit Indexes
    Options MultiViews Indexes SymLinksIfOwnerMatch IncludesNoExec
    <Limit GET POST OPTIONS>
      Order allow,deny
      Allow from all
    </Limit>
    <LimitExcept GET POST OPTIONS>
      Order allow,deny
      Allow from all
    </LimitExcept>
  </Directory>
</IfModule>
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

eputnam:
@c4m4 welcome to github! should Directory match the UserDir path? "/home/*/public_html" from the example above.

eputnam:
@c4m4 thanks for bringing this to our attention! the ability to pass a custom path is a feature many users have requested so I'm hesitant to revert it. Instead, i believe #1683 will make both parties happy so I'm going to close this. Please feel free to re-open if you feel strongly about this feature being reverted.

c4m4:
I solved removing userdir module from my setup, by I think the module need to generate UserDir /home/*/public_html and <Directory "home/*/public_html "> too, in this way, by default MultViews isn't apply to all apache dirs and the users can specify a costum userdir too.

PR #1687, opened 2017-09-05
Add enable capabilities to itk

edestecd:
Recently epel has updated the mod_itk package for RedHat:  https://bugzilla.redhat.com/show_bug.cgi?id=1432881

They have compiled with Linux capabilities support, which is enabled by default.
This breaks many existing setups that were using nfs etc that does not support Linux capabilities.

This adds the ability to turn off Linux capabilities and restore functionality on these setups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2146,6 +2146,54 @@ Installs and manages [`mod_info`][], which provides a comprehensive overview of 
 
   Default: `true`.
 
+##### Class: `apache::mod::itk`
+
+Installs and manages [`mod_itk`][], which is an (MPM) that is loaded and configured for the HTTPD process. [official documentation](http://mpm-itk.sesse.net/)
+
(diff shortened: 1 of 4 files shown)


edestecd:
Any comment on this?

david22swan:
@edestecd apologies for the wait. If you could add docs to match your changes I would be happy to merge.

edestecd:
oops. I did miss the docs. I'll update

edestecd:
@david22swan doc added!

edestecd:
hmm, not really sure whats up with the tests. Looks like a network issue

david22swan:
The error has appeared on other pull requests and is unrelated to yours specifically.

david22swan:
Will return to your PR as soon as it is solved.

PR #1693, opened 2017-09-08
Include mod_wsgi when using wsgi options

ekohl:
This copies the if statement around the wsgi concat fragment so mod_wsgi is loaded if any option is used.


tphoney:
#1702  includes some tests. thanks for the work @ekohl

PR #1701, opened 2017-09-27
MODULES-5649 Do not install mod_fastcgi on el7

tphoney:


manifests/params.pp
```diff
@@ -106,7 +106,11 @@
         default => 'mod_authz_ldap',
       },
       'authnz_pam'            => 'mod_authnz_pam',
-      'fastcgi'               => 'mod_fastcgi',
+      'fastcgi'               => $::apache::version::distrelease ? {
```
ekohl: Wouldn't this produce undefined results on other distros like Debian?

tphoney: If you look here https://github.com/tphoney/puppetlabs-apache/blob/c891950899a35a4af79fb23ce123f29d6da3639c/manifests/params.pp#L62-L110 It is for centos only @ekohl

ekohl: You're right, my bad. It's not an early morning but it does feel like it ...

manifests/mod/fastcgi.pp
```diff
@@ -1,6 +1,8 @@
 class apache::mod::fastcgi {
   include ::apache
-
+  if ($::osfamily == 'Redhat' and $::operatingsystemrelease > '6') {
```
ekohl: Should this use ```versioncmp```?

manifests/mod/fastcgi.pp
```diff
@@ -1,6 +1,8 @@
...
+    fail("mod_fastcgi is no longer supported on el7 and above.")
```
ekohl: I do wonder if other distros will deprecate it too and if there should be a more generalized check but I think this could be fine for now and refactored later.

manifests/params.pp
tphoney: its raining and miserable here, it isnt good. Thanks for looking over the PR. Much appreciated.

PR #1704, opened 2017-09-29
Added listenbacklog to prefork module

c4m4:
worker module has $listenbacklog          = '511',
event module has $listenbacklog          = '511',

I use prefork and I would like to use listenbacklog too.


tphoney:
Thanks !

PR #1714, opened 2017-10-22
Reduce Travis test parallelism

jearls:
Set the maximum parallel test processors to 16, to reduce the load on the Travis servers and prevent tests from being killed.

(This is needed in order for the tests for #1659 to not fail, but is not directly related to that PR, so I'm introducing it in an independent PR)


willmeek:
Thank you for your contribution @jearls

jearls:
These changes were overwritten by a `Module sync` commit.  How do I get these changes to be made permanent?

PR #1715, opened 2017-10-25
Add attributes owner and group to apache::custom_config

aklein-dex:
If the file is owned by user bob UID 1005 on the Puppet server, then it will be owned by the user with UID 1005 on the node where the file is deployed. But on that node maybe UID 1005 is Max. That can lead to a security risk.
So it would be nice to be able to specify the owner (and group) of the file in the class apache::custom_config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3042,6 +3042,20 @@ Boolean.
 
 Default: `true`.
 
+##### `owner`
+
+Specifies the owner of the configuration file. Default: 'root'.
+
+##### `group`
+
+Specifies the group of the configuration file. Default: Depends on your operating system.
+
+- **Debian**: `root`
+- **FreeBSD**: `wheel`
+- **Gentoo**: `wheel`
+- **Red Hat**: `root`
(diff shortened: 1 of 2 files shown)


david22swan:
@aklein-dex please add test's that reflect your changes

david22swan:
Closing and reopening PR to solve bundler issue

hunner:
I believe this is backwards-compatible as the file resource would have been just root/root before, though if any users are using file resource defaults then they would get caught by this.

hunner:
Oh you're right, for file resources the owner/group are read via the file_metadata indirector and would map to the perms on the master... does that make this change backwards-incompatible, or a bugfix/feature?

ekohl:
@hunner IMHO a bugfix

david22swan:
Apologies, but due to the amount of tie that this PR has been left to lie fallow I feel that I must know close it. I apologize if this inconveniences you and we at the Modules team welcome any and all future PRs you may make.
Best Wishes

PR #1716, opened 2017-10-26
added option to include CacheIgnoreHeaders for disk_cache module

gitkent:
https://httpd.apache.org/docs/2.2/mod/mod_cache.html#cacheignoreheaders
http://blog.futtta.be/2011/02/23/3-apache-mod_cache-gotchas/

### DIFF
--- a/templates/mod/disk_cache.conf.erb
+++ b/templates/mod/disk_cache.conf.erb
@@ -2,3 +2,6 @@ CacheEnable disk /
 CacheRoot "<%= @_cache_root %>"
 CacheDirLevels 2
 CacheDirLength 1
+<% if scope['cache_ignore_headers'] != :undef -%>
+CacheIgnoreHeaders <%= @cache_ignore_headers %>
+<% end -%>
--- a/templates/mod/disk_cache.conf.erb
+++ b/templates/mod/disk_cache.conf.erb
@@ -2,6 +2,6 @@ CacheEnable disk /
 CacheRoot "<%= @_cache_root %>"
 CacheDirLevels 2
 CacheDirLength 1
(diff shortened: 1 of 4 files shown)


templates/mod/disk_cache.conf.erb
```diff
@@ -2,3 +2,6 @@ CacheEnable disk /
 CacheRoot "<%= @_cache_root %>"
 CacheDirLevels 2
 CacheDirLength 1
+<% if scope['cache_ignore_headers'] != :undef -%>
+CacheIgnoreHeaders <%= @cache_ignore_headers %>
+<% end -%>
```
ekohl: In general all files have a newline at the end.

templates/mod/disk_cache.conf.erb
```diff
@@ -2,3 +2,6 @@ CacheEnable disk /
 CacheRoot "<%= @_cache_root %>"
 CacheDirLevels 2
 CacheDirLength 1
+<% if scope['cache_ignore_headers'] != :undef -%>
```
ekohl: I think you can simply use ```<% if @cache_ignore_headers -%>``` here

gitkent: I'd looked into this before I use the scope:
https://stackoverflow.com/questions/17096934/check-if-a-variable-is-undef-in-puppet-template

It looks like by simply using if is not sufficient to know that a variable is really undef.

ekohl: It's good to note that that answer is from 2013. Puppet has evolved a lot so I'm pretty sure that's no longer needed.

david22swan:
@gitkent please alter the docs to mirror your changes and add test coverage.

PR #1720, opened 2017-11-14
Fix case of setting apache::mpm_module to false

edestecd:
This commit: https://github.com/puppetlabs/puppetlabs-apache/commit/0dcfa2a0109fc98a5fae76fb55cc9f4dbbb1c48c#diff-2c378354f8ec309dd88357170e38035d

Breaks a common setup of setting  $::apache::mpm_module to false and declaring your own mpm module to set params etc. and results in the following error:
```

Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Left match operand must result in a String value. Got a Boolean. at /etc/puppetlabs/code/environments/production/modules/apache/manifests/mod/cgi.pp:5:10
```

This restores that functionality while still maintaining the intent of the original commit.
The net effect should be the same, but it handles more cases better.

### DIFF
--- a/manifests/mod/cgi.pp
+++ b/manifests/mod/cgi.pp
@@ -1,9 +1,14 @@
 class apache::mod::cgi {
+  include ::apache
   case $::osfamily {
     'FreeBSD': {}
     default: {
-      if $::apache::mpm_module =~ /^(itk|peruser|prefork)$/ {
-        Class["::apache::mod::${::apache::mpm_module}"] -> Class['::apache::mod::cgi']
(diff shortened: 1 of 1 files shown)


edestecd:
A similar pattern to https://github.com/MiamiOH/puppetlabs-apache/blob/7d43453293a54b0479efd753b3b6950d48d5890e/manifests/mod/cgid.pp

was used.

willmeek:
Thank you for your contribution, @edestecd

PR #1713, opened 2017-10-20
this PR fixes tests for #1699

zivis:
PR #1699 introduces operatingsystemmajrelease and this fact has to be added to the tests.
so this PR superseeds #1699

PR #1699, opened 2017-09-23
[MODULES-5644] : Package name is libapache2-mpm-itk for Debian 9

remicellier:

PR #1724, opened 2017-11-17
[MODULES-5644] Package name is libapache2-mpm-itk for Debian 9

zivis:
This PR superseeds #1699 and #1713 

It does basically the same as #1699 but 
* without introducing fact operatingsystemmajrelease instead it uses operatingsystemrelease so the tests do not have to be altered…
* using versioncmp for robustness

### DIFF
--- a/manifests/mpm.pp
+++ b/manifests/mpm.pp
@@ -83,7 +83,7 @@ define apache::mpm (
         }
       }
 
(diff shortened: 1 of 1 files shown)


pmcmaw:
@zivis thank you for your contribution! :-)

PR #1729, opened 2017-11-25
Set default keepalive to on in all distros

hfm:
According to README, $keepalive is off by default but only RedHat and Amazon are set to on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1095,7 +1095,7 @@ Determines whether to enable persistent HTTP connections with the [`KeepAlive`][
 
 Values: 'Off', 'On'.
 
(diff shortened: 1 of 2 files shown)


ekohl:
It was changed in d90bfbd3e074465c1b074618e8bdff23d2de9ded and I agree it's a good change. It also matches the apache defaults according to the docs. IMHO the other distros should also default to ```On``` instead or the docs changed to reflect the reality.

hfm:
Okay, I'll change default keepalive to `On` in the other distros.

hfm:
I changed  by ffd9282.

PR #1732, opened 2017-12-03
Fix typos of README.md

hfm:


tphoney:
Thanks for the diligent work @hfm

PR #1735, opened 2017-12-07
#puppethack Reduce Travis test parallelism, take 2

jearls:
Set the environment variable `PARALLEL_TEST_PROCESSOR` in the global `env` section to avoid overloading the Travis testing containers and killing the tests.
Also add this to `.sync.yml` so it will be permanent (unlike my previous attempt at this).


pmcmaw:
@jearls I will get this merged when it runs through Travis CI. 
Thank you for not only taking the time to contribute but for also participating in our Puppethack.

:-)

PR #1747, opened 2018-01-13
Security HTTP headers

juju4:

This adds all usual Security HTTP headers that are usually advised for a modern website. see also securityheaders.io

1 failure I didn't fix or understand in travis
```
Failures:
  1) apache::vhost os-independent items not everything can be set together... should not contain Class[apache::mod::headers]
     Failure/Error: it { is_expected.to_not contain_class('apache::mod::headers') }
       expected that the catalogue would not contain Class[apache::mod::headers]
     # ./spec/defines/vhost_spec.rb:939:in `block (4 levels) in <top (required)>'
Finished in 4 minutes 8.2 seconds (files took 16.06 seconds to load)
290 examples, 1 failure
```

Can you point me what needs to be fixed?

Thanks

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -100,7 +100,24 @@ define apache::vhost(
   $redirectmatch_dest                                                               = undef,
   $rack_base_uris                                                                   = undef,
   $passenger_base_uris                                                              = undef,
-  $headers                                                                          = undef,
+## FIXME! how to ensure those headers are not overwrited too easily?
+  $headers                                                                          = [
+    'set X-Content-Type-Options "nosniff"',
+    'set X-Frame-Options "sameorigin"',
+    "set Strict-Transport-Security \"max-age=16070400; includeSubDomains\"",
+    ## https://www.w3.org/TR/upgrade-insecure-requests/
+    "set Upgrade-Insecure-Requests \"1\"",
+    "set X-XSS-Protection \"1; mode=block\"",
+    "set Content-Security-Policy \"default-src 'self'; script-src 'self' 'unsafe-inline'; connect-src 'self'; img-src 'self'; style-src 'self' 'unsafe-inline'; object-src 'self'; upgrade-insecure-requests; report-uri /csp/report.php\"",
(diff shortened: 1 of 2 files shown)


hunner:
This PR changes the conditional that determines when `apache::mod::headers` is included so the failing test needs to be modified to match for the conditions.

Because `$security_headers` is always set, then the headers class will always be included. Is this intended? Is this a backwards-compatible change?

hunner:
Ah yes, it looks like this functionality is available via the `$headers` parameter already.

juju4:
Decision is yours but security HTTP headers are hardly business logic, they are infra configuration and the reason why I submit them as default is they are most often NOT configured.
Why I differentiate them from default $headers, is: people will most probably silently overwrite it and don't even see it. Whereas with a different one, they will have to unset it if they don't want it.
= sane secure default.
And yes, it makes headers module/class required but I think that it's most often the case. I don't know any alternative not using extra module.

See for example
https://securityheaders.io/
https://scotthelme.co.uk/hardening-your-http-response-headers/
https://www.troyhunt.com/locking-down-your-website-scripts-with-csp-hashes-nonces-and-report-uri/
https://www.troyhunt.com/understanding-http-strict-transport/
https://www.veracode.com/blog/2014/03/guidelines-for-setting-security-headers
https://www.veracode.com/blog/2012/11/security-headers-report
https://www.veracode.com/blog/2015/11/security-headers-top-1000000-websites-november-2015-report

ekohl:
While I agree, we should use security headers, I think this exact default breaks for 99.9% users and they need to modify it. If you're modifying it that much, the default is incorrect and does more harm than good. HSTS headers with such a long cache duration are very likely to lead to misconfigurations and could then be cached for a long time. CSP takes intimate knowledge of the application and careful configuration. What if the application is already sending out these headers? Will Apache or the application win?

Considering every user has to change this, it'll create more annoyance and people will dislike the module for it. I know I would. From this module I just expect a plan abstraction over apache without any preferences as to how I set up my server. Sadly I don't see a way to do security that works for most users in this case.

PR #1748, opened 2018-01-13
security.txt configuration - https://securitytxt.org/

juju4:

This adds support for security.txt, a file like robots.txt bug giving contact information if vulnerabilities are found on the website or subdomains.

1 failure in travis that I didn't find how to solve if you can help
```
1) apache::vhost os-independent items when not setting nor managing the docroot should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile }
       error during compilation: Parameter path failed on File[false/rspec.example.com-security.txt]: File paths must be fully qualified, not 'false/rspec.example.com-security.txt' at /home/travis/build/juju4/puppetlabs-apache/spec/fixtures/modules/apache/manifests/vhost.pp:1089
     # ./spec/defines/vhost_spec.rb:1096:in `block (4 levels) in <top (required)>'
Finished in 5 minutes 57 seconds (files took 17.28 seconds to load)
290 examples, 1 failure
```

Thanks

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -150,6 +150,8 @@ class apache::params inherits ::apache::version {
       'nss' => 'libmodnss.so',
     }
     $conf_template        = 'apache/httpd.conf.erb'
+    $security_txt         = true
+    $security_txt_template  = 'apache/vhost/security.txt.erb'
     $http_protocol_options  = undef
     $keepalive            = 'On'
     $keepalive_timeout    = 15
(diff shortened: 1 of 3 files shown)


hunner:
This is interesting from an app developer standpoint, but seems very much not related to Apache and more related to operational workflows. The apache module really tries to be hands-off when it comes to site contents and only manage configuration strictly in the apache domain.

PR #1749, opened 2018-01-13
php config hardening

juju4:

Customizable options for php with sane and secure defaults

See also
* https://github.com/sensiolabs/security-checker
* https://github.com/sektioneins/pcc
* https://github.com/psecio/iniscan

1 failure for unknown reason
```
Failures:

  1) apache::mod::php class custom extensions, php_flag, php_value, php_admin_flag, and php_admin_value File "/etc/httpd/conf.d/25-php.example.com.conf" should contain "  php_value include_path \"/usr/share/pear:/usr/bin/php\""
     [31mFailure/Error: [0m[32mit[0m { is_expected.to contain [31m[1;31m"[0m[31m  php_value include_path [1;35m\"[0m[31m/usr/share/pear:/usr/bin/php[1;35m\"[0m[31m[1;31m"[0m[31m[0m }[0m
     [31m  expected File "/etc/httpd/conf.d/25-php.example.com.conf" to contain "  php_value include_path \"/usr/share/pear:/usr/bin/php\""[0m
     [31m  [0m
     [36m# ./spec/acceptance/mod_php_spec.rb:93:in `block (4 levels) in <top (required)>'[0m

Finished in 17 minutes 31 seconds (files took 2 minutes 17.3 seconds to load)
[31m528 examples, 1 failure[0m
```

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -8,6 +8,18 @@ class apache::mod::php (
   $source           = undef,
   $root_group       = $::apache::params::root_group,
   $php_version      = $::apache::params::php_version,
+  # php security settings
+  $php_date_timezone = 'UTC',
+  $php_expose_php   = 'Off',
+  $php_allow_url_fopen    = 'Off',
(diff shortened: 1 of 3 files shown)


hunner:
I'm pretty sure that if the apache module started managing php.ini then other people's configurations would be quite broken. Is there a reason that apache should be the module managing php.ini and not some additional module?

Also, the `ini_setting` resource type is much better to use than `file_line` from the puppetlabs-inifile module when managing ini settings.

juju4:
I think whatever installs a module, should configured it. seems strange to install "broken/incomplete" stuff but it is an open question and I fear no definitive answer.

ekohl:
I'd say a module like https://github.com/voxpupuli/puppet-php is specifically aimed at configuring PHP and this module should IMHO not attempt to do so. Then it should be solved at the profile layer that you ensure it's applied for the correctly vhosts. Mixing these concerns has proven to lead to too specific modules and it's why the community has come up with the [roles and profiles pattern](https://www.craigdunn.org/2012/05/239/) many years ago.

hunner:
It would be great to see this functionality in the Vox Pupuli php module! I think this change is not outside the interests of apache, but it is outside the domain of apache management for sure.

juju4:
Thanks @ekohl @hunner 
will review for a PR on voxpupuli

PR #1751, opened 2018-01-13
add basic csp reporturi page in php

juju4:
Add a (very) basic reporturi page in php for CSP reporting.
inspired from https://mathiasbynens.be/notes/csp-reports

might be better in separate file/class with a conditional. comments?

### DIFF
--- a/manifests/mod/security.pp
+++ b/manifests/mod/security.pp
@@ -130,7 +130,7 @@ class apache::mod::security (
 
   unless $::operatingsystem == 'SLES' { apache::security::rule_link { $activated_rules: } }
 
-  file { "${docroot}/csp":
+  file { "${::apache::params::docroot}/csp":
     ensure  => directory,
     owner   => $::apache::params::user,
     group   => $::apache::params::group,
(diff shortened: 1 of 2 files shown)


manifests/mod/security.pp
```diff
@@ -130,4 +130,35 @@
 
   unless $::operatingsystem == 'SLES' { apache::security::rule_link { $activated_rules: } }
 
+  file { "${::apache::params::docroot}/csp":
```
hunner: I think the `$apache::params::docroot` can be overridden by the base class so is not a reliable location.

manifests/mod/security.pp
```diff
@@ -130,4 +130,35 @@
...
+    ensure  => directory,
+    owner   => $::apache::params::user,
```
hunner: User and group can also be overridden in the base class.

manifests/mod/security.pp
```diff
@@ -130,4 +130,35 @@
...
+    group   => $::apache::params::group,
+    mode    => '0750',
+    force   => true,
+    recurse => true,
+    require => Package['httpd'],
+  }
+
+  file { 'report.php':
```
hunner: There are probably tons of servers with `apache::mod::security` without php installed. These files should be optional and disabled by default.

As you mention, this is probably best moved to its own separate class or define.

PR #1752, opened 2018-01-14
SCL support for httpd and php7.1

kny78:
* SCL is short for Software Collections on CentOS and RHEL.
* Only preliminary / experimental support for php and apache.
* Activaed by adding parameters for apache::params like this (hiera):
	apache::params::scl: true
	apache::params::scl_httpd: 'httpd24'
	apache::params::scl_php_version: '7.1'

**I would like feedback on whether this is a good way to implement support for SCL.**

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -12,7 +12,12 @@ class apache::mod::php (
 
   include ::apache
   $mod = "php${php_version}"
-
+  if $::apache::params::scl_httpd != undef and  $::apache::params::scl_php_version == undef {
+    fail("If you define apache::scl_httpd, then you also need to specify apache::scl_php_version")
+  }
+  if $::apache::params::scl_httpd == undef and  $::apache::params::scl_php_version != undef {
+    fail("If you define apache::scl_php_version, then you also need to specify apache::scl_httpd")
+  }
   if defined(Class['::apache::mod::prefork']) {
     Class['::apache::mod::prefork']->File["${mod}.conf"]
   }
@@ -49,27 +54,31 @@ class apache::mod::php (
     $_package_name = undef
   }
 
-  $_lib = "libphp${php_version}.so"
   $_php_major = regsubst($php_version, '^(\d+)\..*$', '\1')
+  $_php_version_no_dot = regsubst($php_version, '\.', '')
+  if $apache::version::scl_httpd {
+    $_lib = "librh-php${_php_version_no_dot}-php${_php_major}.so"
+  } else {
+    $_lib = "libphp${php_version}.so"
+  }
 
   if $::operatingsystem == 'SLES' {
-      ::apache::mod { $mod:
-        package        => $_package_name,
-        package_ensure => $package_ensure,
-        lib            => 'mod_php5.so',
-        id             => "php${_php_major}_module",
-        path           => "${::apache::lib_path}/mod_php5.so",
-      }
-    } else {
-      ::apache::mod { $mod:
-        package        => $_package_name,
-        package_ensure => $package_ensure,
-        lib            => $_lib,
-        id             => "php${_php_major}_module",
-        path           => $path,
-      }
-
+    ::apache::mod { $mod:
+      package        => $_package_name,
+      package_ensure => $package_ensure,
+      lib            => 'mod_php5.so',
+      id             => "php${_php_major}_module",
+      path           => "${::apache::lib_path}/mod_php5.so",
     }
+  } else {
+    ::apache::mod { $mod:
+      package        => $_package_name,
+      package_ensure => $package_ensure,
+      lib            => $_lib,
+      id             => "php${_php_major}_module",
+      path           => $path,
+    }
+  }
 
 
   include ::apache::mod::mime
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -19,6 +19,7 @@
 # Sample Usage:
 #
 class apache::params inherits ::apache::version {
+
   if($::fqdn) {
     $servername = $::fqdn
   } else {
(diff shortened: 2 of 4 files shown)


manifests/params.pp
```diff
@@ -18,7 +18,11 @@
 #
 # Sample Usage:
 #
-class apache::params inherits ::apache::version {
+class apache::params (
+ Boolean $scl            = false,
```
hunner: It is historically the policy of module design to not allow the params.pp to take parameters due to the ease of fragmenting parameters between the base class and the params class from ye olde days.

Since parse order always has the params class coming before the base class, when there is an exception to modify the behavior of params then the canonical pattern is to have a "globals" class that affects params feeding into the base class. Three examples:
- https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/globals.pp (due to many of the parameters being used by both `postgresql::server` and `postgresql::client` with no base class common between them. Many of these should be in the base class and the globals class should only be used for the repo management)
- https://github.com/voxpupuli/puppet-mongodb/blob/master/manifests/globals.pp (you can see that this is already starting to have parameter creep even WITH a base class)

Since these three parameters do seem to be repo related, it would be best to create an `apache::globals` class for them.

polpot78: I will move them to globals. This is a cleaner design.

But I think we should also look at: https://docs.puppet.com/puppet/4.3/lookup_quick_module.html
But this is a lot larger task...

manifests/globals.pp
```diff
@@ -0,0 +1,10 @@
+#
+# Global apache configuration.
+#
+class apache::globals(
+  Boolean $scl            = false,
```
ekohl: I think there shouldn't be a single SCL parameter. You want one for httpd and PHP. You could solve this by making ```scl_httpd``` an ```Optional[String]``` where undef is the current behavior of no SCL.

manifests/mod/php.pp
```diff
@@ -49,8 +49,13 @@
     $_package_name = undef
   }
 
-  $_lib = "libphp${php_version}.so"
   $_php_major = regsubst($php_version, '^(\d+)\..*$', '\1')
+  $_php_version_no_dot = regsubst($php_version, '\.', '')
```
ekohl: If you had a ```php_scl``` that's actually the SCL name then it'd become:

```puppet
if $apache::php_scl {
  $_lib = "lib${apache::php_scl}-${mod}.so"
} else {
  $_lib = "lib${mod}.so"
}
```

Note the above assumes you change params to inherit from globals.

manifests/params.pp
```diff
@@ -19,6 +19,8 @@
 # Sample Usage:
 #
 class apache::params inherits ::apache::version {
+  include apache::globals
```
ekohl: Looking at this it might be better to change ```apache::version``` to accept parameters rather than to introduce ```apache::globals```.

manifests/params.pp
```diff
@@ -52,14 +54,121 @@
   # no client certs should be trusted for auth by default.
   $ssl_certs_dir          = undef
 
-  if ($::operatingsystem == 'Ubuntu' and $::lsbdistrelease == '10.04') or ($::operatingsystem == 'SLES') {
+  if ($::apache::globals::scl) {
+    $_scl_httpd = $::apache::globals::scl_httpd
+    $_scl_php_version = $::apache::globals::scl_php_version
+    $verify_command = "/opt/rh/${_scl_httpd}/root/usr/sbin/apachectl -t"
```
ekohl: Would ```scl enable ${_scl_httpd} 'apachectl -t'``` be better?

manifests/params.pp
```diff
@@ -52,14 +54,121 @@
...
+  } elsif ($::operatingsystem == 'Ubuntu' and $::lsbdistrelease == '10.04') or ($::operatingsystem == 'SLES') {
     $verify_command = '/usr/sbin/apache2ctl -t'
   } elsif $::operatingsystem == 'FreeBSD' {
     $verify_command = '/usr/local/sbin/apachectl -t'
   } else {
     $verify_command = '/usr/sbin/apachectl -t'
   }
-  if $::osfamily == 'RedHat' or $::operatingsystem =~ /^[Aa]mazon$/ {
+  if ($::apache::globals::scl) {
```
ekohl: This is a RH specific technology so I'd think it should be an if clause within the RH osfamily. Also allows you to reduce a lot of variable duplication.

manifests/params.pp
```diff
@@ -52,14 +54,121 @@
...
+    $user                 = 'apache'
+    $group                = 'apache'
+    $root_group           = 'root'
+    $apache_name          = "${_scl_httpd}-httpd"
```
ekohl: I'd advise introducing helpers (pseudo-code):
```puppet
if $scl {
  $_scl_httpd_prefix = "${_scl_httpd}-"
  $httpd_root = "/opt/rh/${_scl_httpd}/root"
} else {
  $_scl_httpd_prefix = ''
  $httpd_root = ''
}

$apache_name = "${_scl_httpd_prefix}httpd"
$httpd_dir = "${httpd_root}/etc/httpd"
```

Now you can use that within the RH block without any conditionals.

manifests/globals.pp
polpot78: I agree that it is better to make scl_httpd and scl_php_version optional, such that the user has to make a thought-through choice.

As for not having a scl-version, I think that when SCL is used, it is more like a separate distro, than just an updated package, for most everything changes: package names, service names, location on disk, modules etc.

manifests/params.pp
polpot78: It is RedHat that provides it, but I think that it is very different from normal RH software.

Therefore I would like to hold scl separate from normal RedHat-config as it will be fewer lines, but each line will be twice as complex to understand. And changes to SCL will then have the possibility to affect RedHat configuration.

david22swan:
@ekohl Are you happy that the changes you suggested have been completed?

manifests/params.pp
```diff
@@ -52,14 +53,121 @@
...
+    $passenger_conf_file  = 'passenger_extra.conf'
+    $passenger_conf_package_file = 'passenger.conf'
+    $passenger_root       = undef
+    $passenger_ruby       = undef
+    $passenger_default_ruby = undef
+    $suphp_addhandler     = 'php5-script'
+    $suphp_engine         = 'off'
+    $suphp_configpath     = undef
+    $php_version          = $_scl_php_version
+    $mod_packages         = {
```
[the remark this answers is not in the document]
ekohl: I'd start with an acceptance test that verifies the basic scenario works where you only override the SCL parameters and then check if the package gets installed and all the default mods installed.

polpot78:
I have now updated with changes from the PR. Rebased into one commit as well.
Last changes:
 - Removed redundant $apache::version::scl
 - Used local variable
 - Validate that either none or both of scl_httpd or scl_php_version is set. (mod/php.pp)
  - Use scl_php_version in params for php-version.

polpot78:
I have now updated with the changes you proposed in your comments.
On Wed, 2018-02-21 at 11:31 +0000, Ewoud Kohl van Wijngaarden wrote:
> @ekohl commented on this pull request.
> 
> 
> 
> 
> In manifests/params.pp:
> 
> > +    $lib_path             = 'modules'
> +    $mpm_module           = 'prefork'
> +    $dev_packages         = "${_scl_httpd}-httpd-devel"
> +    $default_ssl_cert     = '/etc/pki/tls/certs/localhost.crt'
> +    $default_ssl_key      = '/etc/pki/tls/private/localhost.key'
> +    $ssl_sessioncache     = '/var/cache/mod_ssl/scache(512000)'
> +    $passenger_conf_file  = 'passenger_extra.conf'
> +    $passenger_conf_package_file = 'passenger.conf'
> +    $passenger_root       = undef
> +    $passenger_ruby       = undef
> +    $passenger_default_ruby = undef
> +    $suphp_addhandler     = 'php5-script'
> +    $suphp_engine         = 'off'
> +    $suphp_configpath     = undef
> +    $php_version          = $_scl_php_version
> +    $mod_packages         = {
> 
> I'd start with an acceptance test that verifies the basic scenario
> works where you only override the SCL parameters and then check if
> the package gets installed and all the default mods installed.
> 
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 
> 
> 
>   
>   
> 
> 
> 
> 
> 
> {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c5
> 5493e4bb","name":"GitHub"},"entity":{"external_key":"github/puppetlab
> s/puppetlabs-apache","title":"puppetlabs/puppetlabs-
> apache","subtitle":"GitHub repository","main_image_url":"https://clou
> d.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-
> 95fc-
> 7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent
> .com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-
> b52498112777.png","action":{"name":"Open in GitHub","url":"https://gi
> thub.com/puppetlabs/puppetlabs-
> apache"}},"updates":{"snippets":[{"icon":"PERSON","message":"@ekohl
> commented on #1752"}],"action":{"name":"View Pull Request","url":"ht
> tps://github.com/puppetlabs/puppetlabs-
> apache/pull/1752#discussion_r169608544"}}}
-- 
Kjetil Nygård
pulk.hesten@gmail.com
Tlf +47 4147 4337

david22swan:
@ekohl Are you ahppy that the changes that you requested have been accomplished?

mmoll:
@kny78 this needs a rebase :)
@ekohl is this good from your perspective?

david22swan:
@kny78 I'm sorry, but due to the length of time that has gone by without you responding when pinged I feel that I must close this PR. I apologize if this inconveniences you in any way and we will of course be more than happy to review it again in the future if you ever decide to return to it, or submit some new work.
Best Wishes

mmoll:
@david22swan see GH-1822

PR #1755, opened 2018-01-28
Add a missing END statement in the SSL Vhost Template

diranged:
It looks to me like 7bb35c2293c12ce52329a4391fe1f20389efef06 (MODULES-5471) introduced a bug. It adds an IF statement that never ends, causing everything below it to not evaluate properly.


hunner:
The indentation is odd and kinds of hides the fact that the `end` is down on line 39. All of the SSL directives in the middle are only valid when `SSLVerifyClient` is true so it doesn't make sense to include them otherwise.

At least that's what I understand. If this is incorrect, feel free to move those outside the `if/end` block!

PR #1756, opened 2018-01-30
MODULES-6225 :  fix worker mantain typo

cedef:
Fix key 'worker_mantain' in $workers_file_content

- Add spec tests for advanced mod_jk configuration
- Fix *worker_mantain* typo (see below)
- Call `fail()` function when using the deprecated *worker_mantain* key

Original comment in worker.properties.erb reference a key
*worker_mantain* in the `$workers_file_content` hash.

From the original Apache's mod_jk documentation [1], the parameter name is
`worker.maintain` and has been mistyped as `worker.mantain` (note the
missing *i* letter).

[1] https://tomcat.apache.org/connectors-doc/reference/workers.html

Here is the [Reference in Jira](https://tickets.puppetlabs.com/browse/MODULES-6225)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2276,14 +2276,14 @@ Default: '80'
 **workers\_file\_content**
 
 Each directive has the format `worker.<Worker name>.<Property>=<Value>`. This maps as a hash of hashes, where the outer hash specifies workers, and each inner hash specifies each worker properties and values.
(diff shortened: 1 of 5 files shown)


pmcmaw:
@cedef thank you for taking the time to contribute to this module, it is much appreciated 👍

PR #1762, opened 2018-02-14
3.0.0 Pre Release

david22swan:
Also includes small rubocop fix.

### DIFF
--- a/spec/classes/mod/jk_spec.rb
+++ b/spec/classes/mod/jk_spec.rb
@@ -14,13 +14,13 @@ describe 'apache::mod::jk', type: :class do
   end
 
   shared_examples 'specific workers_file' do |mod_dir|
-    #let (:pre_condition) do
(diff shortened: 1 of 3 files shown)


spec/classes/mod/jk_spec.rb
```diff
@@ -36,26 +36,25 @@
         },
       }
     end
+
     it { is_expected.to compile }
     it { is_expected.to compile.with_all_deps }
+    expected_content = "# This file is generated automatically by Puppet - DO NOT EDIT\n"\
```
tphoney: @david22swan is this intended ?

david22swan: Yep, it failed the Rubocop merge test

david22swan: 'it's' aren't allowed to be more than 5 lines longs.

PR #1764, opened 2018-02-20
MODULES-6635: add option to pass SSLCertificateFile and SSLCertificateKey to ssl.conf

sudodevnull:
This will allow for the option to use mod_ssl with cert file and key if/when vhosts are not defined. Referenced ticket : [MODULES-6635](https://tickets.puppetlabs.com/browse/MODULES-6635)


manifests/mod/ssl.pp
```diff
@@ -3,6 +3,8 @@
   $ssl_cryptodevice                                         = 'builtin',
   $ssl_options                                              = [ 'StdEnvVars' ],
   $ssl_openssl_conf_cmd                                     = undef,
+  $ssl_cert                                                 = undef,
```
ekohl: Should this be ```Optional[String] $ssl_cert``` or even ```Optional[Stdlib::Absolutepath] $ssl_cert```? Same for ssl_key.

sudodevnull: updated.

HelenCampbell:
Thanks @sudodevnull ! Merging.

sudodevnull:
QQ, when will the version be updated so that I may use the new feature?

HelenCampbell:
Probably some time within the next month, the modules due a conversion with PDK and it will be released straight after :)

sudodevnull:
Any better ETA at this point?  I would prefer not to pull down the apache module locally and make modifications to include this feature.  
Management is pushing for an ETA for me to be able to apply our apache module to production and I need this feature in order to not break things.

HelenCampbell:
Apologies! I'm literally in the middle of one at the moment, it should be out momentarily 👍 @sudodevnull

sudodevnull:
Awesome thank you!

PR #1766, opened 2018-02-28
ensure mpm_event is disabled under debian 9 if mpm itk is used

zivis:
On debian 9 using mod_itk as mpm with: 
apache 2.4.25-3+deb9u3 and 
libapache2-mpm-itk 2.4.7-04-1

You have to disable mpm_event before installing libapache2-mpm-itk …

smells a bit like https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=734865 although it should be fixed…

PR #1767, opened 2018-03-02
(MODULES-6677) make sure event mpm is disabled

eputnam:
Currently, on Debian 9 specifically, the event mod may be enabled when trying to use a different mpm. This ensures event is disabled if another mpm is chosen. Also closes #1766

### DIFF
--- a/manifests/mpm.pp
+++ b/manifests/mpm.pp
@@ -73,20 +73,19 @@ define apache::mpm (
         }
       }
 
-      if $mpm == 'itk' and $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease == '14.04' {
-        # workaround https://bugs.launchpad.net/ubuntu/+source/mpm-itk/+bug/1286882
-        exec {
-          '/usr/sbin/a2dismod mpm_event':
-            onlyif  => '/usr/bin/test -e /etc/apache2/mods-enabled/mpm_event.load',
-            require => Package['httpd'],
-            before  => Package['apache2-mpm-itk'],
-        }
-      }
-
       if $mpm == 'itk' and ( ( $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease == '16.04' ) or ( $::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9.0.0') >= 0 ) ) {
         $packagename = 'libapache2-mpm-itk'
       } else {
         $packagename = "apache2-mpm-${mpm}"
+      }
+
+      if $mpm == 'itk' and ( ( $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease == '14.04' ) or ($::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9.0.0') >= 0 ) ) {
+        # workaround https://bugs.launchpad.net/ubuntu/+source/mpm-itk/+bug/1286882
+        exec { '/usr/sbin/a2dismod mpm_event':
+          onlyif  => '/usr/bin/test -e /etc/apache2/mods-enabled/mpm_event.load',
+          require => Package['httpd'],
+          before  => Package[$packagename],
+        }
       }
 
       if versioncmp($apache_version, '2.4') < 0 or $mpm == 'itk' {
--- a/manifests/mpm.pp
+++ b/manifests/mpm.pp
@@ -79,10 +79,20 @@ define apache::mpm (
         $packagename = "apache2-mpm-${mpm}"
       }
 
+      $mod_enabled_dir = $::apache::mod_enable_dir
+
+      if $mpm == 'prefork' and ( $::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9.0.0') >= 0 ) {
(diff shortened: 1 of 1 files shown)


manifests/mpm.pp
```diff
@@ -73,22 +73,31 @@
...
       if $mpm == 'itk' and ( ( $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease == '16.04' ) or ( $::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9.0.0') >= 0 ) ) {
         $packagename = 'libapache2-mpm-itk'
       } else {
         $packagename = "apache2-mpm-${mpm}"
       }
 
+      $mod_enabled_dir = $::apache::mod_enable_dir
+
+      if $mpm == 'prefork' and ( $::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9.0.0') >= 0 ) {
+        exec { '/usr/sbin/a2dismod mpm_event':
```
ekohl: Isn't this basically equivalent to a ```file``` with ```ensure => absent``` or does a2dismod do more than that? In that case you could even generalize it to ```if $mpm != 'event'``` instead of the OS checks.

mpdude:
Just hit this bug with Ubuntu 17.10, will give your PR a try

mpdude:
Is there a simple `:ref` spec I can use in `librarian-puppet` to install a PR branch like this?

ekohl:
@mpdude I guess ```mod 'puppetlabs-apache', :git => 'https://github.com/eputnam/puppetlabs-apache' :ref => 'mpm_event_disable_debian9_itk'``` should work

manifests/mpm.pp
eputnam: I assumed it did something else but it looks like all it does is remove the symlink in mods-enabled. 

I'm hesitant to generalize the conditional only because caution. However, I can't think of a situation where you wouldn't want to have event disabled if it wasn't the selected mpm. Further, maybe there should be logic to exclude any mpm that's not the selected one.

ekohl: > Further, maybe there should be logic to exclude any mpm that's not the selected one.

Heading to directory purging here :)

mpdude:
👍 Works for me on Ubuntu 17.10.

eputnam:
ran into some hard-to-get-around duplicate declarations when trying to use file resource, so I reverted back to the execs.

manifests/mpm.pp
hunner: Yes, that's basically symlink management.

The `a2*mod` commands actually were previously used by the apache module via the `a2mod` type (it's still there, deprecated and never removed; see https://github.com/puppetlabs/puppetlabs-apache/blob/3.0.0/lib/puppet/type/a2mod.rb). We tried writing providers to give similar functionality on several OSs, but it was very difficult to track all of the different changes across many OSs and each OS version.

The apache module was probably the first to specifically deviate from replicating OS defaults (RedHat-like on redhats and Debian-like on debian) and instead go for a one-size-fits-all. We did keep the mod symlink style via `apache::mod` but relied on recursive purging to keep everything inline inside the conf directory. This was 0.7.0 iirc.

What specifically causes the mpm module .load files to clash and not be purged? Is it a dependency between the mods and the file purging? Why do we have to exec out to `a2*mod` at all?

maxdelorme:
If apache > 2.4 and prefork (required for php in my  case) then Package[$packagename] is not installed
see https://github.com/eputnam/puppetlabs-apache/blob/mpm_event_disable_debian9_itk/manifests/mpm.pp#L106
and the before condition : https://github.com/eputnam/puppetlabs-apache/blob/mpm_event_disable_debian9_itk/manifests/mpm.pp#L88 can not be verified

resulting in error : 
`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Could not find resource 'Package[apache2-mpm-prefork]' in parameter 'before' (file: /etc/puppetlabs/code/environments/<my_env>/modules/apache/manifests/mpm.pp, line: 88) on node <my_node>`

zivis:
Is there anything i can do, to help getting this or this: !1767 merged?
If something is not okay with neither of those solutions we can work on it…

cmseal:
This doesn't include a condition for 'prefork' on Ubuntu 18.04... could this be included after line 75:

```
      if $mpm == 'prefork' and ( ( $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease == '18.04' ) or ( $::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9.0.0') >= 0 ) ) {
         exec { '/usr/sbin/a2dismod mpm_event':
           onlyif  => "/usr/bin/test -e ${mod_enabled_dir}/mpm_event.load",
           require => Package['httpd'],
           before  => Package[$packagename],
         }
       }
```

zivis:
@mpdude @cmseal @maxdelorme i have reworked #1766 you can have a look at.
It should work for the scenarios mentioned here.

HelenCampbell:
Closing due to #1766 being merged. Thank you everyone for your work and input!

PR #1768, opened 2018-03-05
(WIP) PDK Convert 1.4.1

HelenCampbell:


.travis.yml
```diff
@@ -1,34 +1,70 @@
-#This file is generated by ModuleSync, do not edit.
 ---
 sudo: false
+dist: trusty
 language: ruby
 cache: bundler
-script: "bundle exec rake release_checks"
+before_install:
+  - bundle -v
+  - rm Gemfile.lock || true
```
ekohl: I think ```rm -f``` is cleaner than ```|| true```

metadata.json
```diff
@@ -72,5 +77,8 @@
       "version_requirement": ">= 4.7.0 < 6.0.0"
     }
   ],
-  "description": "Module for Apache configuration"
+  "description": "Module for Apache configuration",
+  "pdk-version": "1.4.1",
+  "template-url": "file:///opt/puppetlabs/pdk/share/cache/pdk-templates.git",
```
ekohl: It'd be nice if this was a public URL so we could have a look at it.

.travis.yml
HelenCampbell: Made a PR for this against the template https://github.com/puppetlabs/pdk-templates/pull/35

metadata.json
HelenCampbell: The template used is showing like this because it's the cached version that was shipped with the PDK. As a rule the PDK template version used is the same as that of the PDK. The template itself can be found here: https://github.com/puppetlabs/pdk-templates

HelenCampbell:
Closing this PR to get a clean shot at re-doing this, seems to be something funky going on with travis etc.

PR #1769, opened 2018-03-05
Allow overriding passenger_group in apache::vhost

smortex:
Just like we can use `passenger_user` in a VHost, allow setting `passenger_group`.

### DIFF
--- a/templates/vhost/_passenger.erb
+++ b/templates/vhost/_passenger.erb
@@ -32,7 +32,7 @@
   PassengerHighPerformance <%= scope.call_function('apache::bool2httpd', [@passenger_high_performance]) %>
 <% end -%>
 <% if @passenger_nodejs -%>
-  PassengerNodejs <%= @passenger_nodejs -%>
+  PassengerNodejs <%= @passenger_nodejs %>
 <% end -%>
 <% if @passenger_sticky_sessions -%>
   PassengerStickySessions <%= scope.call_function('apache::bool2httpd', [@passenger_sticky_sessions]) %>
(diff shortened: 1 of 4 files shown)


david22swan:
@smortex Please alter the docs to reflect your changes and add some test coverage.

smortex:
Working on it.  The test suite lacks test for the `passenger_*` params… Adding them.

smortex:
Done!  This helped finding an unrelated problem in the passenger template 😉

smortex:
@david22swan maybe the *needs-docs* and *needs-tests* labels can be removed?

templates/vhost/_passenger.erb
```diff
@@ -25,11 +25,14 @@
...
 <% end -%>
+<% if @passenger_group -%>
+  PassengerGroup <%= @passenger_group %>
+<% end -%>
 <% if @passenger_high_performance -%>
   PassengerHighPerformance <%= scope.call_function('apache::bool2httpd', [@passenger_high_performance]) %>
 <% end -%>
 <% if @passenger_nodejs -%>
-  PassengerNodejs <%= @passenger_nodejs -%>
+  PassengerNodejs <%= @passenger_nodejs %>
```
eputnam: nice catch! i hope the config parser is hardy enough to handle a newline :)

smortex: I don't think so 😉 

```
PassengerNodejs takes one argument, The Node.js command to use.
```

eputnam: ah ok, so we def need that newline

PR #1770, opened 2018-03-06
Updating translations for readmes/README_ja_JP.md

hunner:


hunner:
Translations missed the 3.0.0 boat but were still hanging out on the release branch. This brings them back into master so they don't get missed.

PR #1777, opened 2018-04-17
Re-enable the apache_version fact on FreeBSD

smortex:
The fact was confined in 783729a5f2e348a4728f3ce29f0b58c7e0a425dd but running the test suite on FreeBSD will fail because the fact is confined to Linux.

Instead of simulating Linux when testing on FreeBSD, enable the fact on FreeBSD so that the apache version is available in Puppet manifests on this platform.


pmcmaw:
@smortex thank you for your PR 👍 
Your contribution is much appreciated.

PR #1780, opened 2018-04-23
Remove insecure/bad SSL example

Feandil:
These SSL settings are worse than the default one:
- It only enables TLS1, no 1.1 or 1.2
- It enables RC4, which is supposed to be forbidden nowadays (http://www.ietf.org/rfc/rfc7465.txt)


Feandil:
The Travis-CI build error seems to be due to Ruby 2.1.9 deprecation and should be fixed by https://github.com/puppetlabs/puppetlabs-apache/pull/1779

PR #1792, opened 2018-05-21
added $ to replacment param

qs5779:
According to the apache documentation https://httpd.apache.org/docs/2.4/rewrite/remapping.html a $ sign is need before the numbered replacment param in the redirectmatch section.

Mine is working with the $ sign. I did not test without.


pmcmaw:
@qs5779 

Thank you for taking the time to submit this PR! 👍

PR #1797, opened 2018-05-31
Add mod::auth_gssapi manifest

JAORMX:
mod_auth_gssapi has a similar functionality to mod_auth_kerb, and is the
recommended way to deploy authorization with kerberos, through GSSAPI.

This merely adds the necessary module.


david22swan:
Passed on adhoc

PR #1750, opened 2018-01-13
enforce modern crypto ciphers and protocol

juju4:
Goal: provide a sane & secure default configuration for modern webserver

Per recommendation for modern setup.
* https://cipherli.st/
* https://wiki.mozilla.org/Security/Server_Side_TLS
* https://wiki.mozilla.org/Security/TLS_Configurations
* https://raymii.org/s/tutorials/Strong_SSL_Security_On_Apache2.html
* https://openweb.eu.org/articles/https-de-ssl-a-tls-1-3
* https://www.ssi.gouv.fr/uploads/2017/02/security-recommendations-for-tls_v1.1.pdf
* https://bettercrypto.org/static/applied-crypto-hardening.pdf

By default, better not to use TLS 1.0. being deprecated by Debian, Microsoft, Fastly, NewRelic, NIST and many others. PCI DSS enforces TLS1.2+.

PR #1807, opened 2018-08-01
Revert "enforce modern crypto ciphers and protocol"

HelenCampbell:
Reverts puppetlabs/puppetlabs-apache#1750

This PR doesn't work in conjunction with three of the older OSes, SLES11, Redhat5 and Centos5.
As a result I'm taking this back out.


README.md
```diff
@@ -2814,7 +2814,7 @@ To use SSL with a virtual host, you must either set the [`default_ssl_vhost`][]
 
 * `ssl_protocol`
 
-  Default: ['all', '-SSLv2', '-SSLv3', '-TLSv1', '-TLSv1.1'].
+  Default: ['all', '*SSLv2', '*SSLv3'].
```
ekohl: Probably because it's a straight revert this is changed back to `*` but can you fix this to be `-SSLv2` so it matches the default?

PR #1808, opened 2018-08-01
MODULES-7575 reverse sort the aliases

k2patel:
root location alias must be defined last in order for others to work


HelenCampbell:
Thank you!

PR #1843, opened 2018-10-12
Fix default vhost priority in readme

HT43-bqxFqB:
The README.md still references the old `15-` value as default in some places. This PR fixes that and changes it to the `25-` default.


HelenCampbell:
Thanks!

PR #1848, opened 2018-11-14
(MODULES-8429) Update GPG key for phusion passenger

abottchen:
Prior to this commit, a GPG key mismatch prevented the phusion
passenger RPM from installing on RedHat based installs.  This
commit updates the GPGKey to match the one provided by the script
https://packagecloud.io/phusion/passenger/install#bash-rpm.


HelenCampbell:
Thank you for this fix!

PR #1851, opened 2018-11-26
(MODULES-5990) Addition of 'IncludeOptional conf-enabled/*.conf' to '…

david22swan:
…apache2.conf' on Debian Family OS

### DIFF
--- a/templates/httpd.conf.erb
+++ b/templates/httpd.conf.erb
@@ -114,6 +114,9 @@ IncludeOptional "<%= @confd_dir %>/*.conf"
 <%- else -%>
 Include "<%= @confd_dir %>/*.conf"
 <%- end -%>
+<%- if @conf_enabled and scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
+IncludeOptional "<%= @conf_enabled %>/*.conf"
+<%- end -%>
 <% if @vhost_load_dir != @confd_dir -%>
 <%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
 IncludeOptional "<%= @vhost_load_dir %>/<%= @vhost_include_pattern %>"
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -69,6 +69,7 @@ class apache::params inherits ::apache::version {
     $server_root          = '/etc/httpd'
     $conf_dir             = "${httpd_dir}/conf"
     $confd_dir            = "${httpd_dir}/conf.d"
+    $conf_enabled         = undef
     $mod_dir              = $::apache::version::distrelease ? {
       '7'     => "${httpd_dir}/conf.modules.d",
       default => "${httpd_dir}/conf.d",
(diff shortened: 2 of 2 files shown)


templates/httpd.conf.erb
```diff
@@ -109,6 +109,9 @@ LogFormat "<%= format -%>" <%= nickname %>
   <%- end -%>
 <% end -%>
 
+<%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
+IncludeOptional "<%= @conf_dir %>conf-enabled/*.conf"
```
ekohl: Missing at least a slash. This will result in `/etc/apache2conf-enabled/*.conf`.

It's also rather ugly to do this on all distros, even those that don't have this mechanism. I wonder if we need a `$conf_enable_dir` that's `undef` on all other distros so you can test for it in the template.

david22swan: @ekohl Have updated the code to be along the lines of what you suggested

david22swan:
screen shot 2018-12-04 at 9 10 31 am

baurmatt:
I think this is missing something. If the folder gets included by default, it should be managed by default as well. Especially since the default for config dirs is to purge them.
If it doesn't get managed/purged, the default files shipped with e.g. Ubuntu 16.04 get included and might break some stuff.

@david22swan What do you think about it? I'm happy to fix it myself in a PR, just want your opinion if this seems reasonable.

david22swan:
@baurmatt I recently created a pr to set it so it is no longer included by default (https://github.com/puppetlabs/puppetlabs-apache/pull/1869) after it caused issues with another user. But if you have some ideas on how to improve it so that it is managed when included then I encourage you to go for it.

baurmatt:
@david22swan Thats, of course, an option as well ;) I thought about doing this:

```puppet
class apache (
...
  Optional[Stdlib::Absolutepath] $conf_enabled = $apache::params::conf_enabled,
...
) {
...
if $conf_enabled {
  file { $conf_enabled:
    ensure  => directory,
    recurse => true,
    purge    => $purge_confd,
    force     => $purge_confd,
    notify    => Class['Apache::Service'],
    require   => Package['httpd'],
  }
...
}
```

While writing this I realized the /etc/apache/conf-enabled is the correct directory for Debian based systems instead of conf.d, right? Should we just change the $confd_dir for Debian based systems to "${httpd_dir}/conf-enabled"?

david22swan:
@baurmatt Sorry for the wait, I looked into what you said and the code addition looks good. However the second part, about changing the directory from conf.d, I am not so sure. Honestly my overall apache knowledge isn't the greatest and after looking into it I can't see a strong reason for it.
Could you explain your reasoning a bit more?

baurmatt:
@david22swan No worries! :)

As far as I remember, the conf.d directory isn't available by default on Debian based systems.It's a RedHat-ish thing. On Debian based systems the folder is called conf-enabled. Therefore I think it makes sense to replace the conf.d directory with conf-enabled.

david22swan:
@baurmatt I'm still unsure, if you want to go ahead and start work on that I won't stop you, but it's not something that would get merged in quickly and would require some serious discussion throughout my  team and not just me.
Sorry for not being more help.

PR #1856, opened 2018-12-06
(FM-7605) - Disabling conf_enabled on Ubuntu 18.04  by default as it conflicts with Shibboleth causing errors with apache2.

david22swan:


david22swan:
screen shot 2018-12-06 at 12 43 36 pm

manifests/params.pp
```diff
@@ -218,7 +218,11 @@
     $server_root         = '/etc/apache2'
     $conf_dir            = $httpd_dir
     $confd_dir           = "${httpd_dir}/conf.d"
-    $conf_enabled        = "${httpd_dir}/conf-enabled"
+    if !($::operatingsystem == 'Ubuntu' and $::lsbdistrelease == '18.04') {
```
HelenCampbell: This should be moved to the Ubuntu18 specific params section further down.

manifests/params.pp
```diff
@@ -218,7 +218,11 @@
...
+      $conf_enabled      = "${httpd_dir}/conf-enabled"
+    } else {
+      $conf_enabled      = ''
```
eimlav: Could this line not go in the block for Ubuntu 18 at line 300, and then the original $conf_enabled value left untouched?

david22swan:
@HelenCampbell @eimlav Code is update

david22swan:
Have removed the exclusion to it previous location as the new one caused unit test failures.

david22swan:
@ekohl Are you happy with the code as it is now??

PR #1794, opened 2018-05-22
(MODULES-7196) Allow setting CASRootProxiedAs per virtualhost

yakatz:
The README doesn't say much about the use of CASRootProxiedAs, but it makes more sense to set it per vhost. This patch allows that.

PR #1857, opened 2018-12-10
(MODULES-7196) Allow setting CASRootProxiedAs per virtualhost (replaces #1794)

yakatz:
The README doesn't say much about the use of CASRootProxiedAs, but it makes more sense to set it per vhost. This patch allows that. Now includes a test for the new feature and rebased against master, but I can't reopen the old pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1867,7 +1867,7 @@ The `cas_login_url` and `cas_validate_url` parameters are required; several othe
 
   Default: `undef`.
 
-- `cas_root_proxied_as`: Sets the URL end users see when access to this Apache server is proxied.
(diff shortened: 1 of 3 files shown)


Lavinia-Dan:
@yakatz Hello, how is it going? Thank you for your contribution. I have taken a look at your pull request and there is one small problem when running the test , it fails for  the following supported OS(es): RedHat5, Centos5, Ubuntu 18.04, sles11, sles12, sles5 which appear to be using Apache 2.2.3, I've investigated the mod_auth_cas module and apparently it should work for any apache 2.2.x. In order to avoid any OS inconsistencies I've decided it's best to write a unit test instead of an acceptance test. I have taken the liberty to make a new PR based on your commits and with the modifications I just mentioned. Thus I will close this PR. Hope that is ok with you.

PR #1865, opened 2018-12-19
DONT MERGE !!!(FM-7655) test of setting rubygem version

tphoney:


ekohl:
In theforeman org I also saw failures. Is this to fix the issue with github_changelog_generator requiring ruby 2.2.2 despite not being needed in tests or do you see another issue? One difference is that we don't do a `gem update --system` in theforeman though. For comparison: https://travis-ci.org/theforeman/puppet-katello/builds/470122778

tphoney:
@ekohl i am testing out this change https://github.com/puppetlabs/pdk-templates/pull/171 before i merge to pdk-templates master. I will then roll this out to all supported modules assuming everything looks good. the other test prs are https://github.com/puppetlabs/puppetlabs-motd/pull/187 & https://github.com/puppetlabs/puppetlabs-mysql/pull/1149

ekohl:
Ah, looks like a different issue then though perhaps it's the same root cause.

tphoney:
we can likely clean up the travis before_install. but that is a major change before the christmas break. break being the operative word.

thanks for the eyes @ekohl 👍

PR #1867, opened 2018-12-20
Add support for mod_http2

smortex:
[`mod_http2`](https://httpd.apache.org/docs/current/mod/mod_http2.html) was introduced in Apache 2.4.17 and provides HTTP/2 ([RFC 7540](https://tools.ietf.org/html/rfc7540)) support.

This pull request allows configuring HTTP/2 websites using the apache module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1283,6 +1283,18 @@ Sets the path to the file containing Apache ports configuration.
 
 Default: '{$conf_dir}/ports.conf'.
 
+##### `protocols`
+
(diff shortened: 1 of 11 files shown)


HT43-bqxFqB:
:+1:

rtyler:
This change looks great from my perspective @smortex! Thanks for writing up this support. 

I hope it's merged soon so we can use this in the Jenkins project :smile_cat:

PR #1869, opened 2019-01-02
(MODULES-5990) - conf-enabled defaulted to undef

david22swan:


pmcmaw:
Can you run this through the adhoc? I know last time changes around this area caused failures somewhere.

david22swan:
screen shot 2019-01-04 at 3 15 07 pm

PR #1877, opened 2019-01-18
(maint) - Doc's Fix

david22swan:


david22swan:
@jemalizee The fix that you requested has gone in

PR #1880, opened 2019-01-31
MODULES-8541 : Allow HostnameLookups to be modified

k2patel:
https://tickets.puppetlabs.com/projects/MODULES/issues/MODULES-8541

Apache variable HostnameLookups should be allowed to configured, as many of the configuration behind proxy require the domain name lookup in conditional statements. It should be default value set to Off, but it should be allowed to configured.


k2patel:
Sorry, forgot to squash commits

PR #1882, opened 2019-02-03
Manage all mod_remoteip parameters supported by Apache

johanfleury:


david22swan:
@johanfleury Giving this a quick look, my first thought is the lack of doc updates. For a change like this we'd need to see updated documentation covering each added parameter.
I'd also like a more descriptive message attached to the commit, but that's not as important.
Overall looks like a good pr though, thanks for putting in the work.

johanfleury:
Thanks for the quick review @david22swan, I'll work on that.

Regarding the documentation I was wondering what would be best. Should I write it in the manifest using puppet strings syntax or should I write it in the README file?

My preference goes to the former.

david22swan:
@johanfleury If you could do both it would be great, but the manifest is the more important.

johanfleury:
@david22swan I updated the PR accordingly.

Please tell me if the documentation I added is useful and well written (English is not my first language).

david22swan:
@johanfleury Your changes look great and the pr has passed cleanly through the adhoc pipeline so I am more than happy to merge it. Thank you for putting in the work and responding so promptly.
screen shot 2019-02-08 at 10 58 01 am

PR #1894, opened 2019-03-20
Avoid duplicate declaration of the module headers.

Rabie-Zamane:
Avoid duplicate declaration of the apache module headers.
This is the case when the module is also declared in a user defined manifest that uses the puppetlabs-apache module.


Rabie-Zamane:
that's it. I changed our manifest to use the proposed pattern (apache::mod::*). So we avoid duplicate declaration now using the right way.  thank you!

PR #1899, opened 2019-03-26
MODULES-3958 Get rid of mocha-based stubbing across all modules

lionce:


spec/unit/provider/a2mod/gentoo_spec.rb
```diff
@@ -32,9 +32,11 @@
     end
 
     it 'normalizes parameters' do
-      filetype.expects(:read).returns(%(APACHE2_OPTS="-D FOO -D BAR -D BAR"\n))
-      provider_class.expects(:filetype).returns(filetype)
-
+      # filetype.expects(:read).returns(%(APACHE2_OPTS="-D FOO -D BAR -D BAR"\n))
+      allow(filetype).to receive(:read).and_return(%(APACHE2_OPTS="-D FOO -D BAR -D BAR"\n))
```
david22swan: This should be an expect.

spec/unit/provider/a2mod/gentoo_spec.rb
```diff
@@ -32,9 +32,11 @@
...
+      # provider_class.expects(:filetype).returns(filetype)
+      allow(provider_class).to receive(:filetype)
```
david22swan: Is there a reason for this?

spec/unit/provider/a2mod/gentoo_spec.rb
ekohl: Unless you add a `expect(filetype).to have_received(:read)` later on. IIRC there's a rubocop-rspec warning about this.

david22swan:
Need's a clean up to remove commented code, but other than what I've mentioned it looks fine. Unsure about the removed unstubs, the code seem's to pass without them so it look's like the new stubs are overwriting the old one's.

PR #1905, opened 2019-04-08
Remove event mpm when using prefork, worker or itk

tuxmea:
fixes https://tickets.puppetlabs.com/browse/MODULES-8903

### DIFF
--- a/manifests/mpm.pp
+++ b/manifests/mpm.pp
@@ -84,13 +84,10 @@ define apache::mpm (
         }
       }
 
+
       if $mpm == 'itk' {
         if ( ( $::operatingsystem == 'Ubuntu' ) or ( ($::operatingsystem == 'Debian') and ( versioncmp($::operatingsystemrelease, '8.0.0') >= 0 ) ) ) {
-          ensure_resource('exec', '/usr/sbin/a2dismod mpm_event', {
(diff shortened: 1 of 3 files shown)


manifests/mpm.pp
```diff
@@ -101,11 +101,21 @@
 
       if $mpm == 'prefork' {
```
ekohl:
```suggestion
      if $mpm in ['prefork', 'worker'] {
```

tuxmea: AFAIK: prefork must disable event in Debian 9, whereas worker must disable event in Debian 8.
But I might be wrong. We had this issue at a customer today.

tuxmea: @ekohl can you please review?

lionce:
Screenshot 2019-06-06 at 12 19 10

manifests/mpm/disable_mpm_event.pp
```diff
@@ -0,0 +1,8 @@
+class apache::mpm::disable_mpm_event {
+  @exec { '/usr/sbin/a2dismod mpm_event':
```
ekohl: Why a virtual resource? Couldn't you just include the class only when it's needed? You can include multiple times

How about a name and the specify the command explicitly? This makes it easier if some distro decides to change it somehow.

sheenaajay:
Thanks for submitting the PR @tuxmea .

PR #1909, opened 2019-04-12
Remove mod_ldap dependency from mod_authz_ldap module

turnopil:
An error occurs when using the mod_authz_ldap module.

`ESC[1;31mError: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, 'regsubst' parameter 'target' expects a value of type Array or String, got Undef (file: /etc/puppetlabs/code/modules/apache/manifests/mod.pp, line: 55, column: 19) (file: /etc/puppetlabs/code/modules/apache/manifests/mod/ldap.pp, line: 16) on node *******[0m`

### DIFF
--- a/manifests/mod/authnz_ldap.pp
+++ b/manifests/mod/authnz_ldap.pp
@@ -4,7 +4,6 @@ class apache::mod::authnz_ldap (
 ) {
 
   include ::apache
-  include '::apache::mod::ldap'
   ::apache::mod { 'authnz_ldap':
     package => $package_name,
   }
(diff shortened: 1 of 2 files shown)


turnopil:
Or we can set the default value _mod_ldap_ here https://github.com/puppetlabs/puppetlabs-apache/blob/d15420694c54ba6c768acbfe1c7a217c462d5230/manifests/params.pp#L232

```
      'ldap'                  => $::apache::version::distrelease ? {
        '7'     => 'mod_ldap',
        default => undef,
      },
```

PR #1913, opened 2019-04-15
(bugfix) Add default package name for mod_ldap

turnopil:
An error occurs when I'm using the mod_authz_ldap module and Apache 2.4 on Amazon1.

`ESC[1;31mError: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, 'regsubst' parameter 'target' expects a value of type Array or String, got Undef (file: /etc/puppetlabs/code/modules/apache/manifests/mod.pp, line: 55, column: 19) (file: /etc/puppetlabs/code/modules/apache/manifests/mod/ldap.pp, line: 16) on node *******[0m`


manifests/params.pp
```diff
@@ -229,7 +229,7 @@
       'intercept_form_submit' => 'mod_intercept_form_submit',
       'ldap'                  => $::apache::version::distrelease ? {
         '7'     => 'mod_ldap',
-        default => undef,
+        default => 'mod_ldap',
```
smortex: This makes the selector useless (all path lead to `'ldap' => 'mod_ldap'`), it should be removed then.

turnopil: Yes, you are right

turnopil:
Do you need some additional ticker for this PR to be merged or some explanations?

mod_authz_ldap includes mod_ldap https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/authnz_ldap.pp#L7 but with the undefined package name:
- here https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/ldap.pp#L3
- and here https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/params.pp#L235

But someone made check for "_mod24_*_" package names pattern for Amazon and it breaks mod_authz_ldap installation because there is no chance to define package name for included mod_ldap:
- this check https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod.pp#L55

What we can do?
-  Remove include  _'::apache::mod::ldap'_ from mod_authz_ldap
- Add ldap package_name variable into mod_authz_ldap and made reliable inherit of '::apache::mod::ldap' to be possible define package name
- Add default package name for mod_ldap

I choose the last task because it looks the most simpler for me.

sheenaajay:
@turnopil.Thanks for your contribution.
Could you please resolve the conflicts.

turnopil:
The conflict has been resolved.

PR #1931, opened 2019-06-19
Added unmanaged_path and custom_fragment options to userdir

GeorgeCox:
The unmanged_path option stops the template from configuring the directory permissions in userdir.conf, this is useful if you want to have path set as 'public_html' or 'https://mysite/*' etc.

The custom_fragment option allows you to add your own config to userdir.conf, useful when using the unmanaged_path option.

### DIFF
--- a/manifests/mod/userdir.pp
+++ b/manifests/mod/userdir.pp
@@ -21,6 +21,12 @@
 # 
 # @param options
 #   Configures what features are available in a particular directory.
+#
+# @param unmanaged_path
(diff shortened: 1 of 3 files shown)


florindragos:
Hi @GeorgeCox 
The new options look fine but could you please add some tests with `unmanaged_path=true` ?
Thanks!

GeorgeCox:
Hi @florindragos, I've added some tests for unmanaged_path and custom_fragment.
Cheers, George

PR #1934, opened 2019-07-26
(FM-8214) converted to use litmus

tphoney:


tphoney:
closing because of #1938

PR #1935, opened 2019-08-05
(MODULES-9668 ) Please make ProxyRequests setting in vhost.pp configurable

aukesj:
This fork and change makes 'ProxyRequests' in Apache vhost sections managed by puppetlabs-apache configurable. I created a Jira ticket as well (MODULES-9668)

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -4657,7 +4657,7 @@ The following parameters are available in the `apache::mod::proxy` class.
 
 ##### `proxy_requests`
 
-Data type: `Any`
+Data type: `Boolean`
 
 Enables forward (standard) proxy requests.
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -7369,6 +7369,16 @@ X-Forwarded-Host and X-Forwarded-Server) get sent to the backend server.
 
 Default value: `undef`
 
+##### `proxy_requests`
+
+Data type: `Any`
+
+Sets the [ProxyErrorOverride Directive](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxyrequests).
+This prarmeter allows or prevents Apache httpd from functioning as a forward proxy server.
+(Setting ProxyRequests to Off does not disable use of the ProxyPass directive.)
+
+Default value: `false`
+
 ##### `proxy_error_override`
 
 Data type: `Any`
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -7373,7 +7373,7 @@ Default value: `undef`
 
 Data type: `Boolean`
 
-Sets the [ProxyErrorOverride Directive](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxyrequests).
+Sets the [ProxyRequests Directive](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxyrequests).
 This prarmeter allows or prevents Apache httpd from functioning as a forward proxy server.
 (Setting ProxyRequests to Off does not disable use of the ProxyPass directive.)
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -7371,7 +7371,7 @@ Default value: `undef`
 
 ##### `proxy_requests`
 
-Data type: `Any`
+Data type: `Boolean`
 
 Sets the [ProxyErrorOverride Directive](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxyrequests).
 This prarmeter allows or prevents Apache httpd from functioning as a forward proxy server.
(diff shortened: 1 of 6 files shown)


manifests/vhost.pp
```diff
@@ -1705,6 +1705,7 @@
   $proxy_dest_reverse_match                                                         = undef,
   $proxy_pass                                                                       = undef,
   $proxy_pass_match                                                                 = undef,
+  $proxy_requests                                                                   = false,
```
ekohl: Perhaps this should be of the datatype `Boolean`?

aukesj: Thanks for the remark. I've added 'Optional[Boolean]' as data type for proxy_requests variable.

ekohl: Why `Optional`? There's a default (`false`) so there's no way to make it `undef`.

aukesj: Okay, fixed that. Thanks for your patience

ekohl: No worries :) For my understanding (and to more effectively do reviews in the future), what made you choose `Optional`? I've seen others choose `Optional[Boolean]` because they thought it meant the value could be `false` then. Was it that or something else?

aukesj: Well, the way in which the template handles the boolean, if it's not set; the value will become 'false', so I figured setting the boolean would be optional. It's kinda new to me.

ekohl: Ok, good to know. In general I try to enforce clean data types as input. The template can be more robust than the actual input.

aukesj: I see. So concerning this pull request, is there anything you're expecting from me before this feature could go into the main module?

ekohl: I'm not part of the puppetlabs org, just a community member so I can't actually say :) It does look like you should regenerate the REFERENCE.md file so it has the correct data type.

aukesj: Fixed that

REFERENCE.md
```diff
@@ -4657,7 +4657,7 @@ The following parameters are available in the `apache::mod::proxy` class.
 
 ##### `proxy_requests`
 
-Data type: `Any`
+Data type: `Boolean`
```
ekohl: This looks like the wrong place. Note you can regenerate this file using `bundle exec rake strings:generate:reference`

REFERENCE.md
```diff
@@ -7369,6 +7369,16 @@ X-Forwarded-Host and X-Forwarded-Server) get sent to the backend server.
 
 Default value: `undef`
 
+##### `proxy_requests`
+
+Data type: `Any`
```
ThoughtCrhyme: You've added `proxy_requests` as a boolean here https://github.com/puppetlabs/puppetlabs-apache/pull/1935/files#diff-c918e8884948c0fae251b33071374150R1708 but document it it as `Data type: 'Ant'` here. Please correct this, thank you.

REFERENCE.md
ThoughtCrhyme: @aukesj please remove this duplicate section

aukesj:
I ran the tests up to 'bundle exec rspec spec/acceptance' (which fails locally with these errors (12 times):

`An error occurred while loading ./spec/acceptance/vhosts_spec.rb.
Failure/Error: require 'beaker-rspec/spec_helper'
RuntimeError:
  Failed to exec 'vagrant destroy --force'. Error was /usr/lib/ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem vagrant (>= 0.a) with executable vagrant (Gem::GemNotFoundException)
  	from /usr/lib/ruby/2.5.0/rubygems.rb:308:in `activate_bin_path'
  	from /usr/local/bin/vagrant:23:in `<main>'`

The Unit tests succeeded however (`bundle exec rake spec`):

`Finished in 44 minutes 44 seconds (files took 3 minutes 8 seconds to load)
7439 examples, 0 failures`

aukesj:
Furthermore, I'm using the fork succesfully with the option proxy_requests enabled.

REFERENCE.md
```diff
@@ -7369,6 +7369,16 @@ X-Forwarded-Host and X-Forwarded-Server) get sent to the backend server.
 
 Default value: `undef`
 
+##### `proxy_requests`
+
+Data type: `Boolean`
+
+Sets the [ProxyErrorOverride Directive](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxyrequests).
```
ThoughtCrhyme: The link text is incorrect. Should be `[ProxyRequests Directive]`

aukesj: Thanks, fixed that!

PR #1939, opened 2019-08-28
Add shibboleth support for Debian 10

fabbks:
Debian renamed the shib library to mod_shib.so in Debian 10 (buster) so mod shibboleth won't load anymore. This PR fixes that.


fabbks:
I'm not an Ubuntu user but this seems to be true. I changed the condition to make it Ubuntu and Debian version specific.

sheenaajay:
Thanks for submitting the PR @fabbks.Could you please rebase with the latest master.Thank you.
Screen Shot 2019-09-13 at 11 30 23

fabbks:
Done!

PR #1941, opened 2019-08-30
FM-8140 Add Redhat 8 support

sheenaajay:
Screen Shot 2019-09-04 at 10 28 58
Screen Shot 2019-09-02 at 15 09 47
Request for a review
Screen Shot 2019-08-30 at 13 04 35

### DIFF
--- a/spec/classes/mod/security_spec.rb
+++ b/spec/classes/mod/security_spec.rb
@@ -24,10 +24,17 @@ describe 'apache::mod::security', type: :class do
             )
           }
           it { is_expected.to contain_package('mod_security_crs') }
-          if facts[:os]['release']['major'].to_i > 6
+          if facts[:os]['release']['major'].to_i > 6 && facts[:os]['release']['major'].to_i <= 7
             it {
               is_expected.to contain_file('security.conf').with(
                 path: '/etc/httpd/conf.modules.d/security.conf',
+              )
+            }
+          end
+          if facts[:os]['release']['major'].to_i >= 8
+            it {
+              is_expected.to contain_file('security.conf').with(
+                path: '/etc/httpd/conf.d/security.conf',
               )
             }
           end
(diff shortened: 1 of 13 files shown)


spec/classes/mod/security_spec.rb
```diff
@@ -24,13 +24,20 @@
             )
           }
           it { is_expected.to contain_package('mod_security_crs') }
-          if facts[:os]['release']['major'].to_i > 6
+          if facts[:os]['release']['major'].to_i > 6 && facts[:os]['release']['major'].to_i <= 7
```
florindragos: maybe just `facts[:os]['release']['major'].to_i == 7` ?

sheenaajay: yeah we need to test 7.2 also hence added this condition(It failed on couple of OSes since release major version was 7.2)

spec/spec_helper_acceptance_local.rb
```diff
@@ -30,6 +30,12 @@ class { 'epel':
           epel_baseurl => "http://osmirror.delivery.puppetlabs.net/epel${::operatingsystemmajrelease}-\\$basearch/RPMS.all",
           epel_mirrorlist => "http://osmirror.delivery.puppetlabs.net/epel${::operatingsystemmajrelease}-\\$basearch/RPMS.all",
         }
+        } elsif $::operatingsystemmajrelease == '8' {
+          class { 'epel':
+                os_maj_release => "7",
```
tphoney: do we need epel for this ? there is no epel for redhat 8 yet.

spec/classes/mod/security_spec.rb
tphoney: we can do a compare using the gem version functionality. making it a single boolean check
``` if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.2.2')```

PR #1942, opened 2019-09-04
Add owner, group, file_mode and show_diff to apache::custom_config

treydock:
We use `apache::custom_config` for files that contain secrets like mod_auth_openidc so this has become necessary and would be ideal not to have to do this:

```puppet
    ::apache::custom_config { 'auth_openidc':
      content        => template('openondemand/apache/auth_openidc.conf.erb'),
      priority       => false,
    }
    # Hack to set mode of auth_openidc.conf
    File <| title == 'apache_auth_openidc' |> {
      owner     => 'root',
      group     => 'apache',
      mode      => '0640',
      show_diff => false,
    }
```

### DIFF
--- a/manifests/custom_config.pp
+++ b/manifests/custom_config.pp
@@ -36,6 +36,18 @@
 # @param verify_config
 #   Specifies whether to validate the configuration file before notifying the Apache service.
 #
+# @param owner
+#   File owner of configuration file
+#
+# @param group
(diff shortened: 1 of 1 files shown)


sheenaajay:
Thanks for submitting the PR @treydock

sheenaajay:
Screen Shot 2019-09-18 at 14 21 35

PR #1943, opened 2019-09-04
Set package name for proxy_html and xml2enc modules

treydock:


sheenaajay:
@treydock Thanks for submitting the PR. Could you please provide the details for the PR.Is it currently breaking without the changes. Thanks again.

carabasdaniel:
Hello @treydock,

Sadly this functionality is already covered by the apache::mod type and adding the installation of a package at this step could cause issues further down the line on different operating systems. 

Would you be ok closing this PR and adding the work-around you mentioned in the module documentation ?

treydock:
The way this pull request works it won't change anything for operating systems besides RHEL and RHEL with SCL. This makes it so the provided `apache::mode::*` classes are functionally complete. I could update further to ensure necessary packages are installed for all operating systems but as it stands this pull request just fixes missing behavior. Seems like documenting a work around is putting the burden of how to configure Apache on the administrator when this module could handle things.

ekohl:
Shouldn't it introduce a `apache::mod::xml2enc` and include that where it now instantiates `apache::mod { 'xml2enc': }`? Then you can use hiera to override as well.

michaeltlombardi:
Since this hasn't seen any activity in about four months, I'm going to go ahead and close this pull request for the time being. Please re-open this pull request once the next actions are addressed, new information is available, or you have a question related to this pull request. We've become aware of difficulties re-opening pull requests, in the event you cannot please mention `puppetlabs/modules`  with an @ sign in front and we'll re-open this pull request.

Closing the pull request doesn't mean we don't consider this change valuable, just that there are things that need to be addressed before it can be merged. If you have any questions or concerns, please don't hesitate to ping us in slack.puppet.com.

PR #1947, opened 2019-09-10
add possibility to define variables inside VirtualHost definition

trefzer:
The variable can them be used inside the vhost tag as a variable.
At the footer, the variable is undefined to avoid side effects
in other vhosts.

also see https://httpd.apache.org/docs/2.4/mod/core.html#define


tphoney:
A fair point well made, i have to ask :)

Thanks for the great PR, have a good one !

PR #1953, opened 2019-09-20
Add parameter version for mod security

tuxmea:
We want to be able to select mod security 1 or 2.
Depending on Version, package names and config files are different.

Fixes: https://tickets.puppetlabs.com/browse/MODULES-9947


carabasdaniel:
Hello @tuxmea,

Thank you for your contribution.

PR #1957, opened 2019-10-10
[WIP] Add vhost::directories validation and include mods

ekohl:
This commit adds data type validation on $directories and uses this to include mods when needed. This prevents misconfigurations by avoiding the need to manually add include statements.

Currently it's incomplete. Only about half of the parameters are present and others may be incorrect. Before I complete this I'd like feedback if this is desired. It has a great chance of breaking setups, but there's a good chance that those setups are already incorrect. By breaking during manifest compliation, users can find this out in early on.


types/vhost/directory.pp
```diff
@@ -0,0 +1,68 @@
...
+    'provider'                                            => Optional[String],
+    'headers'                                             => Optional[Variant[String[1], Array[String[1]]]],
+    'geoip_enable'                                        => Optional[Apache::Bool],
+    'options'                                             => Optional[Variant[String[1], Array[String[1]]]],
+    'index_options'                                       => Optional[Variant[String[1], Array[String[1]]]],
+    'index_order_default'                                 => Optional[Variant[String[1], Array[String[1]]]],
+    'index_stylesheet'                                    => Optional[String],
+    'allow_override'                                      => Optional[Variant[String[1], Array[String[1]]]],
+    'satify'                                              => Optional[String],
+    'limit'                                               => Any, # TODO: complex type
```
ekohl: I'm wondering if I should take this approach with all keys: just start with listing them all with the Any type. Tighten validation from there.

ekohl:
Closing since I don't have time to finish this.

PR #1958, opened 2019-10-10
always quote ExpiresDefault in vhost::directories

evgeni:
this was already done in expires.conf, but not in a vhost

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -320,7 +320,7 @@
     ExpiresActive <%= directory['expires_active'] %>
     <%- end -%>
     <%- if directory['expires_default'] -%>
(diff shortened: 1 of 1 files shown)


evgeni:
unrelated test failure is unrelated :woman_shrugging:

ekohl:
This would be breaking since users might already work around this by adding quotes.

michaeltlombardi:
Investigating the weird syntax check fail, may need to add a maint commit and have you rebase once I fix it.

michaeltlombardi:
Fixed the problem and merged in at #1959, tests are now green.

evgeni:
Looks like tests ran again and passed now, so I don't need to rebase, right?

PR #1961, opened 2019-10-19
(MODULES-9948) Allow switching of thread modules

tuxmea:
On Debian based systems (I have not tested behavior on other OS) you can have multiple thread modules installed in parallel. An admin must ensure that only one thread module is active by disabling any old thread module when adding a new one.

Changing thread module was not possible in Puppet Apache Module.
Once a thread module was active it never got disabled when switching to another one, causing the apache process to not be able to start.

This PR adds the capability to switch thread modules on existing, Puppet managed apache installations on Debian based systems.

Usual thread modules are: prefork, event or worker.
When switching from one thread module to another we must ensure that the
old, no longer to use, thread module gets disabled.

Puppet Apache module will always use the new mpm base thread modules and place
them in files without mpm_ prefix.
The a2dismod utility only checks for file names, not the real thread
module.

### DIFF
--- a/manifests/mpm.pp
+++ b/manifests/mpm.pp
@@ -109,6 +109,7 @@ define apache::mpm (
       if $mpm == 'worker' {
         if ( ( $::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease,'18.04') >= 0 ) or ( $::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '8.0.0') >= 0 ) ) {
           include apache::mpm::disable_mpm_event
+          include apache::mpm::disable_mpm_prefork
         }
       }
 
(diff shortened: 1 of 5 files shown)


david22swan:
@tuxmea Could you add some additional documentation to cover this change please?

tuxmea:
@david22swan I have updated the PR description.

david22swan:
@tuxmea The updated readme is great thanks, just wanted a bit more info before I made my decision. Anyway the PR looks good to me but will have to wait a bit on a merge as the pipeline is currently failing.

tuxmea:
@david22swan any idea, when pipeline is working again?

david22swan:
@tuxmea Thanks for the pr, the pipeline wasn't totally green but enough so your changes looked like they wouldn't effect it negatively.
Look forward to seeing more from you in the future. :)

PR #1965, opened 2019-10-31
(FM-8391) Update README per team practices

michaeltlombardi:
This commit updates the readme to remove the language marking litmus
as an experimental tool as well as clarifying the instructions for
running acceptance tests and noting the schedule for the IA Content
office hours.

### DIFF
--- a/README.md
+++ b/README.md
@@ -915,8 +915,6 @@ The [`apache::mod::suphp`][] class is untested since repositories are missing co
 
 ### Testing
 
-Due to the difficult and specialised nature of acceptance testing mods in apache IE (high OS specificity), we have replaced  acceptance tests with unit tests.
-
 To run the unit tests, install the necessary gems:
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -935,10 +933,19 @@ To check the code coverage, run:
...
 
-We want to make it as easy as possible to contribute changes so our modules work in your environment, but we also need contributors to follow a few guidelines to help us maintain and improve the modules' quality.
+If you have problems getting this module up and running, please [contact Support](http://puppetlabs.com/services/customer-support).
 
-For more information, please read the complete [module contribution guide][] and check out [CONTRIBUTING.md][].
\ No newline at end of file
+If you submit a change to this module, be sure to regenerate the reference documentation as follows:
+
+```bash
+puppet strings generate --format markdown --out REFERENCE.md
```
ekohl: There's also a rake task provided by puppet-strings. `bundle exec rake strings:generate:reference` should IMHO be encouraged. I think that works as long as puppet-strings is available as a gem and a recent(ish) puppetlabs_spec_helper is used.

README.md
```diff
@@ -935,10 +933,19 @@ To check the code coverage, run:
 COVERAGE=yes bundle exec rake parallel_spec
```
ekohl: Shouldn't `rake check` be the recommended approach since it also runs all the linting? It's provided by `puppetlabs_spec_helper`.

michaeltlombardi:
@ekohl I suppose I should've waited til I saw your feedback to submit a dozen of these... 😭

ekohl:
I am wondering if this should be in `CONTRIBUTING.md`. https://github.com/puppetlabs/puppetlabs-apache/blob/master/CONTRIBUTING.md#testing looks a bit out of date.

michaeltlombardi:
Ugh, to be honest, we probably need to do a proper full-on docs overhaul for the modules at some point in the near future.

ekohl:
Makes sense. In Voxpupuli we just modulesync since contributing is generic for all repos in the org. Not sure if the same can be done when using generic PDK templates though.

michaeltlombardi:
Possibly, but I don't think pdksync _currently_ manages the readme. I think the templates probably only make sense on an organizational level, too. 🤔

README.md
michaeltlombardi: I'm going to leave this one alone for now as out of scope for this current change set, but it'll need a proper tackling when we come back through.

README.md
michaeltlombardi: This will require the removal of the Japanese translations, also putting it out of scope for this round of changes, I think, until that's properly removed.

ekohl:
> Possibly, but I don't think pdksync _currently_ manages the readme

README should be unique for every repository so I don't think you ever want this.

> I think the templates probably only make sense on an organizational level, too. thinking

Indeed, `CONTRIBUTING.md` often only makes sense within an organization.

lionce:
@ekohl all your remarks are correct. CONTRIBUTING.md should be updated. Read me also needs to be revised. We'll take them into consideration and come back with solutions. Thanks! 
 
This PR has just updated the file with the latest changes about testing and some extra details about office hours, the github url for the project and helper / training videos 

It looks good to me 👍

PR #1966, opened 2019-11-01
(maint) FM-8656 minor fix for sles

sheenaajay:
Screen Shot 2019-11-01 at 11 26 49


sheenaajay:
Screen Shot 2019-11-01 at 11 46 10

PR #1805, opened 2018-07-06
Add support for ProxyTimeout

agoodno:
Separately changes the default for Timeout to the Apache default per this discussion. https://tickets.puppetlabs.com/browse/MODULES-5792.

PR #1975, opened 2019-11-08
Change mod_proxy's ProxyTimeout to follow Apache's global timeout

gcoxmoz:
[apache docs](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxytimeout) for ProxyTimeout indicate that the default timeout tracks the global Timeout.  The code introduced in #1805 for `apache::mod::proxy` codes in '60' seconds for `proxy_timeout`.  While 60 is the normal default for the global `timeout` 'out of the box', it means that 60 becomes hard-specified and no longer tracks the global timeout.  This leads to a surprise when $apache::timeout is raised to 300 and proxied requests bomb out at 60.  Not that I know anyone who went through that, no ma'am.

The change edits the default of ProxyTimeout to track along with the global timeout unless told otherwise.

- Without a global `timeout` set, the `proxy_timeout` parameter will default to the global (undef) and lead to the `ProxyTimeout` config line disappearing. (minor config file change, no net apache behavior change).

- With a global `timeout` parameter, `proxy_timeout` will track to the global number instead of being 60 (my desired change.  config files and operations will change to line up with apache docs, under 'principle of least surprise').

- Specifying different numbers between `timeout` and `proxy_timeout` is still possible, and behavior is unchanged from the current world.

### DIFF
--- a/manifests/mod/proxy.pp
+++ b/manifests/mod/proxy.pp
@@ -27,9 +27,10 @@ class apache::mod::proxy (
   $apache_version = undef,
   $package_name   = undef,
   $proxy_via      = 'On',
-  $proxy_timeout  = $apache::timeout,
+  $proxy_timeout  = undef,
 ) {
   include ::apache
+  $_proxy_timeout = $apache::timeout
   $_apache_version = pick($apache_version, $apache::apache_version)
   ::apache::mod { 'proxy':
     package => $package_name,
--- a/manifests/mod/proxy.pp
+++ b/manifests/mod/proxy.pp
@@ -27,7 +27,7 @@ class apache::mod::proxy (
   $apache_version = undef,
   $package_name   = undef,
   $proxy_via      = 'On',
-  $proxy_timeout  = '60',
+  $proxy_timeout  = $::apache::timeout,
 ) {
   include ::apache
   $_apache_version = pick($apache_version, $apache::apache_version)
--- a/manifests/mod/proxy.pp
+++ b/manifests/mod/proxy.pp
@@ -27,7 +27,7 @@ class apache::mod::proxy (
   $apache_version = undef,
   $package_name   = undef,
   $proxy_via      = 'On',
-  $proxy_timeout  = $::apache::timeout,
+  $proxy_timeout  = $apache::timeout,
 ) {
   include ::apache
   $_apache_version = pick($apache_version, $apache::apache_version)
--- a/templates/mod/proxy.conf.erb
+++ b/templates/mod/proxy.conf.erb
@@ -25,5 +25,7 @@
   # Set to one of: Off | On | Full | Block
   ProxyVia <%= @proxy_via %>
 
+  <%- if @proxy_timeout -%>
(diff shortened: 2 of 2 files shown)


gcoxmoz:
Testing looks to fail because we're now referring to an 'outside' variable.  Looking at `apache::mod::authz_default` it seems like that's okay.  `apache::mod::alias` inherits from `apache::params`.  I'm unclear about which way the module should go, so I'm not sure how to clear this test up.

manifests/mod/proxy.pp
```diff
@@ -27,7 +27,7 @@
   $apache_version = undef,
   $package_name   = undef,
   $proxy_via      = 'On',
-  $proxy_timeout  = '60',
+  $proxy_timeout  = $::apache::timeout,
```
sheenaajay: @gcoxmoz could you please try updating with following
$proxy_timeout  = ::apache::timeout
We tried and unit test cases ran clean. Now running the full acceptance suite to see if the test cases are running clean. Will let you know. Thank you for submitting the PR.

sheenaajay: Screen Shot 2019-11-15 at 10 40 07

sheenaajay: Screen Shot 2019-11-20 at 09 38 46
The fix for redhat is available in master. If you could rebase your branch, we can get the PR run successfully on redhat8 too. Please let us know if you need more information.

gcoxmoz: Messily updated per your suggestions (sorry, please squash if accepted).
I have no issue if you want to make edits in order to accept.  You have a much better handle on the actual code than I do.  Thanks!

sheenaajay:
@gcoxmoz Thank you. Will rerun the tests and merge the PR.

manifests/mod/proxy.pp
```diff
@@ -27,9 +27,10 @@
   $apache_version = undef,
   $package_name   = undef,
   $proxy_via      = 'On',
-  $proxy_timeout  = '60',
+  $proxy_timeout  = undef,
 ) {
   include ::apache
+  $_proxy_timeout = $apache::timeout
```
alexjfisher, line 33: What is this set for??

gcoxmoz: I don't follow what specifically you're asking.  The rationale for this patch is in the main conversation.

alexjfisher: That variable declaration is completely unused?  It doesn't do anything.  Or am I missing something? The change otherwise looks correct, (it makes sense to default to `undef` and allow Apache's main `timeout` setting to be used).

Was just stumped as to why this particular line was asked for.  It still exists in the latest version.  I might create a PR to remove it.

gcoxmoz: skim skim skim ...  huh.  Yeah, that's weird.
I'm guessing that testing had an intermittent and they lucked into passing when this was added, and I didn't know enough back then to question it.  Yeah, feels like "zap it" (if err-force will even look at a community PR)

PR #1976, opened 2019-11-12
move unless into manage_security_corerules

SimonHoenscheid:
This PR moves the unless block inside the manage manage_security_corerules block because it is only needed if core rules are there. It also prevents the apache::security::rule_link define to be called if security core rules are not managed

### DIFF
--- a/manifests/mod/security.pp
+++ b/manifests/mod/security.pp
@@ -229,10 +229,10 @@ class apache::mod::security (
       require => File[$modsec_dir],
       notify  => Class['apache::service'],
     }
-  }
 
-  # Debian 9 has a different rule setup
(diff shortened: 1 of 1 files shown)


florindragos:
Hello @SimonHoenscheid,
Thanks for your contribution!
Can you provide some more info on why this is necessary?
From what I can see, the unless has been outside manage_security_corerules for a long time.

tuxmea:
@florindragos : we recommend to only activate the security rules when security rules are set at all.
The actual code will try to set rules, even when `$manage_security_crs` is set to false.

SimonHoenscheid:
@florindragos I updated the description to make the intention more clear

florindragos:
got it! makes sense, was just wondering why it was not fixed earlier.

PR #1979, opened 2019-11-29
When using mod jk, we expect the libapache2-mod-jk package to be installed

tuxmea:


sheenaajay:
@tuxmea Thanks for submitting the PR.
Ran adhoc job and it ran successfully on all nodes.
pid 36260 exit 0
Successful on 20 nodes:

PR #1983, opened 2019-12-09
Add Github Actions release workflow

lionce:
add github actions to master

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,52 @@
+name: "release"
+
+on: 
+  push:
+    branches:
+      - 'release'
+
+jobs:
+  LitmusAcceptance:
+
+    runs-on: self-hosted
+
+    strategy:
+      matrix:
+        ruby_version: [2.5.x]
+        puppet_gem_version: [~> 6.0]
+        platform: [release_checks]
+        agent_family: ['puppet5', 'puppet6']
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Litmus Parallel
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+      with:
+        platform: ${{ matrix.platform }}
+        agent_family: ${{ matrix.agent_family }}
+
+  Spec:
+    runs-on: self-hosted
+
+    strategy:
+      matrix:
+        check: [parallel_spec, 'syntax lint metadata_lint check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop']
```
ekohl: I noticed the same thing was used in Travis. Have you considered a proper task in [puppetlabs_spec_helper](https://github.com/puppetlabs/puppetlabs_spec_helper) to capture these? Perhaps just a `check:all` to capture all the `check:*` tasks there would help.

Also note that `syntax metadata_lint` should be [validate](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/4fb572b8eab2662b751f32b522f5f8f2c99c8be7/lib/puppetlabs_spec_helper/rake_tasks.rb#L231-L245) to also check `.rb` files.

carabasdaniel: @ekohl Thanks for the great suggestion, we will change these in the near future, maybe removing entirely the unit tests from this workflow as those are run already on Travis, however this PR is meant for synchronisation of our release branch with the master branch.

ekohl: My bad, didn't see it. Still not entirely used to having a release branch.

PR #1984, opened 2019-12-10
Fix newline being added before proxy params

oxc:
When a proxy_pass with additional params is defined within a directories location entry, the params get appended in a new line (which breaks config).
The change in this pull request removes the superfluous newline.


oxc:
The failed tests seem unrelated to my changes.

sheenaajay:
Thanks @oxc for submitting the PR.
Screen Shot 2019-12-11 at 11 49 24

PR #1990, opened 2020-01-09
(MODULES-10391) ssl_protocol includes SSLv2 and SSLv3 on all platforms

legooolas:
(Reported as MODULES-10391)

The default for the `apache::mod::ssl_protocol` parameter was changed to allow SSLv2 and SSLv3 on all platforms, which isn't desirable on anything with an older version of OpenSSL.  Noticed on CentOS 7 hosts after upgrading this module, without changing any parameters.

Hence this PR is to move the default setting to the params class and only set it for RHEL 8, which appears to be in the intention of the prior change in FM-8721 (PR #1691 

This then fixes this regression on other platforms.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -791,4 +791,10 @@ class apache::params inherits ::apache::version {
   } else {
     $verify_command = '/usr/sbin/apachectl -t'
   }
+
+  if ($::operatingsystem == 'RedHat' and ($facts['operatingsystemmajrelease'] == '8') ) {
+    $ssl_protocol = ['all'] # Implementations of the SSLv2 and SSLv3 protocol versions have been removed from OpenSSL (and hence mod_ssl) because these are no longer considered secure. For additional documentation https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/setting-apache-web-server_deploying-different-types-of-servers
+  } else {
+    $ssl_protocol = [ 'all', '-SSLv2', '-SSLv3' ]
+  }
 }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -792,7 +792,7 @@ class apache::params inherits ::apache::version {
     $verify_command = '/usr/sbin/apachectl -t'
   }
 
-  if ($::operatingsystem == 'RedHat' and ($facts['operatingsystemmajrelease'] == '8') ) {
+  if ($::osfamily == 'RedHat' and ($facts['operatingsystemmajrelease'] >= '8') ) {
     $ssl_protocol = ['all'] # Implementations of the SSLv2 and SSLv3 protocol versions have been removed from OpenSSL (and hence mod_ssl) because these are no longer considered secure. For additional documentation https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/setting-apache-web-server_deploying-different-types-of-servers
   } else {
     $ssl_protocol = [ 'all', '-SSLv2', '-SSLv3' ]
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -792,7 +792,7 @@ class apache::params inherits ::apache::version {
     $verify_command = '/usr/sbin/apachectl -t'
   }
 
-  if ($facts['os']['family'] == 'RedHat' and ($facts['os']['release']['major'] >= '8') ) {
+  if $::osfamily == 'RedHat' and versioncmp($::operatingsystemrelease, '8.0') >= 0 {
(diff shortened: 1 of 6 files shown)


manifests/params.pp
```diff
@@ -791,4 +791,10 @@
   } else {
     $verify_command = '/usr/sbin/apachectl -t'
   }
+
+  if ($::operatingsystem == 'RedHat' and ($facts['operatingsystemmajrelease'] == '8') ) {
```
ekohl: Probably better to check the os family and if it's `>= 8`. That way you also apply it to CentOS and future versions.

legooolas: Good plan!  Updated to resolve this  :)

TwizzyDizzy:
Thanks for reporting this... I just discovered this [commit](
https://github.com/puppetlabs/puppetlabs-apache/commit/2926250d8c8e706436d0895b7db68a8e9cba8e0f) and am baffled that this kind of change could ever pass review with this kind of inconspicuous commit message. Also this should never have been committed as combined commit but as seperate commits to keep things atomic.

This also raises questions about quality control on side and I would indeed appreciate a statement how those kind of dangerous changes can be avoided in the future.

/cc ing @sheenaajay here!

manifests/params.pp
```diff
@@ -791,4 +791,10 @@
   } else {
     $verify_command = '/usr/sbin/apachectl -t'
   }
+
+  if ($::osfamily == 'RedHat' and ($facts['operatingsystemmajrelease'] >= '8') ) {
```
bastelfreak: since Puppet deprecated those facts, please use the new ones:
```suggestion
  if ($facts['os']['family'] == 'RedHat' and ($facts['os']['release']['major'] >= '8') ) {
```

sheenaajay:
@legooolas Thanks for submitting this PR. Apologise for this. Will review them and get it in asap.Thanks again.

sheenaajay:
@TwizzyDizzy Thanks for the valuable feedback. Apologies, Will get the changes reviewed and merged. Running them on Adhoc pipeline.

manifests/params.pp
legooolas: Updated  :)

(Was keeping it the same as the other fact uses in the params file -- they all need switching to the facts array really...)

manifests/params.pp
```diff
@@ -791,4 +791,10 @@
   } else {
     $verify_command = '/usr/sbin/apachectl -t'
   }
+
+  if ($facts['os']['family'] == 'RedHat' and ($facts['os']['release']['major'] >= '8') ) {
+    $ssl_protocol = ['all'] # Implementations of the SSLv2 and SSLv3 protocol versions have been removed from OpenSSL (and hence mod_ssl) because these are no longer considered secure. For additional documentation https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/setting-apache-web-server_deploying-different-types-of-servers
+  } else {
+    $ssl_protocol = [ 'all', '-SSLv2', '-SSLv3' ]
+  }
```
runejuhl: Since this module has a dependency on Puppet `>= 5.5.10` I'd love for new parameters to be added to Hiera files instead of using `params.pp`.

I'm not sure what your policy is, so it might be out of the question, but it seems a bit silly to continue using the deprecated `params.pp` style defaults when you can use Hiera instead. I'd even prefer to keep `params.pp` around (until parameters can be migrated to Hiera), and use Hiera for any new parameters, even if it means that both systems will be used for a while.

runejuhl: ...and as per https://puppet.com/docs/puppet/latest/hiera_migrate.html#adding_hiera_data_to_a_module using Hiera is the preferred way:

> Modules need default values for their class parameters. Before, the preferred way to do this was the “params.pp” pattern. With Hiera 5, you can use the “data in modules” approach instead.

sheenaajay:
@legooolas Apologies for the regression.
Thanks for reporting the error and opening the PR.
Will update the PR with the fix for the Travis failures and will update the tests.Thank you.

manifests/params.pp
legooolas: Again this is just to match the existing manifests, and updating to use hiera module data should probably be a separate PR to change this for all params at once.

runejuhl: Yes, I understand that. I'd simply prefer to start moving parameters to Hiera *now*, instead of doing it at `Optional[String] $future_date` -- Hiera is just so much easier to comprehend and maintain than `params.pp` :smile:

sheenaajay:
@legooolas Following changes are committed to this PR.
1. Use versioncmp($::operatingsystemrelease, '8.0') to compare the versions
2. Added unit tests to verify the default values of ssl_protocol
3. Added acceptance tests to verify the default values of sll_protocol wrt different OS.
4.Travis is running green 
5.Ran release checks on this branch too
Screen Shot 2020-01-20 at 13 39 01

PR #1994, opened 2020-01-24
Fix proxy_html Module to work on Debian 10

buchstabensalat:
In https://github.com/puppetlabs/puppetlabs-apache/commit/8e7ef1550cc8a4f816a0f4d7235df36b48664998 a fix was made to support Ubuntu 10.04, where libxml2 is at another location.
But since the "case"-clause is using $::osfamily as parameter it now matches on Debian 10 too.

This fix implements the distinction between Debian and Ubuntu systems.


cmccrisken-puppet:
Thanks for the fix @buchstabensalat

PR #2003, opened 2020-02-20
pdksync - Add weekly scheduled workflows

carabasdaniel:
Add weekly scheduled workflows
pdk version: `1.16.0`

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -1,0 +1,42 @@
+name: "weekly"
+
+on:
+  schedule:
+    - cron: '0 0 * * 6'
+
+jobs:
+  LitmusAcceptance:
+    runs-on: self-hosted
+    strategy:
+      matrix:
+        ruby_version: [2.5.x]
+        puppet_gem_version: [~> 6.0]
+        platform: [release_checks]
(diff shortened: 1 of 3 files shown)


carabasdaniel:
Hi @ekohl 

With Github actions self-hosted runners, as we have set up, only the default branch of the repository will be used in the scheduled triggered build.

ekohl:
But if someone forks this repository (or updates their default branch), will their fork also attempt to run it?

carabasdaniel:
This workflow is using the self-hosted runner and that is tied only to this repository so it should not attempt to run it.

ekohl:
That doesn't really answer my question. Any user or organization can have self hosted runners.

carabasdaniel:
Workflows don't run on a forked repository by default. See the details for scheduled workflows here: https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule

ekohl:
I didn't get that from those docs, which is why I was still unsure about it. It's good to know because in theforeman we're also considering switching over. In Voxpupuli we do have a few modules that are forks and Puppetlabs does too (https://github.com/puppetlabs/puppetlabs-apt)

PR #2005, opened 2020-02-21
pdksync - Test pinning to concurrent-ruby 1.1.5

DavidS:
Test pinning to concurrent-ruby 1.1.5
pdk version: `1.16.0`


DavidS:
puppet-module-gems 0.4.4 was released with this change

PR #2013, opened 2020-03-18
(maint) Lock facter version < 4

florindragos:


florindragos:
@DavidS There are 4 unit tests failing when running with facter 4. All of them are using facterdb and failing on Gentoo. Looks like a custom fact from the puppet-portage module is not being resolved. We're not sure yet where the problem is, so it's probably best to keep using facter 2 until we find and fix the issue.

DavidS:
@florindragos  Do you have a ticket for that work that I can subscribe?

florindragos:
@DavidS https://tickets.puppetlabs.com/browse/FACT-2478

DavidS:
thanks!

PR #2014, opened 2020-03-19
Add ProxyIOBufferSize option

jplindquist:
Add [ProxyIOBufferSize](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxyiobuffersize) for `mod_proxy`

Found we need to adjust this in combination w/ AJP to account for larger headers (cookies)

### DIFF
--- a/manifests/mod/proxy.pp
+++ b/manifests/mod/proxy.pp
@@ -19,6 +19,9 @@
 # @param proxy_timeout
 #   Network timeout for proxied requests.
 #
+# @param proxy_iobuffersize
+#   Set the size of internal data throughput buffer
+#
 # @see https://httpd.apache.org/docs/current/mod/mod_proxy.html for additional documentation.
 #
 class apache::mod::proxy (
@@ -28,7 +31,7 @@ class apache::mod::proxy (
   $package_name       = undef,
   $proxy_via          = 'On',
   $proxy_timeout      = undef,
-  $proxyiobuffersize  = undef,
+  $proxy_iobuffersize = undef,
 ) {
   include ::apache
   $_proxy_timeout = $apache::timeout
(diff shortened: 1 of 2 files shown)


manifests/mod/proxy.pp
```diff
@@ -1,33 +1,34 @@
...
 # @param proxy_via
 #   Set local IP address for outgoing proxy connections.
-# 
+#
 # @param proxy_timeout
 #   Network timeout for proxied requests.
-# 
+#
 # @see https://httpd.apache.org/docs/current/mod/mod_proxy.html for additional documentation.
 #
```
sheenaajay, line 23: Could you please add the new parameter information

manifests/mod/proxy.pp
```diff
@@ -1,33 +1,34 @@
...
-  $package_name   = undef,
-  $proxy_via      = 'On',
-  $proxy_timeout  = undef,
+  $proxy_requests     = 'Off',
+  $allow_from         = undef,
+  $apache_version     = undef,
+  $package_name       = undef,
+  $proxy_via          = 'On',
+  $proxy_timeout      = undef,
+  $proxyiobuffersize  = undef,
```
sheenaajay, line 31: minor suggestion, can we keep the same naming format for the variable $proxy_iobuffersize

sheenaajay:
Thanks @jplindquist for submitting the PR.

manifests/mod/proxy.pp
jplindquist: Renamed to `proxy_iobuffersize` as suggested, thanks!

manifests/mod/proxy.pp
jplindquist: Done

jplindquist:
@sheenaajay The tests had previously passed, and after making the changes requested, there were a couple of failures?  Is that an issue, or is there a way to retry those, or something else I need to do to get that passing?

sheenaajay:
Thank you @jplindquist for submitting the PR.

PR #2017, opened 2020-04-06
Changed the apache-mod-ldap.conf title to ldap.conf

amorphina:
Otherwise no symlink for the ldap.conf is created in /etc/apache2/mods-enabled as required by mod.pp. Thus the custom ldap options in the /etc/apache2/mods-available are not applied in the /etc/apache2/mods-enabled and are not loaded.


sheenaajay:
Thank you @amorphina for submitting the PR.

PR #2021, opened 2020-04-29
MODULES-10586 Centos 8: wrong package used to install mod_authnz_ldap

farebers:
see https://tickets.puppetlabs.com/browse/MODULES-10586

the change was tested on a fresh centos 8 install

note on unit tests: 

- I edited the tests 'default configuration with parameters on a RedHat OS' in authnz_ldap_spec.rb and ldap_spec.rb (updated the facts operatingsystemrelease line to '8') in the puppetlabs module (not containing my change) 
- so those tests should have failed but didn't 

so I was unable to provide a sensible unit test

### DIFF
--- a/spec/classes/mod/authnz_ldap_spec.rb
+++ b/spec/classes/mod/authnz_ldap_spec.rb
@@ -41,50 +41,53 @@ describe 'apache::mod::authnz_ldap', type: :class do
   end # Debian
 
   context 'default configuration with parameters on a RedHat OS' do
-    let :facts do
-      {
-        osfamily: 'RedHat',
-        operatingsystemrelease: '6',
-        id: 'root',
-        kernel: 'Linux',
-        operatingsystem: 'RedHat',
-        path: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
-        is_pe: false,
-      }
-    end
+    on_supported_os.each do |os, os_facts|
+      next unless os.start_with?('redhat')
+      context "On #{os}" do
+        let :facts do
+          os_facts
+        end
 
-    it { is_expected.to contain_class('apache::params') }
-    it { is_expected.to contain_class('apache::mod::ldap') }
-    it { is_expected.to contain_apache__mod('authnz_ldap') }
+        it { is_expected.to contain_class('apache::params') }
+        it { is_expected.to contain_class('apache::mod::ldap') }
+        it { is_expected.to contain_apache__mod('authnz_ldap') }
 
-    context 'default verify_server_cert' do
-      it { is_expected.to contain_file('authnz_ldap.conf').with_content(%r{^LDAPVerifyServerCert On$}) }
-    end
+        if os_facts[:operatingsystemmajrelease].to_i >= 7
+          it { is_expected.to contain_package('mod_ldap') }
+        else
+          it { is_expected.to contain_package('mod_authz_ldap') }
+        end
 
-    context 'verify_server_cert = false' do
-      let(:params) { { verify_server_cert: false } }
+        context 'default verify_server_cert' do
+          it { is_expected.to contain_file('authnz_ldap.conf').with_content(%r{^LDAPVerifyServerCert On$}) }
+        end
 
-      it { is_expected.to contain_file('authnz_ldap.conf').with_content(%r{^LDAPVerifyServerCert Off$}) }
-    end
+        context 'verify_server_cert = false' do
+          let(:params) { { verify_server_cert: false } }
 
-    context 'verify_server_cert = wrong' do
-      let(:params) { { verify_server_cert: 'wrong' } }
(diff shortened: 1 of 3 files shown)


sanfrancrisko:
Hi @farebers,

Firstly, thank-you for the fix and highlighting the gap in the test coverage to capture this. As you pointed out, the unit tests did not catch this failure. We can certainly extend `spec/classes/mod/authnz_ldap_spec.rb` to verify that we are using the correct package name on RHEL systems based on their major versions, but, unfortunately this will not catch the failure until the manifest is actually compiled on an OS. 

I think the best thing is to to add an acceptance test for this particular scenario in `spec/acceptance/mod_authnz_ldap_spec.rb`, something along the lines of this:

```ruby
require 'spec_helper_acceptance'
apache_hash = apache_settings_hash

describe 'apache::mod_authnz_ldap' do
  context 'Default mod_authnz_ldap module installation' do
    pp = <<-MANIFEST
      class { 'apache': }
      class { 'apache::mod::auth::authnz_ldap': }
    MANIFEST

    it 'succeeds in installing the mod_authnz_ldap module' do

      describe file("#{apache_hash['mod_dir']}/authnz_ldap.load") do
        it { is_expected.to contain 'mod_authnz_ldap.so' }
      end
    end
  end
end
```

This will ensure that the installation of the package works correctly.

I have been taking a look at how we could enhance the unit tests, doing something like this:
```ruby
on_supported_os.each do |os, facts|
      context "On #{os}" do
        let :facts do
          facts
        end

        it { is_expected.to contain_class('apache::params') }
        it { is_expected.to contain_class('apache::mod::ldap') }
        it { is_expected.to contain_apache__mod('authnz_ldap') }

        if facts[:os]['name'] == 'Redhat' and facts[:os]['version'].to_i < 7
          it { is_expected.to contain_package('mod_authz_ldap') }
        elsif facts[:os]['name'] == 'Redhat' and facts[:os]['version'].to_i > 6
          it { is_expected.to contain_package('mod_ldap')}
        end
```

...but there's a few issues I'm running in to with this approach.

I think the path of least resistance would be to add an acceptance test, similar to above. Let me know if I can be of any further help.

spec/classes/mod/authnz_ldap_spec.rb
```diff
@@ -41,50 +41,53 @@
...
-    end
-
-    context 'verify_server_cert = wrong' do
-      let(:params) { { verify_server_cert: 'wrong' } }
-
-      it 'raises an error' do
-        expect { is_expected.to raise_error Puppet::Error }
+    on_supported_os.each do |os, facts|
+      next unless os.start_with? 'redhat'
+      context "On #{os}" do
```
ekohl, line 46: I'm usually not a fan of `next`
```suggestion
      context "On #{os}", if: os.start_with?('redhat') do
```

spec/classes/mod/authnz_ldap_spec.rb
```diff
@@ -41,50 +41,53 @@
...
-
-      it { is_expected.to contain_file('authnz_ldap.conf').with_content(%r{^LDAPVerifyServerCert Off$}) }
-    end
-
-    context 'verify_server_cert = wrong' do
-      let(:params) { { verify_server_cert: 'wrong' } }
-
-      it 'raises an error' do
-        expect { is_expected.to raise_error Puppet::Error }
+    on_supported_os.each do |os, facts|
```
ekohl, line 44: Recently the recommended pattern changed to use `os_facts` so you don't have an overlapping variable with a let block.
```suggestion
    on_supported_os.each do |os, os_facts|
```

spec/classes/mod/authnz_ldap_spec.rb
```diff
@@ -41,50 +41,53 @@
...
+          let(:params) { { verify_server_cert: false } }
+
+          it { is_expected.to contain_file('authnz_ldap.conf').with_content(%r{^LDAPVerifyServerCert Off$}) }
+        end
+
+        context 'verify_server_cert = wrong' do
+          let(:params) { { verify_server_cert: 'wrong' } }
+
+          it 'raises an error' do
+            expect { is_expected.to raise_error Puppet::Error }
```
ekohl, line 75: I know it was there before, but you may want to use a more descriptive syntax:
```suggestion
            is_expected.to compile.and_raise_error
```

spec/classes/mod/authnz_ldap_spec.rb
```diff
@@ -41,50 +41,53 @@
...
+          end
+        end
+
+        context 'SCL' do
```
ekohl, line 79: I think the SCL is only relevant on EL 6 and 7

spec/classes/mod/authnz_ldap_spec.rb
sanfrancrisko: I've had to reinstate `next` as it violates Rubocop's syntax checks

PR #2023, opened 2020-05-02
Replace validate_apache_loglevel() with data type

alexjfisher:
The function was introduced before Puppet 4 was released.  It was
originally created to reduce duplicated use of stdlib's `validate_re`.

See https://github.com/puppetlabs/puppetlabs-apache/pull/1097


spec/type_aliases/loglevel_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    'warn ssl:info',
+    'warn mod_ssl.c:info',
+    'warn mod_ssl.c:info',
+    'warn ssl_module:info',
+    'trace4',
+  ].each do |allowed_value|
+    it { is_expected.to allow_value(allowed_value) }
+  end
+
+  [
```
ekohl, line 15: How about `xemerg` to test if `^` at the start is needed?

types/loglevel.pp
```diff
@@ -0,0 +1,2 @@
+# @summary As per http://httpd.apache.org/docs/current/mod/core.html#loglevel
```
ekohl, line 1: Would `@see` be useful for the URL?

spec/type_aliases/loglevel_spec.rb
alexjfisher: The regex is copied from the the function and is very lax and could be better.  I thought about trying to make it stricter, but decided that a change like that might be suited to a separate PR.

As things stand a `^` at the beginning would be bad. The apache docs describe the allowed syntax as...

> Syntax
> LogLevel [module:]level [module:level] ...

I think in a perfect world, $log_level would have actually accepted an `Array`. I doubt it's worth introducing that sort of change now though.

types/loglevel.pp
alexjfisher: I've had a go at making the puppet-strings based docs better. Changes to the REFERENCE.md aren't in this commit as it's very difficult to exclude unrelated changes.  It looks like the REFERENCE.md will be updated as part of the release process.

manifests/init.pp
```diff
@@ -195,8 +195,8 @@
 #   > **Note**: Do not configure this parameter manually without special reason.
 #
 # @param log_level
-#   Changes the error log's verbosity. Valid options are: `alert`, `crit`, `debug`, `emerg`, `error`, 
-#   `info`, `notice` and `warn`.
+#   Configures the apache [LogLevel](https://httpd.apache.org/docs/current/mod/core.html#loglevel) directive
```
sanfrancrisko, line 198:
```suggestion
#   Configures the Apache [LogLevel](https://httpd.apache.org/docs/current/mod/core.html#loglevel) directive
```

PR #2026, opened 2020-05-05
[Function] Directory /var/www [Fix] Variable vhost_load_dir not defined

vegaaz:
Hi,

for own purposes I needed to find a way to add the /var/www Directory in the apache2.conf, therefore I implemented it with given code.

In this time I realised the used Variable vhost_load_dir was not defined, I changed it to vhost_dir.

Greetings,
vegaaz


vegaaz:
Realised the variable fix is unnecessary

PR #2027, opened 2020-05-05
[Feature] Add /var/www Directory to apache2.conf

vegaaz:
Hi,

for own needs I implemented the opportunity to define /var/www directory in apache2.conf.
I think this can be useful and was implemented with existing code.

Greetings,
vegaaz


sanfrancrisko:
Hi @vegaaz - thanks for your proposed enhancement! Can I confirm what benefit this change would bring above defining `/var/www` as a directory, like so:
```
directories => [
  { path => '/var/www', },
],
```

vegaaz:
> Hi @vegaaz - thanks for your proposed enhancement! Can I confirm what benefit this change would bring above and beyond defining `/var/www` as a directory, like so:
> 
> ```
> directories => [
>   { path => '/var/www', options => ['FollowSymLinks'], },
> ],
> ```

Hi @sanfrancrisko  thanks for your comment.
To be honest, I didn't see this way... Thanks!
So there is no benefit... 

A little bit ashamed, but happy for the new knowledge I close this request. :D

sanfrancrisko:
No problem @vegaaz - glad to be able to help, please do not hesitate to reach out if you have any queries.

PR #2032, opened 2020-05-20
(IAC-746) - Add ubuntu 20.04 support

david22swan:
IAC-746
pdk version: `1.17.0`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,22 +6,39 @@ on:
       - 'release'
 
 jobs:
-  LitmusAcceptance:
+  LitmusAcceptancePuppet5:
     env:
       HONEYCOMB_WRITEKEY: 7f3c63a70eecc61d635917de46bea4e6
       HONEYCOMB_DATASET: litmus tests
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -6,22 +6,39 @@ on:
...
     steps:
     - uses: actions/checkout@v1
+    - name: Litmus Parallel
+      uses: puppetlabs/action-litmus_parallel@master
+      with:
+        platform: ${{ matrix.platform }}
+        agent_family: ${{ matrix.agent_family }}
+  LitmusAcceptancePuppet6:
+    env:
+      HONEYCOMB_WRITEKEY: 7f3c63a70eecc61d635917de46bea4e6
```
ekohl, line 30: Shouldn't you use secrets for this?

david22swan: @ekohl Sorry for the late reply, since this is a writekey and can only be used to send events to a specific dataset we don't currently hide it. We may look into it the possibility in the future however, depending on how things turn out.

PR #1891, opened 2019-03-08
(MODULES-8731) Allow CIDRs for proxy_ips/internal_proxy in remoteip

JAORMX:
The recent addition of data types for this module [1] introduced an
issue, where CIDRs are no longer allowed. This allows those sort of
values.

This addresses https://tickets.puppetlabs.com/browse/MODULES-8731

[1] https://github.com/puppetlabs/puppetlabs-apache/commit/1503035d788f9fdaa2bab9cdaabfb66e858c2c39#diff-c2ea3c67760696a0d67bab9fb81757c6

PR #2033, opened 2020-05-23
Allow IPv6 CIDRs for proxy_protocol_exceptions in mod remoteip

thechristschn:
This is the same fix as in https://github.com/puppetlabs/puppetlabs-apache/pull/1891


sheenaajay:
Thank you @thechristschn for submitting the PR.

PR #2034, opened 2020-05-25
Allow switching of thread module

tuxmea:
Extends #1961 to also work correctly on Debian 10 and Ubuntu 20


tuxmea:
Disabling class uses another filename: event.conf and event.load instead of the file name on the system which are prefixed with mpm_.
Needs checking for both possible file names and disabling them.

adrianiurca:
Hi @tuxmea,
Thank you for your contribution.

PR #2052, opened 2020-07-27
Use python3-mod_wsgi instead of mod_wsgi on CentOS8

kajinamit:
The mod_wsgi package was renamed to python3-mod_wsgi on CentOS8, and
now the old name doesn't work since the provides and obsoletes for it
was removed[1].

[1] https://centos.pkgs.org/8/centos-appstream-x86_64/python3-mod_wsgi-4.6.4-4.el8.x86_64.rpm.html

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -240,7 +240,10 @@ class apache::params inherits ::apache::version {
       # See http://wiki.aaf.edu.au/tech-info/sp-install-guide
       'shibboleth'            => 'shibboleth',
       'ssl'                   => 'mod_ssl',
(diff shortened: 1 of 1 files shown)


daianamezdrea:
Hi @kajinamit, this looks good and well documented! Thank you! Cheers!

kajinamit:
@mwhahaha 
I agree that it's better to update the default to avoid future effort, but there are some more places where we use default values for older releases and this is why I fixed only CentOS8 in this patch.
Note that to update the default we should be careful about Amazon Linux because Amazon Linux depends on the same logics but has completely different version assignment.

PR #2053, opened 2020-07-27
[CentOS] Always use distrelease to detect OS major versions

kajinamit:
This change makes sure that apache::version::distrelease is always used
to detect OS major version, instead of facts, so that the same logic is
used for the version detection.


daianamezdrea:
Hi @kajinamit, this looks good to me ! Thank you!

PR #2060, opened 2020-08-10
Use Ruby 2.7 compatible string matching

ekohl:
In Ruby 2.7, performing true =~ %r{false}i raises a deprecation warning.

### DIFF
--- a/lib/puppet/functions/apache/bool2httpd.rb
+++ b/lib/puppet/functions/apache/bool2httpd.rb
@@ -17,8 +17,14 @@ Puppet::Functions.create_function(:'apache::bool2httpd') do
   #   apache::bool2httpd(undef) # returns 'Off'
   #
   def bool2httpd(arg)
-    return 'Off' if arg.nil? || arg == false || arg =~ %r{false}i || arg == :undef
-    return 'On' if arg == true || arg =~ %r{true}i
+    return 'Off' if arg.nil? || arg == false || matches_string?(arg, %r{false}i) || arg == :undef
+    return 'On' if arg == true || matches_string?(arg, %r{true}i)
     arg.to_s
+  end
+
+  private
+
+  def matches_string?(value, matcher)
+    value.is_a?(String) && value.match?(matcher)
   end
 end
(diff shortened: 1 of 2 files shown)


blodone:
since this code change our run failes with 
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, Failed to parse template apache/httpd.conf.erb:
  Filepath: /etc/puppetlabs/code/environments/production_master/modules/apache/lib/puppet/functions/apache/bool2httpd.rb
  Line: 28
  Detail: undefined method `match?' for "Off":String
 (file: /etc/puppetlabs/code/environments/production_master/modules/apache/manifests/init.pp, line: 795, column: 18)
```

ekohl:
Which Puppet and Ruby versions are used? The module is compatible with Puppet 5.5.10 and newer. The AIO install bundles Ruby 2.4 and that has the method.

blodone:
we are using puppet 5.5.14 and ruby 2.3.1p112. seems like its a ruby version thing

ekohl:
It does indeed look like the method was introduced in Ruby 2.4.

0x6d617474:
We are running puppet 5.5.22 with ruby 2.4.10p364 and also get the same error. 
```
# /opt/puppetlabs/puppet/bin/ruby --version
ruby 2.4.10p364 (2020-03-31 revision 67879) [x86_64-linux]
```

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, Failed to parse template apache/httpd.conf.erb:
  Filepath: /etc/puppetlabs/code/environments/puppet_module_updates/modules/apache/lib/puppet/functions/apache/bool2httpd.rb
  Line: 28
  Detail: undefined method `match?' for "On":String
 (file: /etc/puppetlabs/code/environments/puppet_module_updates/modules/apache/manifests/init.pp, line: 795, column: 18)
```

sanfrancrisko:
Hi folks, 

We have a fix for this merged in #2102 and hope to get a new version of the module out this week.

PR #2064, opened 2020-09-10
removes spaces from default logfile

rsynnest:
If the user's classname has spaces in it and no logfile parameters are set (e.g. error_log_file, error_log_pipe, error_log_syslog), then our default logfile name will have spaces in it. This commit replaces spaces with underscores in the default logfile. This helps prevent bugs like logrotate failing on logfiles with spaces, which leads to infinite apache logs.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -6468,9 +6468,9 @@ Default value: `false`
 Data type: `Any`
 
 Sets the filename of the `*_access.log` placed in `logroot`. Given a virtual host ---for
-instance, example.com--- it defaults to 'example.com_ssl.log' for
+instance, `my example.com` --- it defaults to 'my_example.com_ssl.log' for
 [SSL-encrypted](https://httpd.apache.org/docs/current/ssl/index.html) virtual hosts and
-`example.com_access.log` for unencrypted virtual hosts.
+`my_example.com_access.log` for unencrypted virtual hosts.
 
 Default value: `false`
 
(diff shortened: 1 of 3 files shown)


igalic (approved):
this is possibly a breaking change, but one that makes a lot of sense.

daianamezdrea:
Hi @igalic, what do you mean by  **possibly a breaking change** ? Do you think this apache change is breaking enough for a major version bump?

igalic:
I've been thinking about that, since leaving my review

i firmly believe that "version numbers are cheap" and that's all there is to it

but there are enough people out there (especially in Ops) who are afraid of them, so I've gotten more careful about slapping on the backwards-incompatible label on issues

ekohl:
> i firmly believe that "version numbers are cheap" and that's all there is to it

I firmly believe that that is false. Dealing with major version bumps is painful on the consumer side. However, probably not as painful as a breaking change that was not labeled as such. They are both a pain and a blessing.

In this case, the potential implications are that other tools process the logs and hardcode the paths. Spaces are less common so it doesn't affect all users.

Technically I'd agree this changes behavior and this could warrant a major version bump. However, I wouldn't do a major version bump just for this single patch. There are too many consumers out there unaffected by this (widely used) module. That leaves keeping the PR open and releasing it as minor. Personally I'd release it as a minor version.

sanfrancrisko:
Hi @rsynnest , @igalic , @ekohl 

We have a ticket raised to implement this work and #2068 's in a manner that will allow us to do it in a minor version bump: https://tickets.puppetlabs.com/browse/IAC-1186

We'll raise a PR that encompasses and closes this one and request your reviews on it too.

sanfrancrisko:
Hi @rsynnest , @igalic , @ekohl

Would you be able to have a peek at #2086 and see if it would be a satisfactory combination of this PR and #2068 's work?

PR #2066, opened 2020-09-14
Resolve easy puppet-lint notices

jcpunk:
This resolves the majority of puppet-lint notices.


igalic:
the test failures are facter related and have nothing to do with your change

sanfrancrisko:
@jcpunk This looks like it is a similar issue to what @adrianiurca mentioned [here](https://github.com/puppetlabs/puppetlabs-firewall/pull/943#issuecomment-693520842)

jcpunk:
I've taken a stab at getting the spec tests to set the right facts for the modern layout.

I'll confess I could use a hand with `spec_helper_acceptance_local.rb` among others.

jcpunk:
I just looked at the size of that patch....

New plan, I'm going to just post a patch for the non-fact fixes...

adrianiurca:
Hi @jcpunk, thank you for your contribution!

adrianiurca (approved):

PR #2068, opened 2020-09-24
Make the logfiles use servername instead of name in their name.

pstray:
Since we already use $ssl to add "_ssl" to the log filenames, and the
examples for a mix of SSL and not typically would result in logfiles
with names like "mix.example.com non-ssl_access.log" and
"mix.example.com ssl_access_ssl.log", this patch fixes that by using
$servername instead and thus getting what it probably more expected:
"mix.example.com_access.log" and "mix.example.com_access_ssl.log"

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2070,9 +2070,9 @@ define apache::vhost(
     $error_log_destination = $error_log_syslog
   } else {
     if $ssl {
-      $error_log_destination = "${logroot}/${name}_error_ssl.log"
+      $error_log_destination = "${logroot}/${servername}_error_ssl.log"
     } else {
-      $error_log_destination = "${logroot}/${name}_error.log"
+      $error_log_destination = "${logroot}/${servername}_error.log"
     }
   }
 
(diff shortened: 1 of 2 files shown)


sanfrancrisko:
Thanks for the suggested tweak @pstray .

There's another PR currently open that is proposing changes in the same area: https://github.com/puppetlabs/puppetlabs-apache/pull/2064

I'll discuss with the IAC Team on how we want to proceed with both of these changes. Both very useful, however, we need to take in to consideration how we publicise and version this, as there could be assumptions in place about log file name formats, for other tools, etc...

sanfrancrisko:
Hi @pstray

Further to my comment above, we have a ticket raised to implement this work and #2064's in a manner that will allow us to do it in a minor version bump: https://tickets.puppetlabs.com/browse/IAC-1186

We'll raise a PR that closes this one and request your review on it too.

sanfrancrisko:
Hi @pstray 

Would you be able to have a peek at #2086 and see if it would be a satisfactory combination of this PR and #2064 's work

pstray:
Yes that seems like a good way to do it

sanfrancrisko:
Thanks @pstray , I'll close this PR given #2086 incorporates this work (_I thought it would have been automatically closed from that PR, strangely_).

Thanks again for your suggestion.

PR #2070, opened 2020-09-28
Replace some occurrences of non-ASCII chars

DLeich:
Several lines in the comments of apache::vhost and apache::mod::jk have non-ascii characters embedded.  This can throw off some testing suites, such as Oneceover, which will complain about it.  This pull request replaces non-ascii characters with acceptable alternatives.


sanfrancrisko (approved):
Thanks for the fix up @DLeich

PR #2071, opened 2020-09-30
make apache::mod::fcgid redhat 8 compatible

creativefre:
before -> rhel 8, fcgid.load is loaded before unixd.load  which causes httpd to fail
So the name unixd_fcgid.load should also be implemented for rhel 8


adrianiurca:
Hi @creativefre, thank you for your contribution.

adrianiurca (approved):

PR #2072, opened 2020-10-02
Add ssl_proxy_machine_cert_chain param to vhost class

AbelNavarro:
Added the Apache vhost proxy config option SSLProxyMachineCertificateChainFile.

Reference: https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslproxymachinecertificatechainfile

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -9070,6 +9070,20 @@ apache::vhost { 'sample.example.net':
 
 Default value: `undef`
 
+##### `ssl_proxy_machine_cert_chain`
+
+Data type: `Any`
+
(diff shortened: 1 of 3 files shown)


REFERENCE.md
```diff
@@ -9070,6 +9070,20 @@ apache::vhost { 'sample.example.net':
 
 Default value: `undef`
 
+##### `ssl_proxy_machine_cert_chain`
+
+Data type: `Any`
```
igalic, line 9075: it would be nice if we could have this as Stdlib::Absolutepath… eventually.

sanfrancrisko (approved):
Agree with [your comment](https://github.com/puppetlabs/puppetlabs-apache/pull/2072#discussion_r499281401) @igalic , however, happy enough to merge now.

Thanks for the enhancement @AbelNavarro 👍

PR #2074, opened 2020-10-06
Use Ruby 2.7 compatible string matching

sanfrancrisko:
#2060 was incorrectly merged to `master` instead of `main`:
> In Ruby 2.7, performing `true =~ %r{false}i` raises a deprecation warning.


ekohl (approved):
I also noticed the default branch was switched and doesn't include this anymore, leading to deprecation warnings again.

PR #2081, opened 2020-11-02
Improve SSLVerify options

bovy89:
Enable SSLVerifyClient and SSLVerifyDepth inside locations. Add SSLVerifyClient validation. SSLCACertificate (and related options) can now be specified even without SSLVerifyClient (can be inside a location).

### DIFF
--- a/templates/vhost/_ssl.erb
+++ b/templates/vhost/_ssl.erb
@@ -18,6 +18,10 @@
   <%- end -%>
   <%- if @ssl_verify_client -%>
   SSLVerifyClient         <%= @ssl_verify_client %>
+  <%- if @ssl_verify_depth -%>
+  SSLVerifyDepth          <%= @ssl_verify_depth %>
+  <%- end -%>
+  <%- end -%>
   <%- if @ssl_certs_dir && @ssl_certs_dir != '' -%>
   SSLCACertificatePath    "<%= @ssl_certs_dir %>"
   <%- end -%>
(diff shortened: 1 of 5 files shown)


templates/vhost/_ssl.erb
```diff
@@ -16,8 +16,12 @@
   <%- if @ssl_honorcipherorder -%>
   SSLHonorCipherOrder     <%= @ssl_honorcipherorder %>
   <%- end -%>
-  <%- if @ssl_verify_client -%>
+  <%- if @ssl_verify_client and @ssl_verify_client.match('(none|optional|require|optional_no_ca)') -%>
```
ekohl, line 19: I think it'd better to have a data type to enforce it. This silently ignores invalid values.

bovy89: Is this a requirement or can be ignored in this case.

ekohl: I think this is a bad pattern and should be avoided. I opened https://github.com/puppetlabs/puppetlabs-apache/pull/2084 which is IMHO the correct pattern. This notifies the user that they did something wrong. With your solution they might find out only in production when it's broken.

sanfrancrisko: Thanks @bovy89 for the PR and @ekohl for the suggestion in #2084 - @bovy89 if you incorporate @ekohl 's suggestion to this PR, or want to continue working from there, we'll be happy to merge.

bovy89: done

sanfrancrisko (approved):
Thanks for the improvement @bovy89 and the help reviewing @ekohl 👍 

Closes: #2084

PR #2084, opened 2020-11-04
Check the type for ssl_verify_client

ekohl:
These are the only valid options according to the manual and prevents misconfiguration.


sanfrancrisko:
Thanks for the help with #2081 @ekohl - I'll close this PR given it's now merged

PR #2085, opened 2020-11-05
Fix apreq2 LoadModule

TigerKriika:
The LoadModule ID is currently wrong created.

The module creates the .load-File with
>LoadModule **apreq2_module**    modules/mod_apreq2.so

instead of

> LoadModule **apreq_module**    modules/mod_apreq2.so

http://httpd.apache.org/apreq/docs/libapreq2/group__mod__apreq2.html

### DIFF
--- a/manifests/mod/apreq2.pp
+++ b/manifests/mod/apreq2.pp
@@ -1,0 +1,11 @@
+# @summary
+#   Installs `mod_apreq2`.
+#
+# @see http://httpd.apache.org/apreq/docs/libapreq2/group__mod__apreq2.html for additional documentation.
+#
+
+class apache::mod::apreq2 {
+  ::apache::mod { 'apreq2':
+    id => 'apreq_module',
+  }
+}


sanfrancrisko:
Thanks for the new feature to add support for the `apreq2_module`. From what you're describing, you seem to be inferring that with the current behaviour, there is a `.load` file being created with the module name set to `apreq2_module`? I'm not sure I see that behaviour currently - could you provide a manifest that will trigger the load of the `apreq2_module` module?

This change looks good, but it would need an acceptance test to verify that the module is indeed being loaded. Let me know if I can be of any help in that regard

TigerKriika:
I used the following Manifest on a Debian 10 (buster) machine:

```
class profile::web {
	class { 'apache' :
		default_charset   => 'utf-8',
		keepalive_timeout => 5,
		timeout           => 300,
		server_signature  => 'Off',
		trace_enable      => 'Off',
		mpm_module        => false,
		log_formats       => {
			combined       => '%a %l %u %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\"',
			vhost_combined => '%v:%p %a %l %u %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\"',
			common         => '%a %l %u %t \"%r\" %>s %O',
			referer        => '%{Referer}i -> %U',
			agent          => '%{User-agent}i'
		},
	}

	class { 'apache::mod::apreq2': }
	class { 'apache::mod::expires': }
	class { 'apache::mod::headers': }
	class { 'apache::mod::perl': }
	class { 'apache::mod::prefork': }
	class { 'apache::mod::rewrite': }

}
```

This installs the Apache and a apreq2.load File is created with the following content
> LoadModule apreq2_module modules/mod_apreq2.so

The Apache can't start with this load File since the module name is wrong. You will get the following error-message:
> apache2: Syntax error on line 41 of /etc/apache2/apache2.conf: Syntax error on line 1 of /etc/apache2/mods-enabled/apreq2.load: Can't locate API module structure `apreq2_module' in file /usr/lib/apache2/modules/mod_apreq2.so: /usr/lib/apache2/modules/mod_apreq2.so: undefined symbol: apreq2_module

If I install the apache and the libapache2-mod-apreq2 package manually it creates a apreq2.load File with the following content and the apache starts
> LoadModule apreq_module modules/mod_apreq2.so



>This change looks good, but it would need an acceptance test to verify that the module is indeed being loaded. Let me know if I can be of any help in that regard

 I just started to work with puppet so I would appreciate your help regarding a acceptance test.

sanfrancrisko:
Thanks for the detailed explanation @TigerKriika 

Firstly, regarding the acceptance tests, I would just keep it quite simple to begin with. If you create a new file under `spec/acceptance/mod_apreq2_spec.rb`, then paste this basic test in:

```ruby
require 'spec_helper_acceptance'
apache_hash = apache_settings_hash

describe 'apache::mod::apreq2' do
  pp = <<-MANIFEST
    class { 'apache' : }
    class { 'apache::mod::apreq2': }
  MANIFEST

  it 'succeeds in installing the mod_authnz_apreq2 module' do
  end
end
```

This will reproduce the current issue you are encountering.

You'll need to implement a few more steps to add support for this MOD. The part missing is the package installation of the necessary MODs for the OS for this to be successful.

The way this is typically done is by this _not so great_ method of updating the `$mod_packages` parameter in `params.pp`, [like this](https://github.com/puppetlabs/puppetlabs-apache/blob/main/manifests/params.pp#L360-L454):

Typically, the Apache MOD support is going to be difficult to maintain across all Linux distros and versions. You may want to have a read at [this blog post](https://puppetlabs.github.io/iac/news/2020/07/03/apache-mod-test-improvements.html) I made a while back.

So I would:
- Identify what platform(s) you want to get this MOD working on
- Annotate the `apache::mod::apreq2` class def in `manifests/mod/apreq2.pp` with the `# @note Unsupported platforms:` tag, as per described in the [blog post](https://puppetlabs.github.io/iac/news/2020/07/03/apache-mod-test-improvements.html)
- Add the `if: mod_supported_on_platform('apache::mod::apache::mod::apreq2')` RSpec filtering condition, as outlined the [blog post](https://puppetlabs.github.io/iac/news/2020/07/03/apache-mod-test-improvements.html)
- Enhance the `params.pp` to install the `libapache2-mod-apreq2` package:
```
$mod_packages = {
        'apreq2'                => 'libapache2-mod-apreq2',
        ...
```

sanfrancrisko:
@TigerKriika I've created a PR to your fork with the above changes: https://github.com/TigerKriika/puppetlabs-apache/pull/1 

If you want to expand on the basic acceptance test and/or increase the amount of platforms the MOD is supported on, feel free to do so.

TigerKriika:
@sanfrancrisko 
Thanks for your help. I merged your PR.

PR #2086, opened 2020-11-09
(IAC-1186) Add new $use_servername_for_filenames parameter

sanfrancrisko:
Prior to this commit, the `$filename` variable, which is used to
construct the filename of Apache's various config and log files,
obtained it's default value from the `$name` parameter.

As #2064 highlights, it is possible for `$name` to contain spaces
in it and this can cause cumbersome log file names, albeit POSIX
compliant.

Also related is #2068, which changes the `$filename` variable source
from `$name` to `$servername`. This arguably seems more appropriate,
especially given that `$servername` defaults to `$name` if undefined.

This commit attempts to create a satisfactory solution to both #2064
and #2068 by introducing the `$use_servername_for_filenames` param.
When set to `true`, a sanitized `$servername` parameter value will be
used to construct `$filename`.

When undefined or set to `$false`, it will retain the existing
behaviour and use the `$name` parameter value.

This will default to `false` until the next major release (v6.0.0),
after which it will default to `true`. Then, in the subsequent major
release (v7.0.0) it will be deprecated altogether and the default
behaviour will be to use the sanitized value of `$servername` for the
`$filename` var.

Warning messages are output to the console to alert users of this
change in behaviour.

Closes: #2064 , #2068

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1717,6 +1717,12 @@
 #   client request exceeds that limit, the server will return an error response
 #   instead of servicing the request.
 #
+# @param $use_servername_for_filenames
+#   When set to true, default log / config file names will be derived from the sanitized
+#   value of the $servername parameter.
+#   When set to false (default), the existing behaviour of using the $name parameter
+#   will remain.
+
 define apache::vhost (
   Variant[Boolean,String] $docroot,
   $manage_docroot                                                                   = true,
@@ -1782,6 +1788,7 @@ define apache::vhost (
   $access_log_format                                                                = false,
   $access_log_env_var                                                               = false,
   Optional[Array] $access_logs                                                      = undef,
+  Optional[Boolean] $use_servername_for_filenames                                   = false,
   $aliases                                                                          = undef,
   Optional[Variant[Hash, Array[Variant[Array,Hash]]]] $directories                  = undef,
   Boolean $error_log                                                                = true,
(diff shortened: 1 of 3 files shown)


manifests/vhost.pp
```diff
@@ -2036,8 +2043,39 @@
     $priority_real = '25-'
   }
 
-  ## Apache include does not always work with spaces in the filename
-  $filename = regsubst($name, ' ', '_', 'G')
+  # IAC-1186: A number of configuration and log file names are generated using the $name parameter. It is possile for
```
sheenaajay, line 2046:
```suggestion
  # IAC-1186: A number of configuration and log file names are generated using the $name parameter. It is possible for
```

sheenaajay:
@sanfrancrisko The changes looks good. Thank you.

sheenaajay (approved):

manifests/vhost.pp
```diff
@@ -1717,6 +1717,12 @@
 #   client request exceeds that limit, the server will return an error response
 #   instead of servicing the request.
 #
+# @param $use_servername_for_filenames
+#   When set to true, default log / config file names will be derived from the sanitized
+#   value of the $servername parameter.
+#   When set to false (default), the existing behaviour of using the $name parameter
+#   will remain.
+
```
ekohl, line 1725: This broke puppet-strings: https://github.com/puppetlabs/puppetlabs-apache/pull/2165.

Yet another reason why I hate this change and would still be strongly in favor of reverting it.

PR #2087, opened 2020-11-11
Allow to disable a virtual host

promasu:
This commit adds a param to the apache::vhost type to disable it.
The default behavior stays that a vhost is enabled but the capability to disable it is added.
Sometimes its usefull to deploy a disabled vhost for example to use it with maintenance scripts.

### DIFF
--- a/examples/vhost.pp
+++ b/examples/vhost.pp
@@ -254,3 +254,9 @@ apache::vhost { 'twentysecond.example.com':
   docroot        => '/var/www/twentysecond',
   rack_base_uris => ['/passengerapp1', '/passengerapp2'],
 }
+
+# Disabled basic vhost
+apache::vhost { 'twentythird.example.com':
+  port    => '80',
+  docroot => '/var/www/twentythird',
+}
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1769,6 +1769,7 @@ define apache::vhost (
   $override                                                                         = ['None'],
   $directoryindex                                                                   = '',
   $vhost_name                                                                       = '*',
+  Boolean $enable                                                                   = true,
   $logroot                                                                          = $apache::logroot,
   Enum['directory', 'absent'] $logroot_ensure                                       = 'directory',
   $logroot_mode                                                                     = undef,
@@ -2296,7 +2297,7 @@ define apache::vhost (
   }
   # NOTE(pabelanger): This code is duplicated in ::apache::vhost::custom and
   # needs to be converted into something generic.
-  if $apache::vhost_enable_dir {
+  if $enable and $apache::vhost_enable_dir {
     $vhost_enable_dir = $apache::vhost_enable_dir
     $vhost_symlink_ensure = $ensure ? {
       'present' => link,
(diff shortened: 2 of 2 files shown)


manifests/vhost.pp
```diff
@@ -1214,6 +1214,9 @@
 #   If the virtual host has no assigned IP or port, the virtual host name is set to the 
 #   title of the resource.
 #
+# @param vhost_enable
+#   Specifies if the symlink to enable the virtual host is created or not.
```
igalic, line 1218: can't wer use `ensure` or `enable` parameter name? analogous to `service` and a million other puppet types?

igalic (changes requested):
🙋🏻‍♀️

manifests/vhost.pp
promasu: Yes sure. I changed it to enable as I think ensure is not the right property for this type of parameter

examples/vhost.pp
```diff
@@ -254,3 +254,9 @@
   docroot        => '/var/www/twentysecond',
   rack_base_uris => ['/passengerapp1', '/passengerapp2'],
 }
+
+# Disabled basic vhost
```
igalic, line 258: this comment is wrong, given our default for `enable` is `true`

igalic (changes requested):
it would be nice to see a spec test, confirming this desired behaviour

manifests/vhost.pp
```diff
@@ -2312,6 +2313,11 @@
       require => Concat["${priority_real}${filename}.conf"],
       notify  => Class['apache::service'],
     }
+  } elsif !$enable and $apache::vhost_enable_dir {
+    file { "${priority_real}${filename}.conf symlink":
+      ensure => 'absent',
+      path   => "${vhost_enable_dir}/${priority_real}${filename}.conf",
+      notify  => Class['apache::service'],
```
pcworld, line 2320: The `file` block seems to be missing a closing `}`.  
I suppose this is a good indicator that this should be tested before merging.

manifests/vhost.pp
```diff
@@ -2296,7 +2297,7 @@
   }
   # NOTE(pabelanger): This code is duplicated in ::apache::vhost::custom and
```
pcworld, line 2298: As indicated by this comment, this code is duplicated in https://github.com/puppetlabs/puppetlabs-apache/blob/c3fc5a74019ead9869194c33c8577ce06441c52f/manifests/vhost/custom.pp and one might want to consider replicating the changes there.

pcworld (changes requested):

sanfrancrisko:
Thanks for the proposed change @promasu - there are still a few comments on the PR that look to be valid issues that need addressed. If you still want to get this merged but need some help getting it over the line, please let us know.

pmcmaw:
In line with our current policy, after 90 days of inactivity we close PRs. 

As we haven't heard from you @promasu I have decided the best course of action is to close this PR. 
If you ever get the bandwidth to pick this up, please feel free to reopen and continue working on it. 

As always, I would like to thank you for taking the time to contribute to our supported modules and I look forward to hearing from you in the future.

PR #2089, opened 2020-11-16
Add cas_cookie_path in vhosts

yakatz:
`CASCookiePath` can be set in vhosts


sanfrancrisko (approved):
Thanks for the enhancement @yakatz

PR #2090, opened 2020-11-16
Add support for mod_md

smortex:
This PR add support for [mod_md](https://httpd.apache.org/docs/current/en/mod/mod_md.html), a module for managing domains across virtual hosts and certificate provisioning via the ACME protocol.

### What is implemented

Each of these points have been added as a separate commit for clarity:

* [x] Add support for `mod_watchdog` — it's a dependency of `mod_md`;
* [x] Add support for `mod_md` — enable the module and configure it;
* [x] Connect `mod_md` with `apache::vhost` to make it usable;
* [x] Add basic integration test when the module is enabled.

It brings full support for mod_md into the Apache module.

### What is not supported

Support for the [\<MDomain\> sections](https://httpd.apache.org/docs/current/en/mod/mod_md.html#mdomainsetsection) is not part of this PR.  *\<MDomain\>* allows to override general configuration when requesting certificates from multiple providers.  This use case is rather advanced so I guess support for it can be skipped as of today.  Future contributors might be able to add support for this by allowing to pass an `Enum` to `$apache::vhost::mdomain` with all custom parameters.

### Basic use-case

Lines marked with `<<<` are the one added to the basic configuration thanks to this module:

```puppet
class { 'apache':
}

class { 'apache::mod::md':                        # <<<
  md_certificate_agreement => 'accepted',         # <<<
  md_contact_email         => 'root@example.com', # <<<
}                                                 # <<<

apache::vhost { 'example.com':
  docroot => '/var/www/example.com',
  port    => 443,
  ssl     => true,
  mdomain => true,                                # <<<
}
```

### DIFF
--- a/manifests/mod/watchdog.pp
+++ b/manifests/mod/watchdog.pp
@@ -1,0 +1,31 @@
+# @summary
+#   Installs and configures `mod_watchdog`.
+#
+# @param watchdog_interval
+#   Sets the interval at which the watchdog_step hook runs.
+#
+# @see https://httpd.apache.org/docs/current/mod/mod_watchdog.html for additional documentation.
+class apache::mod::watchdog (
+  Optional[Integer] $watchdog_interval = undef,
+) {
+  include apache
+
+  $module_builtin = $facts['os']['family'] in ['Debian']
+
+  unless $module_builtin {
+    apache::mod { 'watchdog':
+    }
+  }
+
+  if $watchdog_interval {
+    file { 'watchdog.conf':
+      ensure  => file,
+      path    => "${apache::mod_dir}/watchdog.conf",
+      mode    => $apache::file_mode,
+      content => "WatchdogInterval ${watchdog_interval}\n",
+      require => Exec["mkdir ${apache::mod_dir}"],
+      before  => File[$apache::mod_dir],
+      notify  => Class['apache::service'],
+    }
+  }
+}
--- a/templates/vhost/_file_header.erb
+++ b/templates/vhost/_file_header.erb
@@ -3,6 +3,14 @@
 # Managed by Puppet
 # ************************************
 <%= [@comment].flatten.collect{|c| "# #{c}"}.join("\n") -%>
+<% if @mdomain -%>
+
+  <%- if @mdomain.is_a?(String) -%>
+MDomain <%= @mdomain %>
+  <%- else -%>
+MDomain <%= @servername %>
+  <%- end -%>
+<% end -%>
 
 <VirtualHost <%= [@nvh_addr_port].flatten.compact.join(' ') %>>
 <% @define.each do | k, v| -%>
--- a/templates/vhost/_ssl.erb
+++ b/templates/vhost/_ssl.erb
@@ -2,8 +2,10 @@
 
   ## SSL directives
   SSLEngine on
+  <%- unless @mdomain -%>
   SSLCertificateFile      "<%= @ssl_cert %>"
   SSLCertificateKeyFile   "<%= @ssl_key %>"
+  <%- end -%>
   <%- if @ssl_chain -%>
   SSLCertificateChainFile "<%= @ssl_chain %>"
   <%- end -%>
--- a/manifests/mod/md.pp
+++ b/manifests/mod/md.pp
@@ -1,0 +1,137 @@
+# @summary
+#   Installs and configures `mod_md`.
+#
+# @param md_activation_delay
+#   -
+#
+# @param md_base_server
+#   Control if base server may be managed or only virtual hosts.
+#
+# @param md_ca_challenges
+#   Type of ACME challenge used to prove domain ownership.
+#
+# @param md_certificate_agreement
+#   You confirm that you accepted the Terms of Service of the Certificate
+#   Authority.
+#
+# @param md_certificate_authority
+#   The URL of the ACME Certificate Authority service.
+#
+# @param md_certificate_check
+#   -
+#
+# @param md_certificate_monitor
+#   The URL of a certificate log monitor.
+#
+# @param md_certificate_protocol
+#   The protocol to use with the Certificate Authority.
+#
+# @param md_certificate_status
+#   Exposes public certificate information in JSON.
+#
+# @param md_challenge_dns01
+#   Define a program to be called when the `dns-01` challenge needs to be
+#   setup/torn down. 
+#
+# @param md_contact_email
+#   The ACME protocol requires you to give a contact url when you sign up.
+#
+# @param md_http_proxy
+#   Define a proxy for outgoing connections.
+#
+# @param md_members
+#   Control if the alias domain names are automatically added.
+#
+# @param md_message_cmd
+#   Handle events for Manage Domains.
+#
+# @param md_must_staple
+#   Control if new certificates carry the OCSP Must Staple flag.
+#
+# @param md_notify_cmd
+#   Run a program when a Managed Domain is ready.
+#
+# @param md_port_map
+#   Map external to internal ports for domain ownership verification.
+#
+# @param md_private_keys
+#   Set type and size of the private keys generated.
+#
+# @param md_renew_mode
+#   Controls if certificates shall be renewed.
+#
+# @param md_renew_window
+#   Control when a certificate will be renewed.
+#
+# @param md_require_https
+#   Redirects http: traffic to https: for Managed Domains.
+#
+# @param md_server_status
+#   Control if Managed Domain information is added to server-status.
+#
+# @param md_staple_others
+#   Enable stapling for certificates not managed by mod_md.
+#
+# @param md_stapling
+#   Enable stapling for all or a particular MDomain.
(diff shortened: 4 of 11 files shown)


sanfrancrisko (commented):
Thank you very much @smortex for this excellent enhancement. We really appreciate the atomic commits, test coverage and detail you've put in to the description 👍 

I'd be happy to merge this as is, but I do wonder if you would be up for adding a basic acceptance test to ensure that the MOD installs OK on a platform? I was having [a conversation](https://github.com/puppetlabs/puppetlabs-apache/pull/2085#issuecomment-732276278) with another contributor on [their PR](https://github.com/puppetlabs/puppetlabs-apache/pull/2085) about this too. 

Fully aware that trying to support Apache MODs on all versions of all Linux distros is somewhat of a nightmare (_and even if it works now, it may not do so in a future release!_). 

We recently added functionality to limit the test execution to platforms we only wanted to support - you can read about that [here](https://github.com/puppetlabs/puppetlabs-apache#apache-mod-test--support-lifecycle).

If you wanted to identify the platform(s) you're interested in and add a basic acceptance test  similar to what I did in https://github.com/TigerKriika/puppetlabs-apache/pull/1, then it would allow us to catch any potential future regressions caused by a tweak in package name. This is a very common problem with Apache MODs between new versions of OSs.

Thanks again for the excellent contribution - will look forward to getting it over the line very soon. Give me a shout if I can be of any help regarding a basic acceptance test.

templates/vhost/_ssl.erb
```diff
@@ -2,8 +2,10 @@
 
   ## SSL directives
   SSLEngine on
+  <%- unless @mdomain -%>
   SSLCertificateFile      "<%= @ssl_cert %>"
   SSLCertificateKeyFile   "<%= @ssl_key %>"
+  <%- end -%>
   <%- if @ssl_chain -%>
```
igalic, line 9: is this still useful? or should we add that to be excluded?

igalic (commented):
🙋🏻‍♀️

templates/vhost/_ssl.erb
smortex: According to [mod_ssl doc](https://httpd.apache.org/docs/current/en/mod/mod_ssl.html#sslcertificatechainfile):

> SSLCertificateChainFile became obsolete with version 2.4.8, when SSLCertificateFile was extended to also load intermediate CA certificates from the server certificate file.

Some supported old distributions might still require it (It looks like RedHat 6 has Apache 2.2)…

smortex (commented):
Thanks for suggesting adding some acceptance test @sanfrancrisko, it helped spot an issue that affected Debian :+1: 

I fixed this issue by amending the commit, and added the acceptance test as a new commit.

templates/vhost/_file_header.erb
```diff
@@ -3,6 +3,14 @@
...
 # ************************************
 <%= [@comment].flatten.collect{|c| "# #{c}"}.join("\n") -%>
+<% if @mdomain -%>
+
+  <%- if @mdomain.is_a?(String) -%>
+MDomain <%= @mdomain %>
+  <%- else -%>
+MDomain <%= @servername %>
+  <%- end -%>
+<% end -%>
```
igalic, line 13: do you know how well this template will fare with `MDRequireHttps permanent` given that our standard vhost spreads a http and https over two files?

igalic (commented):
i was just looking at my own config and how i handle that stuff there, and I use `apache::vhost::custom` to generate the following:

```apache
MDomain igalic.co auto

<VirtualHost *:80>
  ServerName igalic.co
  DocumentRoot '/var/empty'
</VirtualHost>

<VirtualHost *:443>
  ServerName igalic.co
  DocumentRoot /srv/igalic.co/site
  SSLEngine On
</VirtualHost>
```

That being said:

templates/vhost/_file_header.erb
smortex: I have not yet used `MDRequireHttps`, I rely on the code I was previously using in my `profile::apache` which supposedly does the same.  I guess that I would use `MDRequireHttps temporary` when attempting to change to ensure no side effect occurs…

Are you in a position you can check?

smortex: On my infra: after removing the shared HTTP VirtualHost that handled redirects to https, setting `MDRequireHttps` and adding a `apache::listen { '80': }` so that the server listen on port 80; the redirect do not happen.

A quick test shows that you _must_ have a HTTP VirtualHost beside each HTTPS VirtualHost like you have, whatever the content…  For now, I reverted to my custom single point of redirection from my apache profile.

igalic: > Are you in a position you can check?

i'm in a bathtub.

what if you create an HTTP vhost, with the `MDomain` stanza, and without SSL (obviously?).
does the redirect work then?

smortex: It does redirect to https.  If you have `MDRequireHttps` and a VirtualHost on port 80, it will redirect to https.  No need for a VirtualHost on port 443 :neutral_face:.

igalic: i think we might need to warn users, either in the code or in the docs about this behaviour

smortex: My feeling is that it's really a misconfiguration from the user: it really makes no sense to setup a http-only VirutalHost and set MDomain for it…  So a broken server in such a case is clearly what I would expect.  Just like if I set docroot to a directory that does not exist.

Maybe I do not understand your point because during all the tests i did in this conversation I was wondering what you actually wanted to know / where you wanted to go.

manifests/mod/watchdog.pp
```diff
@@ -10,7 +10,11 @@
 ) {
   include apache
 
-  apache::mod { 'watchdog':
+  $module_builtin = facts['os']['family'] in ['Debian']
+
+  if $module_builtin {
+    apache::mod { 'watchdog':
+    }
```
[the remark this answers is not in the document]
sanfrancrisko: Interesting! Tests FTW 😃  

Out of curiosity, what other platforms did you run the tests against? I can run this against [our internal infrastructure](https://github.com/puppetlabs/puppetlabs-apache/blob/bd7942973532003ee00c8a3e32fc4399ef75ee28/provision.yaml#L43-L100) to get more coverage if you felt it was valuable?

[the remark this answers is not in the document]
smortex: I proceeded this way:

1. add an acceptance test, commit and push, see where it fails
2. inspect failures and fix them appropriately when it's not supposed to fail (RedHat 8: add a dependency on the `mod_md` package; Debian 10 / Ubuntu 20: handle the built-in module);
3. filter all failures which are expected because the version of apache is too old or no package is availalbe using the `@note Unsupported platforms` stanza.

It seems that all plaforms listed in your link should either work or be excluded, so running against your internal infrastructure should pass :wink:

templates/vhost/_file_header.erb
igalic: > it really makes no sense to setup a http-only VirutalHost and set MDomain for it… So a broken server in such a case is clearly what I would expect. Just like if I set docroot to a directory that does not exist.

sorry, that's not what i meant:
I meant that when you setup `MDRequireHttps $redirect`; then you *also* need to setup an http Vhost (with `MDomain`?), not just an https vhost.

> Maybe I do not understand your point because during all the tests i did in this conversation I was wondering what you actually wanted to know / where you wanted to go.

so i  meant that we should warn people when that's not the case, but i now realize that's really hard to do; so the least we can do is put it into the documentation.

smortex: Ah!  Got it!

@igalic, what about adding a line in the in-line documentation like:

```diff
 #
 # @param md_require_https
 #   Redirects http: traffic to https: for Managed Domains.
+#   An http: Virtual Host must nevertheless be setup for that domain.
 #
 # @param md_server_status
 #   Control if Managed Domain information is added to server-status.
 
```

smortex: Fixed, thanks!

manifests/mod/watchdog.pp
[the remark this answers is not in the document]
smortex: Kindly reminder for @sanfrancrisko in case my last message slept through :wave:.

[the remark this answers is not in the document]
sanfrancrisko: Apologies for the delay getting back @smortex - there was a mix up on covering this module last week. I'll look to get this over the line and released this week.

sanfrancrisko (approved):
This looks good @smortex - thanks again. I'll leave it open for another 48 hours or so to give @igalic a chance to approve / confirm all her queries were addressed. I believe that is the case, but just want to be 100% sure.

igalic (approved):

david22swan:
@smortex 
This looks good to me and already has two approves on it so I'm gonna go ahead and merge.
Thanks for the work you put in :)

PR #2093, opened 2020-11-20
Add ssl_user_name vhost parameter

bodgit:
Maps to SSLUserName directive.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1673,6 +1673,9 @@
 #   Can be used to set the [SSLStaplingReturnResponderErrors](http://httpd.apache.org/docs/current/mod/mod_ssl.html#sslstaplingreturnrespondererrors) directive.<br />
 #   This parameter only applies to Apache 2.4 or higher and is ignored on older versions.
 #
+# @param ssl_user_name
+#   Sets the [SSLUserName](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslusername) directive.
+#
 # @param use_canonical_name
 #   Specifies whether to use the [`UseCanonicalName directive`](https://httpd.apache.org/docs/2.4/mod/core.html#usecanonicalname),
 #   which allows you to configure how the server determines it's own name and port.
@@ -1767,6 +1770,7 @@ define apache::vhost (
   Optional[Boolean] $ssl_stapling                                                   = undef,
   $ssl_stapling_timeout                                                             = undef,
   $ssl_stapling_return_errors                                                       = undef,
+  Optional[String] $ssl_user_name                                                   = undef,
   $priority                                                                         = undef,
   Boolean $default_vhost                                                            = false,
   $servername                                                                       = $name,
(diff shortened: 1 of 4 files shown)


manifests/vhost.pp
```diff
@@ -1660,6 +1660,9 @@
 # @param ssl_proxyengine
 #   Specifies whether or not to use [SSLProxyEngine](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslproxyengine).
 #
+# @param ssl_require
+#   Sets the [SSLRequire](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslrequire) directive.
```
ekohl, line 1664: This states:
> SSLRequire is deprecated and should in general be replaced by Require expr.

Is it really needed to add a deprecated option?

bodgit: If I can convince my users to use that, sure. Where/How do you set that, looking for the word "require" in a Puppet codebase brings up a few hits 😉

ekohl: I think they can be set via the `directories` parameter. It supports a lot of options and and one of them is `require`.

bodgit: I've removed that parameter now.

bodgit:
Updated to just add the `ssl_user_name` parameter given `SSLRequire` is a deprecated directive.

spec/defines/vhost_spec.rb
```diff
@@ -85,6 +85,7 @@
               'ssl_proxyengine'             => true,
               'ssl_proxy_cipher_suite'      => 'HIGH',
               'ssl_proxy_protocol'          => 'TLSv1.2',
+              'ssl_user_name'               => 'SSL_CLIENT_S_DN_CN',
```
igalic, line 88: where is the verification of this spec addition?

igalic (commented):
🙋🏻‍♀️

spec/defines/vhost_spec.rb
bodgit: It looks like a lot of the SSL options aren't actually verified https://github.com/puppetlabs/puppetlabs-apache/blob/bd7942973532003ee00c8a3e32fc4399ef75ee28/spec/defines/vhost_spec.rb#L1041

I'll add something for this.

bodgit: Done.

david22swan (changes requested):
Thank you for the added functionality, have just reviewed this and while it look's good overall to me I would like to see an acceptance test added to cover the functionality. Doesn't need to be complex, just an instance of it being set:
https://github.com/puppetlabs/puppetlabs-apache/blob/main/spec/acceptance/vhost_spec.rb
If you can do this I would feel happy merging.

sanfrancrisko:
Thanks for the enhancement @bodgit - as @david22swan mentioned [here|https://github.com/puppetlabs/puppetlabs-apache/pull/2093#pullrequestreview-540773693], if we can get an acceptance test added, then we'd be happy to merge. Please let me know if we can be of any assistance in that regard.

bodgit:
Acceptance test added, and corrected a small typo.

sanfrancrisko (approved):
Thanks for the enhancement @bodgit and the review help @ekohl , @igalic , @david22swan

PR #2103, opened 2020-12-04
(MODULES-10887) Set `use_servername_for_filenames` for defaults

towo:
The default vhost definitions are called without
`use_servername_for_filenames` being set, which leads to a deprecation
warning on every run.

With this PR, the parameters are added to the default entries and thus
avoid generating said error message.

This was verified on a local installation.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -826,33 +826,35 @@ class apache (
     }
 
     ::apache::vhost { 'default':
-      ensure          => $default_vhost_ensure,
-      port            => '80',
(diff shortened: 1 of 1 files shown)


towo:
Force pushed to fix a typo in the commit message.

sanfrancrisko (approved):
Thanks for the improvement @towo 

I've raised [IAC-1288](https://tickets.puppetlabs.com/browse/IAC-1288) to improve this logging, based on your feedback and this PR.

PR #2104, opened 2020-12-07
(fix) Convert unnecessary multi line warnings to single lines

rj667:
Single line warnings are better readable and greppable from server logs.
Fixes https://tickets.puppetlabs.com/browse/MODULES-8518


sanfrancrisko (approved):
Thanks @rj667 - this seems like a sensible change to make, indeed 👍

PR #2109, opened 2020-12-16
Update to puppet-module-gems 1.0, pdk-templates and new rubocop

DavidS:


.travis.yml
```diff
@@ -27,89 +27,81 @@ stages:
 jobs:
   fast_finish: true
   include:
-    -
+    - bundler_args: --with system_tests
```
ekohl, line 30: I thought `--with` as a parameter was deprecated. I've started to set it as an env var. That way you don't have to repeat it either:
```yaml
env:
  BUNDLE_WITH: system_tests
```

DavidS: neat

DavidS: Created https://github.com/puppetlabs/pdk-templates/pull/381 to implement that across all modules. Only the duplication doesn't go away, because we have cells both ways. since it is through pdk-templates :shrug: it's not a big deal, and the environment var is nice because it is really "global" for all bundler commands in that run.

sanfrancrisko (approved):

PR #2110, opened 2020-12-22
Compare with major release version

ekohl:
CentOS 8 stream doesn't have a minor version so this check likely fails. However, it's irrelevant here and we can use the major release version fact instead.

There are a lot more odd cases with facts, but I don't want to touch too much code.

### DIFF
--- a/manifests/default_mods.pp
+++ b/manifests/default_mods.pp
@@ -17,7 +17,7 @@ class apache::default_mods (
       if versioncmp($apache_version, '2.4') >= 0 {
         # Lets fork it
         # Do not try to load mod_systemd on RHEL/CentOS 6 SCL.
-        if ( !($::osfamily == 'redhat' and versioncmp($::operatingsystemrelease, '7.0') == -1) and !($::operatingsystem == 'Amazon') ) {
+        if ( !($::osfamily == 'redhat' and versioncmp($::operatingsystemmajrelease, '7') == -1) and !($::operatingsystem == 'Amazon') ) {
           if ($use_systemd) {
             ::apache::mod { 'systemd': }
           }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -551,7 +551,7 @@ class apache (
     default => '(event|itk|prefork|worker)'
   }
 
-  if $::osfamily == 'RedHat' and $apache::version::distrelease == '7' {
+  if $::osfamily == 'RedHat' and $facts['operatingsystemmajrelease'] == '7' {
     # On redhat 7 the ssl.conf lives in /etc/httpd/conf.d (the confd_dir)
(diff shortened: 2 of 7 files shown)


sanfrancrisko (changes requested):
Thanks for the tweak @ekohl - it appears as though we have test failures on Travis due to the `versioncmp` being passed an undefined variable: https://travis-ci.com/github/puppetlabs/puppetlabs-apache/jobs/464049567

ekohl:
That's probably because the module doesn't use rspec-puppet-facts but rather specifies them manually. I had a look at the module, but its test infra is in serious need of an update.

sanfrancrisko:
> That's probably because the module doesn't use rspec-puppet-facts but rather specifies them manually. I had a look at the module, but its test infra is in serious need of an update.

Agreed - it would make things a lot nicer to use `rspec-puppet-facts` consistently across the module. I took a look at `spec/classes/apache_spec.rb` and frankly, it's going to be quite a while to fix that one test class up - it's not going to be as simple as encapsulating the tests within a `on_supported_os.each` block (_not that I ever thought it would!_). There's a heck of a lot of duplication and specific tests coupled to specific OSs. I haven't even looked in to the other test classes failing either - I suspect a similar scenario awaits in them.

I'm not saying that it isn't the right thing to do, but, it would be quite an undertaking and a bit more surgical than a bulk edit. I'll discuss with some of the team on how we want to proceed. If you have any suggestions, we'd be keen to hear, of course 👍

ekohl:
I added another commit I wrote while looking into this. I think the main pain point comes from remaining compatible with Facter 1.6 for too long. https://github.com/puppetlabs/puppetlabs-stdlib/pull/1154 is another thing that should have died long ago with Facter 1.6.

About the conversion to rspec-puppet-facts. I think you're right that this won't be a trivial conversion. Another thing to note is that the test runtime will go up a lot since it'll test a lot more combinations. Perhaps a decent compromise is to define a default test OS in some shared file and use that instead of defining the same facts over and over again.

ekohl:
In https://github.com/puppetlabs/puppetlabs-apache/pull/2120 I took a stab at a the preparation. Not a full conversion to rspec-puppet-facts, but by using shared examples a big step forward is taken to make it easier.

ekohl:
This now includes https://github.com/puppetlabs/puppetlabs-apache/pull/2120 since hopefully that's sufficient to make the tests pass. Once that's merged, I'll rebase for easier review.

ekohl:
Rebased and squashed into a single commit. Also updated some fact sets to provide a major version.

sanfrancrisko (approved):

david22swan (approved):
LGTM

david22swan:
@ekohl Thanks for the excellent work as always
Looking forward to more in the future.

PR #2112, opened 2021-01-05
Use the correct matches in rspec

ekohl:
Rather than checking the notify/require properties, it uses the matchers that check the dependency graph. That means it can also check transitive requires/notifies.

It also uses compile and compile.and_raise_error as a shorter way to check for errors (or the absense of them).

This is just a tiny improvement. The specs need a major update since a lot of the examples are using hardcoded facts of EOL OSes.

### DIFF
--- a/spec/classes/mod/php_spec.rb
+++ b/spec/classes/mod/php_spec.rb
@@ -292,7 +292,7 @@ describe 'apache::mod::php', type: :class do
           end
 
           it 'raises an error' do
-            expect { expect(subject).to contain_apache__mod('php5') }.to raise_error Puppet::Error, %r{mpm_module => 'prefork' or mpm_module => 'itk'}
+            is_expected.to compile.and_raise_error(%r{mpm_module => 'prefork' or mpm_module => 'itk'})
           end
         end
       end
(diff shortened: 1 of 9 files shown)


david22swan:
@ekohl 
There where some minor Rubocop errors in your code and so I pushed up a quick fix for them in order to save time.
Apologies for any offense

spec/classes/mod/php_spec.rb
```diff
@@ -290,7 +290,7 @@
           end
 
           it 'raises an error' do
-            expect { expect(subject).to contain_apache__mod('php5') }.to raise_error Puppet::Error, %r{mpm_module => 'prefork' or mpm_module => 'itk'}
+            is_expected.to compile.and_raise_error(Puppet::Error, %r{mpm_module => 'prefork' or mpm_module => 'itk'})
```
ekohl, line 293: Note to self:
```suggestion
            is_expected.to compile.and_raise_error(%r{mpm_module => 'prefork' or mpm_module => 'itk'})
```

ekohl:
> There where some minor Rubocop errors in your code and so I pushed up a quick fix for them in order to save time.
> Apologies for any offense

That's great. If it gets things merged quicker I'm happy to.

david22swan:
@ekohl
Your changes have got a few failures that need to be looked at.
But overall things look good.

ekohl:
There are acceptance test failures but I fail to see how they can be caused by this.

david22swan:
@ekohl After a few rekicks everything looks green so I feel happy to merge.
Thanks for putting in the work :)

PR #2113, opened 2021-01-11
(MAINT) Perform pdk update

sanfrancrisko:
This was primarily done to add the following line to the Rakefile:
`FastGettext.default_text_domain = 'default-text-domain'` and
prevent syntax checks failing because of the text domain is not
set (which generates a return code of 1).

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
     - name: "Honeycomb: Start recording"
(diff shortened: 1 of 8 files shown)


DavidS:
see also https://github.com/puppetlabs/puppetlabs-apache/pull/2109 ?

sanfrancrisko:
> see also #2109 ?

Yeah, I forgot about that PR - I'll get it over the line first

PR #2119, opened 2021-02-05
Fix issues when using php8

cdamage:
PHP8 dropped the conventions of including the major versions in apache modules. This fix makes it so that moving forward the module will no longer add the major in the relevant places. 

before change:
/etc/httpd/conf.modules.d/php8.conf
/etc/httpd/conf.modules.d/php8.load
^---Content---^
LoadModule php8_module modules/libphp8.so

Apache won't load

After change:
/etc/httpd/conf.modules.d/php.conf
/etc/httpd/conf.modules.d/php.load
^---Content---^
LoadModule php_module modules/libphp.so

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -17,7 +17,11 @@ class apache::mod::php (
   $libphp_prefix    = 'libphp'
 ) inherits apache::params {
   include apache
-  $mod = "php${php_version}"
+  if $php_version <= '7' {
+    $mod = "php${php_version}"
+  } else {
+    $mod = "php"
+  }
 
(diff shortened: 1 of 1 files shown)


sanfrancrisko:
Thanks for getting the fix put together @cdamage . I made a few tweaks to address a few issues and incorporated your changes with them in https://github.com/puppetlabs/puppetlabs-apache/pull/2121 - if you're happy with it, I'll get it merged and a new module release turned around a.s.a.p.

cdamage:
All good with me @sanfrancrisko. Many thanks.

sanfrancrisko:
This work was merged as part of #2121

PR #2120, opened 2021-02-06
(MAINT) Use shared examples for facts + other spec cleanups

ekohl:
This extracts common sets of facts to shared examples to reduce duplication. It also makes it easier to actually provide correct OS facts and use modern facts.

It also has 2 other commits that slightly improve the specs.

I realize this is a large PR and that's why I haven't even started with making actual changes. It should have a net effect that it's the same coverage but with less code duplication. It's a preparation for https://github.com/puppetlabs/puppetlabs-apache/pull/2110.

### DIFF
--- a/spec/classes/apache_spec.rb
+++ b/spec/classes/apache_spec.rb
@@ -4,18 +4,7 @@ require 'spec_helper'
 
 describe 'apache', type: :class do
   context 'on a Debian OS' do
-    let :facts do
-      {
-        id: 'root',
-        kernel: 'Linux',
-        lsbdistcodename: 'squeeze',
-        osfamily: 'Debian',
-        operatingsystem: 'Debian',
-        operatingsystemrelease: '6',
-        path: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
-        is_pe: false,
-      }
-    end
+    include_examples 'Debian 6'
 
     it { is_expected.to contain_class('apache::params') }
     it {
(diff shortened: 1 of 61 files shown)


ekohl:
I don't see why this would fail on acceptance tests. I changed nothing in those nor in manifests.

sanfrancrisko:
> I don't see why this would fail on acceptance tests. I changed nothing in those nor in manifests.

Thanks for embarking on this campaign - it'll definitely makes the tests a lot more legible and easier to maintain. 😃 

We've still got a few robustness issues with test execution, but we're [constantly chipping away at those](https://github.com/puppetlabs/provision_service/milestones). I think you may have been unlucky with some network issues, etc...

I've given this a first pass - will do another one too whilst a re-run of these tests goes, but, as you have already hinted at, none of your changes should have any effect on existing tests.

ekohl:
> We've still got a few robustness issues with test execution, but we're [constantly chipping away at those](https://github.com/puppetlabs/provision_service/milestones). I think you may have been unlucky with some network issues, etc...

Looks like provision_service is a private repository so I can't see any milestones.

sanfrancrisko:
> > We've still got a few robustness issues with test execution, but we're [constantly chipping away at those](https://github.com/puppetlabs/provision_service/milestones). I think you may have been unlucky with some network issues, etc...
> 
> Looks like provision_service is a private repository so I can't see any milestones.

D'oh, sorry....yes it is. My bad.

The TL;DR - it's the service that is responsible for provisioning the test environment running on the PRs via Github runners

sanfrancrisko (approved):

PR #2121, opened 2021-02-08
(MODULES-10899) Handle PHP8 MOD package naming convention changes

sanfrancrisko:
This PR incorporates the changes from https://github.com/puppetlabs/puppetlabs-apache/pull/2119 and https://github.com/puppetlabs/puppetlabs-apache/pull/2124 along with a few other additions:

- Change range of `versioncmp` to `< 8` to reduce churn in previous conventions for `7.x` PHP versions
- Enable and update acceptance test expectations for PHP mod packages on SLES 15. In hindsight, I'm not sure this was required after all, but I wanted to catch any potential issues around the changes being performed given SLES is a bit special with it's MOD PHP package names.
- Update test expectations for OSs using PHP >= 7 based on the new naming conventions in

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -16,7 +16,7 @@ class apache::mod::php (
   $libphp_prefix    = 'libphp'
 ) inherits apache::params {
   include apache
-  if (versioncmp($php_version, '7') <= 0) {
+  if (versioncmp($php_version, '8') < 0) {
     $mod = "php${php_version}"
   } else {
     $mod = 'php'
@@ -78,7 +78,7 @@ class apache::mod::php (
       id             => "php${_php_major}_module",
       path           => "${apache::lib_path}/mod_${mod}.so",
     }
-  } elsif (versioncmp($php_version, '7') <= 0) {
+  } elsif (versioncmp($php_version, '8') < 0) {
     ::apache::mod { $mod:
       package        => $_package_name,
       package_ensure => $package_ensure,
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -3,7 +3,6 @@
 # 
 # @todo
 #   Add docs
-# @note Unsupported platforms: SLES: all
 class apache::mod::php (
   $package_name     = undef,
   $package_ensure   = 'present',
@@ -17,7 +16,7 @@ class apache::mod::php (
   $libphp_prefix    = 'libphp'
 ) inherits apache::params {
   include apache
-  if $php_version <= '7' {
+  if (versioncmp($php_version, '7') <= 0) {
     $mod = "php${php_version}"
   } else {
     $mod = "php"
@@ -79,7 +78,7 @@ class apache::mod::php (
       id             => "php${_php_major}_module",
       path           => "${apache::lib_path}/mod_${mod}.so",
     }
-  } elsif $php_version <= '7' {
+  } elsif (versioncmp($php_version, '7') <= 0) {
     ::apache::mod { $mod:
       package        => $_package_name,
       package_ensure => $package_ensure,
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -69,30 +69,27 @@ class apache::mod::php (
     # Controls php version and libphp prefix
     $_lib = "${libphp_prefix}${php_version}.so"
   }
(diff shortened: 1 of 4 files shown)


gguillotte (approved):
@sanfrancrisko LGTM

beatchristen:
I've tried out this branch with debian10 and [libapache2-mod-php8.0](https://packages.debian.org/sid/amd64/libapache2-mod-php8.0/filelist) and the module name in `/etc/apache2/mods-enabled/php.load` now points at nothing:

`Syntax error on line 1 of /etc/apache2/mods-enabled/php.load: Cannot load /usr/lib/apache2/modules/libphp.so into server: /usr/lib/apache2/modules/libphp.so: cannot open shared object file: No such file or directory`

overriding the `$path` like below fixes it (`$php_version = '8.0'`:
```
  class { 'apache::mod::php':
    path        => "/usr/lib/apache2/modules/libphp${webserver::php_version}.so",
    php_version => $webserver::php_version,
    require     => Package["libapache2-mod-php${webserver::php_version}"],
  }
```

gguillotte:
Worth looking into. Notably for reproduction purposes, those php8 packages are only in sid/unstable at the moment.

gguillotte:
Re: https://php.watch/versions/8.0/mod_php-rename, the version number is removed from `php_module` but not from the `libphp` filename itself:

```diff
- LoadModule php7_module "/usr/lib/apache2/modules/libphp7.4.so"
+ LoadModule php_module "/usr/lib/apache2/modules/libphp8.0.so"
```

So for:

https://github.com/puppetlabs/puppetlabs-apache/blob/721878a05d2d456ff1f30bc062461e983676fae4/manifests/mod/php.pp#L73-L97 

L93 may need to use

```
       lib            => $_lib, 
```

instead of

```
       lib            => "${libphp_prefix}.so", 
```

sanfrancrisko:
Thanks for highlighting @beatchristen and the feedback and suggestions @gguillotte .

@beatchristen I did a quick test with the module, as it is right now and it appears as though the same problem exists - can you confirm?

@gguillotte L93, that you highlighted, should only be hit in cases where `$php_version < 8`, which is the legacy behaviour. I'd be concerned that changing that would have knock on effects to existing versions.

For reference, this was the result when using `a2enmod`, with `sid (unstable)` at the moment:

`/etc/apache2/mods-enabled/php8.0.load`:
```
# Conflicts: php5
# Depends: mpm_prefork
LoadModule php_module /usr/lib/apache2/modules/libphp8.0.so
```

I'm not sure what Debian are going to do when PHP 8 moves from unstable -> stable in terms of package naming convention. 

The original issue reports that this was affecting CentOS 7 and Debian 10 - I was going to suggest we raise this as a separate issue to fix on Debian 10 (using Sid) but if the customer in the original issue is using Debian 10 perhaps we need to address it now @gguillotte ?

beatchristen:
@sanfrancrisko not sure what problem you specifically encountered. IMHO, the module up until 7.x works as expected on Debian10.

I've added two tests and fixes for the problems reported on Debian 10 here:
https://github.com/puppetlabs/puppetlabs-apache/pull/2124

> I'm not sure what Debian are going to do when PHP 8 moves from unstable -> stable in terms of package naming convention.

There's probably no reason for Debian 11 to change anything about the structure within the file, as it was initiated further upstream with the PHP team itself. 
So when SLES get's a PHP 8.0 version, it should work once it's added to apache::version.

sanfrancrisko:
@beatchristen I was referring to the issue of trying to install PHP 8.0 using the sid repo on Debian 10 affecting the module as-is on the Forge right now (i.e. without this change merged).

There's a bit of time pressure on this fix for [MODULES-10899](https://tickets.puppetlabs.com/browse/MODULES-10899) so I was going to get PR merged and then we could continue investigating the issues with PHP 8.0 on Debian 10 on a separate PR.

And yes, to clarify, I can also confirm the module is working fine up to PHP 7.x with Debian 10 👍 

Thanks for putting up #2124 - looks good but just had one comment. I'll likely try to land this PR today, as mentioned, to address [MODULES-10899](https://tickets.puppetlabs.com/browse/MODULES-10899) - as far as I can see it will not cause any breaking changes for the module so far, but by all means scream if I'm missing something 😄

PR #2122, opened 2021-02-13
(IAC-1186) Add $use_port_for_filenames parameter

smortex:
When using $use_servername_for_filenames = true, virtual hosts with the
same $servername but a different $port will introduce a duplicate
resource declaration.

Introduce a $use_port_for_filenames that act in concert with
$use_servername_for_filenames to avoid this situation.


smortex:
This is a follow-up to #2086, so ping @sanfrancrisko and @sheenaajay who worked on this previous PR and @DavidS who proposed this change in IAC-1186.

sanfrancrisko (approved):
Apologies, I was a bit less involved in the conversation on IAC-1186 than I wanted to (was out on training most of last week). 

Thanks for getting the enhancement up 👍

PR #2125, opened 2021-02-15
pdksync - Remove Puppet 5 from testing and bump minimal version to 6.0.0

carabasdaniel:
Remove Puppet 5 from testing and bump minimal version to 6.0.0
pdk version: `1.19.0.pre (47)`

### DIFF
--- a/.github/workflows/auto_release.yml
+++ b/.github/workflows/auto_release.yml
@@ -46,7 +46,7 @@ jobs:
       if: ${{ github.repository_owner == 'puppetlabs' }}
       id: gv
       run: |
-        echo "::set-output name=ver::$(cat metadata.json | jq .version | tr -d \")"
+        echo "::set-output name=ver::$(jq --raw-output .version metadata.json)"
 
     - name: "Commit changes"
       if: ${{ github.repository_owner == 'puppetlabs' }}
--- a/.github/workflows/auto_release.yml
+++ b/.github/workflows/auto_release.yml
@@ -51,7 +51,7 @@ jobs:
     - name: "Commit changes"
       if: ${{ github.repository_owner == 'puppetlabs' }}
       run: |
-        git config --local user.email "action@github.com"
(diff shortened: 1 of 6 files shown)


.github/workflows/auto_release.yml
```diff
@@ -0,0 +1,81 @@
...
+      with:
+        args: 'release prep --force'
+      env:
+        CHANGELOG_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: "Get Version"
+      if: ${{ github.repository_owner == 'puppetlabs' }}
+      id: gv
+      run: |
+        echo "::set-output name=ver::$(cat metadata.json | jq .version | tr -d \")"
```
ekohl, line 49: Probably needs to be fixed in pdk-templates, but there's a oneliner for this
```suggestion
        echo "::set-output name=ver::$(jq --raw-output .version metadata.json)"
```

.github/workflows/auto_release.yml
```diff
@@ -0,0 +1,81 @@
...
+
+    - name: "Commit changes"
+      if: ${{ github.repository_owner == 'puppetlabs' }}
+      run: |
+        git config --local user.email "action@github.com"
```
ekohl, line 54: Is it really a good idea to claim some email address `@github.com`? For anonymous commits they use `USER@users.noreply.github.com` so I'd suggest:
```suggestion
        git config --local user.email "${{ github.repository_owner }}@users.noreply.github.com"
```

.github/workflows/auto_release.yml
pmcmaw: This has been updated in the pdk-templates: 
https://github.com/puppetlabs/pdk-templates/pull/409

.github/workflows/auto_release.yml
pmcmaw: This has been updated in the pdk-templates: 
https://github.com/puppetlabs/pdk-templates/pull/409

PR #2128, opened 2021-02-18
Opv

DavidS:
Identifying the changes wrt opv

1. Test Failures 

- Host not reachable (fixed by updating hosts file)
- Time Out issue( Port not open, fixed by opening the ports)
-  Connection refused error for vhost setup( ssl certs/keys missing and due to incorrect setup the apache service is not starting and causing the tests to fail)
Solution is to generate the ssl certs/keys and place them to /tmp/ssl_key /tmp/ssl_cert . Exploring the existing modules available to manage ssl keys
https://forge.puppet.com/modules/camptocamp/openssl

2. Setup requirements
3. Workaround required to test opv

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2381,10 +2381,15 @@ define apache::vhost (
   }
 
   unless $servername in ['default', 'default-ssl'] {
-    if $port {
-      $check_http_url = "$servername:$port"
+    if $ip {
+      $check_http_server = $ip
     } else {
-      $check_http_url = $servername
+      $check_http_server = $servername
+    }
+    if $port {
+      $check_http_url = "$check_http_server:$port"
+    } else {
+      $check_http_url = $check_http_server
     }
     if ($ssl) {
       check_http { "https://${check_http_url}": }
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2380,6 +2380,12 @@ define apache::vhost (
     }
   }
 
+  if ($ssl) {
+    check_http { "https://${servername}": }
+  } else {
+    check_http { "http://${servername}": }
+  }
+
   concat { "${priority_real}${filename}.conf":
     ensure  => $ensure,
     path    => "${apache::vhost_dir}/${priority_real}${filename}.conf",
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2380,7 +2380,7 @@ define apache::vhost (
     }
   }
 
-  if $servername != 'default' and $servername != 'default-ssl' {
+  unless $servername in ['default', 'default-ssl'] {
     if $port {
       $check_http_url = "$servername:$port"
     } else {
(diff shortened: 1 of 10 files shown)


manifests/vhost.pp
```diff
@@ -2360,6 +2360,12 @@
     }
   }
 
+  if ($ssl) {
+    check_http { "https://${servername}": }
+  } else {
+    check_http { "http://${servername}": }
```
ekohl, line 2366: You should take the port(s) into account as well.

ekohl (commented):
Interesting idea. I wonder if you can expose sufficient metadata to also execute the checks from a different host.

Also makes me think about a more generic thing than https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/lib/puppet/util/postgresql_validator.rb. There's equivalent versions out there for other databases. If you can provide an easy mechanism to implement such a type/provider that'd be nice. Especially if it behaves well with noop.

manifests/vhost.pp
DavidS: does this look like I'm finished yet? :rofl:

DavidS:
> Interesting idea. I wonder if you can expose sufficient metadata to also execute the checks from a different host.
> 
> Also makes me think about a more generic thing than https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/lib/puppet/util/postgresql_validator.rb. There's equivalent versions out there for other databases. If you can provide an easy mechanism to implement such a type/provider that'd be nice. Especially if it behaves well with noop.

How do you like https://github.com/puppetlabs/opv/blob/main/lib/puppet/provider/check_http/check_http.rb with regards to simplicity? that works fine with noop, etc.

ekohl:
That looks simple enough. I'm sure at some point it will become complicated when people need to use proxies etc, but I like the concept. For many things the systemd ready protocol is already a good start to ensure it's up (compared to sysvinit where a service may return exit code 0 but not really start) but this is a good layer on top of it. Apache is a prime example where the service may run, but the actual service may not be running correctly.

A common thing in monitoring is also more explicit checks, like a status URL should return a certain HTTP code and/or some content. Probably a nice next step.

manifests/vhost.pp
```diff
@@ -2380,6 +2380,19 @@
     }
   }
 
+  if $servername != 'default' and $servername != 'default-ssl' {
```
bastelfreak, line 2383: maybe that's a bit better to read?
```suggestion
  unless $servername in ['default', 'default-ssl'] {
```

manifests/vhost.pp
```diff
@@ -2380,6 +2380,19 @@
...
+    if $port {
+      $check_http_url = "$servername:$port"
+    } else {
+      $check_http_url = $servername
+    }
+    if ($ssl) {
+        check_http { "https://${check_http_url}": }
```
bastelfreak, line 2390: I think the indentation is off by two whitespaces here

sheenaajay: Thank you @bastelfreak for the quick review.

PR #2129, opened 2021-02-22
Drop Ubuntu 10.04 compatibility code

ekohl:
This is long End of Life and should not be supported.


ekohl:
Last time I checked with legal I was not allowed to sign the Puppet CLA, which is why I never contributed anything to puppet.git. If this will be enforced in modules as well I'll have to stop contributing.

daianamezdrea:
Hi @ekohl, thank you for your contribution! I'll check now with the team and our manager @nheaney18  how we can approach the situation and what we can do.

david22swan:
@ekohl
Thank's for your work, have gone ahead and merged it.
Hope to see a lot more from you in the future.

PR #2135, opened 2021-03-01
Fix dav_svn for Debian 10

martijndegouw:
As far as I can tell Debian 10 still needs this work-around.

### DIFF
--- a/manifests/mod/dav_svn.pp
+++ b/manifests/mod/dav_svn.pp
@@ -21,15 +21,11 @@ class apache::mod::dav_svn (
 
   ::apache::mod { 'dav_svn': }
 
-  if $::osfamily == 'Debian' and ! ($::operatingsystemmajrelease in ['6', '9', '16.04', '18.04']) {
-    $loadfile_name = undef
-  } else {
-    $loadfile_name = 'dav_svn_authz_svn.load'
-  }
-
   if $authz_svn_enabled {
     ::apache::mod { 'authz_svn':
-      loadfile_name => $loadfile_name,
+      # authz_svn depends on symbols from the dav_svn module,
+      # therefore, make sure authz_svn is loaded after dav_svn.
+      loadfile_name => 'dav_svn_authz_svn.load',
       require       => Apache::Mod['dav_svn'],
     }
   }
--- a/spec/acceptance/mod_dav_svn_spec.rb
+++ b/spec/acceptance/mod_dav_svn_spec.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'apache::mod::dav_svn class' do
+  context 'dav_svn module with authz_svn disabled' do
+    pp = <<-MANIFEST
+      class { 'apache': }
+      class { 'apache::mod::dav_svn':
+        authz_svn_enabled => false,
+      }
+    MANIFEST
+
+    it 'applies with no errors' do
+      apply_manifest(pp, catch_failures: true)
+    end
+
+    it 'applies a second time without changes' do
+      apply_manifest(pp, catch_changes: true)
+    end
+  end
+
+  context 'dav_svn module with authz_svn enabled' do
+    pp = <<-MANIFEST
+      class { 'apache': }
+      class { 'apache::mod::dav_svn':
+        authz_svn_enabled => true,
+      }
+    MANIFEST
+
+    it 'applies with no errors' do
+      apply_manifest(pp, catch_failures: true)
+    end
+
+    it 'applies a second time without changes' do
+      apply_manifest(pp, catch_changes: true)
+    end
+  end
+end
--- a/spec/classes/mod/dav_svn_spec.rb
+++ b/spec/classes/mod/dav_svn_spec.rb
@@ -24,7 +24,7 @@ describe 'apache::mod::dav_svn', type: :class do
         it { is_expected.to contain_apache__mod('dav_svn') }
         it { is_expected.to contain_package('libapache2-svn') }
         it { is_expected.to contain_apache__mod('authz_svn') }
-        it { is_expected.to contain_file('authz_svn.load').with_content(%r{LoadModule authz_svn_module}) }
+        it { is_expected.to contain_file('dav_svn_authz_svn.load').with_content(%r{LoadModule authz_svn_module}) }
       end
     end
     context 'on a RedHat OS' do


manifests/mod/dav_svn.pp
```diff
@@ -21,7 +21,7 @@
 
   ::apache::mod { 'dav_svn': }
 
-  if $::osfamily == 'Debian' and ! ($::operatingsystemmajrelease in ['6', '9', '16.04', '18.04']) {
+  if $::osfamily == 'Debian' and ! ($::operatingsystemmajrelease in ['6', '9', '10', '16.04', '18.04']) {
```
ekohl, line 24: This feels like a workaround on some older Debian versions while the newer ones behave as you would expect. Perhaps it should be checked which versions need this. I'd expect Ubuntu 20.04 behaves the same as Debian 10.

Then you would get to something like:
```suggestion
  if $::osfamily == 'Debian' and $::operatingsystemmajrelease in ['7', '8', '14.04'] {
```

Untested, but I think you get the idea.

martijndegouw: Yeah I got it. I also did found some patches from Debian for the dav_svn module which should fix this in the package itself, but the issue still reproduces on my Debian 10 testing machine.

martijndegouw: I just checked. The Debian 10 package has pretty similar word-around in mods-available/dav_svn.load, but that file is overwritten by this puppet module.

ekohl: I must admit I haven't looked at the dav_svn module for ages. Perhaps you can write an acceptance test for the module so we know it really works rather than guessing? That also gives the best chance against regressions.

martijndegouw: I will give a shot, It's not hard to reproduce since apache2 simply won't start when including apache::mod::dav_svn :D

ekohl: Exactly, those are the things that are fairly easy to test.

martijndegouw: Given the result of the acceptance test and looking at how Debian implemented their work-around it's safe to say that the whole version checking part of that if statement can be removed. Agreed?

spec/acceptance/mod_dav_svn_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    pp = <<-MANIFEST
+      class { 'apache': }
+      class { 'apache::mod::dav_svn':
+        authz_svn_enabled => true,
+      }
+    MANIFEST
+
+    it 'succeeds in installing the dav_svn module' do
+      apply_manifest(pp, catch_failures: true)
+    end
```
ekohl, line 16: Could you add an idempotency check to this?

manifests/mod/dav_svn.pp
ekohl: I think that makes sense but it's probably good to check with the actual maintainers.

spec/acceptance/mod_dav_svn_spec.rb
martijndegouw: I’m not sure what you mean by this. Do you want to run the apply_manifest twice? To check if it fails/passes 2 times in a row?

ekohl: Copied from another module:
```ruby
  it 'applies with no errors' do
  end

  it 'applies a second time without changes' do
    apply_manifest(pp, catch_changes: true)
  end
```

martijndegouw:
Great, SLES 15 test is failing, on the exact same issue I have on my SLES server :/.
`Note: Signing data enables the recipient to verify that no modifications occurred after the data were signed. Accepting data with no, wrong or unknown signature can lead to a corrupted system and in extreme cases even to a system compromise.
Note: File 'repomd.xml' is the repositories master index file. It ensures the integrity of the whole repo.

Warning: This file was modified after it has been signed. This may have been a malicious change, so it might not be trustworthy anymore! You should not continue unless you know it's safe.

Signature verification failed for file 'repomd.xml' from repository 'Puppet 6 Nightly Repository sles 15 - x86_64'
`

manifests/mod/dav_svn.pp
sanfrancrisko: Thanks for the fix @martijndegouw and the assistance during the PR @ekohl . The test is great - allows us to check we haven't inadvertently stomp over some logic put in there for the older versions, as you said @ekohl .

sanfrancrisko:
> Great, SLES 15 test is failing, on the exact same issue I have on my SLES server with the puppetlabs repo :/.
> `Note: Signing data enables the recipient to verify that no modifications occurred after the data were signed. Accepting data with no, wrong or unknown signature can lead to a corrupted system and in extreme cases even to a system compromise.
> Note: File 'repomd.xml' is the repositories master index file. It ensures the integrity of the whole repo.
> 
> Warning: This file was modified after it has been signed. This may have been a malicious change, so it might not be trustworthy anymore! You should not continue unless you know it's safe.
> 
> Signature verification failed for file 'repomd.xml' from repository 'Puppet 6 Nightly Repository sles 15 - x86_64'
> `

😞  Would this be a consistently reproducible error? Re-running seems to have gotten me past that issue now, but, if you're experiencing that pretty consistently, I'd like to get the issue raised with our Agent team to look in to.

sanfrancrisko:
PR looks good, but there seems to be a [unit test failure](https://travis-ci.com/github/puppetlabs/puppetlabs-apache/jobs/487770533#L11884). I can't see why we wouldn't still generate a catalog with [this File resource](https://github.com/puppetlabs/puppetlabs-apache/blob/main/spec/classes/mod/dav_svn_spec.rb#L27) when your change is applied. Will have another look at it but just highlighting it here in case yourself or @ekohl figure it out before me.

martijndegouw:
> 
> 
> PR looks good, but there seems to be a [unit test failure](https://travis-ci.com/github/puppetlabs/puppetlabs-apache/jobs/487770533#L11884). I can't see why we wouldn't still generate a catalog with [this File resource](https://github.com/puppetlabs/puppetlabs-apache/blob/main/spec/classes/mod/dav_svn_spec.rb#L27) when your change is applied. Will have another look at it but just highlighting it here in case yourself or @ekohl figure it out before me.

Sorry, I misted that one. I've updated the unit test as well.

david22swan:
@martijndegouw This PR looks good to me, thanks for putting in the work.

PR #2138, opened 2021-03-11
Fix HEADER* and README* wildcards in IndexIgnore

keto:
The original intent of HEADER* and README* in the IndexIgnore was
probably to ignore the files that are included at the top (HeaderName) and
bottom (ReadmeName) of the index listing. But the wildcard will hide other
files, like README.txt, so use full filenames instead.


carabasdaniel:
Hi @keto,
Thank you for your contribution, this looks good 👍

PR #2150, opened 2021-05-11
support for uri for severname with use_servername_for_filenames

Zarne:
in the apache documentation https://httpd.apache.org/docs/2.4/fr/mod/core.html#servername
```Syntax:	ServerName [scheme://]domain-name|ip-address[:port]```
add support for this syntax when use_servername_for_filenames is set to true.


pmcmaw:
Thank you for taking the time to contribute to our supported module. 
Great change and test coverage, again thank you :-)

PR #2154, opened 2021-05-26
Only warn about servername logging if relevant

ekohl:
The warning about $use_servername_for_filenames is long and may not affect users. If they use a normal $name, there would effectively be no difference. This compares $name to $normalized_servername. The warning is only relevant Only if they differ.

Technically there could be a space in $servername (since there is no data type to enforce validation) but Apache wouldn't accept that anyway.

It also fixes two typos in the warnings.

Please make sure my logic is correct here, I'm not 100% sure it is.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2111,7 +2111,7 @@ define apache::vhost (
     false => $name,
   }
 
-  if ! $use_servername_for_filenames {
+  if ! $use_servername_for_filenames and $name != $normalized_servername {
     $use_servername_for_filenames_warn_msg = '
     It is possible for the $name parameter to be defined with spaces in it. Although supported on POSIX systems, this
     can lead to cumbersome file names. The $servername attribute has stricter conditions from Apache (i.e. no spaces)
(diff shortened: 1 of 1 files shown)


david22swan:
@ekohl You seem to have a lot of test failures here, have you had a chance to look at them more closely?
Do you need ay assistance?

ekohl:
I think when I pushed this I was unable to run Puppet locally since the ecosystem was Ruby 3 incompatible. Right now https://github.com/rodjek/rspec-puppet/pull/830 still makes it harder than it needs to be. Also that there's no `puppet-module-posix-default-r3.0` is yet another reason I never use PDK. It's always late.

Looks like I had a syntax error which caused the whole tests to fail for obvious reasons.

What we do in Voxpupuli is a single test job that runs basic validations prior to running any other tests. It also determines the matrix. That saves a lot of CI resources on these trivial mistakes.

david22swan (approved):
LGTM

david22swan:
@ekohl Sorry for the trouble, but thank you for the work. This is the nice sort of change that only someone who actually work's with the module would think of, gonna go ahead and merge it.
Release containing it should be out later today.

Once again, thank you for putting in the effort :)

dhs-rec:
Hmm, I've just updated to 6.3.0, but I still see that warning although none of my vhosts have whitespace in their names.

PR #2156, opened 2021-06-07
pdksync - (maint) PDKSync - PDK Update

carabasdaniel:
(maint) PDKSync - PDK Update
pdk version: `2.1.0`

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -48,6 +48,18 @@ RSpec.configure do |c|
   c.after(:suite) do
     RSpec::Puppet::Coverage.report!(0)
   end
+
+  # Filter backtrace noise
+  backtrace_exclusion_patterns = [
+    %r{spec_helper},
+    %r{gems},
+  ]
+
+  if c.respond_to?(:backtrace_exclusion_patterns)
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -48,6 +48,18 @@
...
+  backtrace_exclusion_patterns = [
+    %r{spec_helper},
+    %r{gems},
+  ]
+
+  if c.respond_to?(:backtrace_exclusion_patterns)
+    c.backtrace_exclusion_patterns = backtrace_exclusion_patterns
+  elsif c.respond_to?(:backtrace_clean_patterns)
+    c.backtrace_clean_patterns = backtrace_exclusion_patterns
+  end
```
ekohl, line 62: Shouldn't this be contributed to https://github.com/puppetlabs/puppetlabs_spec_helper?

carabasdaniel: Hi @ekohl, 
For now we will keep these changes in the pdk-templates and if this works out nicely for the currently supported modules then we will probably port it to the puppetlabs_spec_helper and this change will be removed on a future pdk update.

michaeltlombardi (approved):

PR #2157, opened 2021-06-10
Add feature to reload apache service when content of ssl files has changed

timdeluxe:
This fixes issue MODULES-10847, which we also wanted to solve.

The solution with those helper files seems a bit odd, but the original ssl files could be managed by any means outside of the apache module or even outside of puppet, so this was the only way i could image to achieve it.

(Slightly offtopic: Let me add that this module gave me a love/hate relationship with spec/acceptance tests. First i blamed wrong tests to be the cause for failing, but in the end it was my own stupid code. Finally i could manage getting tests run green at my desktop - hope they will do in the modules CI setup as well.)

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -109,6 +109,7 @@
 * `apache::confd::no_accf`: Manages the `no-accf.conf` file.
 * `apache::default_confd_files`: Helper for setting up default conf.d files.
 * `apache::default_mods`: Installs and congfigures default mods for Apache
+* `apache::mod::ssl::reload`: Manages the puppet_ssl folder for ssl file copies, which is needed to track changes for reloading service on changes
 * `apache::package`: Installs an Apache MPM.
 * `apache::params`: This class manages Apache parameters
 * `apache::php`: This class installs PHP for Apache.
@@ -207,6 +208,7 @@ The following parameters are available in the `apache` class:
 * [`default_ssl_crl_path`](#default_ssl_crl_path)
 * [`default_ssl_crl_check`](#default_ssl_crl_check)
 * [`default_ssl_key`](#default_ssl_key)
+* [`default_ssl_reload_on_change`](#default_ssl_reload_on_change)
 * [`default_ssl_vhost`](#default_ssl_vhost)
 * [`default_type`](#default_type)
 * [`default_vhost`](#default_vhost)
@@ -433,6 +435,14 @@ this parameter with your SSL key's location before deploying this server in a pr
 environment.
 
 Default value: `$apache::params::default_ssl_key`
+
+##### <a name="default_ssl_reload_on_change"></a>`default_ssl_reload_on_change`
+
+Data type: `Boolean`
+
+Enable reloading of apache if the content of ssl files have changed.
+
+Default value: ``false``
 
 ##### <a name="default_ssl_vhost"></a>`default_ssl_vhost`
 
@@ -6369,6 +6379,7 @@ The following parameters are available in the `apache::mod::ssl` class:
 * [`ssl_stapling`](#ssl_stapling)
 * [`ssl_stapling_return_errors`](#ssl_stapling_return_errors)
 * [`ssl_mutex`](#ssl_mutex)
+* [`ssl_reload_on_change`](#ssl_reload_on_change)
 * [`apache_version`](#apache_version)
 * [`package_name`](#package_name)
 * [`ssl_sessiontickets`](#ssl_sessiontickets)
@@ -6524,6 +6535,14 @@ Default based on the OS and/or Apache version:
 - Debian/Ubuntu + Apache < 2.4: 'file:${APACHE_RUN_DIR}/ssl_mutex'.
 
 Default value: ``undef``
+
+##### <a name="ssl_reload_on_change"></a>`ssl_reload_on_change`
+
+Data type: `Boolean`
+
+Enable reloading of apache if the content of ssl files have changed. It only affects ssl files configured here and not vhost ones.
+
+Default value: ``false``
 
 ##### <a name="apache_version"></a>`apache_version`
 
@@ -7780,6 +7799,7 @@ The following parameters are available in the `apache::vhost` defined type:
 * [`ssl_stapling_timeout`](#ssl_stapling_timeout)
 * [`ssl_stapling_return_errors`](#ssl_stapling_return_errors)
 * [`ssl_user_name`](#ssl_user_name)
+* [`ssl_reload_on_change`](#ssl_reload_on_change)
 * [`use_canonical_name`](#use_canonical_name)
 * [`define`](#define)
 * [`auth_oidc`](#auth_oidc)
@@ -10628,6 +10648,14 @@ Data type: `Optional[String]`
 Sets the [SSLUserName](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslusername) directive.
 
 Default value: ``undef``
+
+##### <a name="ssl_reload_on_change"></a>`ssl_reload_on_change`
+
+Data type: `Boolean`
+
+Enable reloading of apache if the content of ssl files have changed.
+
+Default value: `$apache::default_ssl_reload_on_change`
 
 ##### <a name="use_canonical_name"></a>`use_canonical_name`
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -91,6 +91,9 @@
 #   this parameter with your SSL key's location before deploying this server in a production 
 #   environment.
 #
+# @param default_ssl_reload_on_change
+#   Enable reloading of apache if the content of ssl files have changed.
+#
 # @param default_ssl_vhost
 #   Configures a default SSL virtual host.
 #   If `true`, Puppet automatically configures the following virtual host using the 
@@ -472,6 +475,7 @@ class apache (
   $default_ssl_crl_path                                                 = undef,
   $default_ssl_crl                                                      = undef,
   $default_ssl_crl_check                                                = undef,
+  Boolean $default_ssl_reload_on_change                                 = false,
   $default_type                                                         = 'none',
   $dev_packages                                                         = $apache::params::dev_packages,
   $ip                                                                   = undef,
--- a/manifests/mod/ssl.pp
+++ b/manifests/mod/ssl.pp
@@ -62,6 +62,9 @@
 #   - Debian/Ubuntu + Apache >= 2.4: 'default'.
 #   - Debian/Ubuntu + Apache < 2.4: 'file:${APACHE_RUN_DIR}/ssl_mutex'.
 #
+# @param ssl_reload_on_change
+#   Enable reloading of apache if the content of ssl files have changed. It only affects ssl files configured here and not vhost ones.
+#
 # @param apache_version
 #   Used to verify that the Apache version you have requested is compatible with the module.
 #
@@ -97,6 +100,7 @@ class apache::mod::ssl (
   Optional[String] $stapling_cache                          = undef,
   Optional[Boolean] $ssl_stapling_return_errors             = undef,
   $ssl_mutex                                                = undef,
+  Boolean $ssl_reload_on_change                             = false,
   $apache_version                                           = undef,
   $package_name                                             = undef,
 ) inherits ::apache::params {
@@ -172,6 +176,24 @@ class apache::mod::ssl (
 
   if versioncmp($_apache_version, '2.4') >= 0 {
     include apache::mod::socache_shmcb
+  }
+
+  if $ssl_reload_on_change {
+    [$ssl_cert, $ssl_key, $ssl_ca].each |$ssl_file| {
+      if $ssl_file {
+        include apache::mod::ssl::reload
+        $_ssl_file_copy = regsubst($ssl_file, '/', '_', 'G')
+        file { $_ssl_file_copy:
+          path    => "${apache::params::puppet_ssl_dir}/${_ssl_file_copy}",
+          source  => "file://${ssl_file}",
+          owner   => 'root',
+          group   => $apache::params::root_group,
+          mode    => '0640',
+          seltype => 'cert_t',
+          notify  => Class['apache::service'],
+        }
+      }
+    }
   }
 
   # Template uses
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1679,6 +1679,9 @@
 # @param ssl_user_name
 #   Sets the [SSLUserName](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslusername) directive.
 #
+# @param ssl_reload_on_change
+#   Enable reloading of apache if the content of ssl files have changed.
+#
 # @param use_canonical_name
 #   Specifies whether to use the [`UseCanonicalName directive`](https://httpd.apache.org/docs/2.4/mod/core.html#usecanonicalname),
 #   which allows you to configure how the server determines it's own name and port.
@@ -1762,6 +1765,7 @@ define apache::vhost (
   $ssl_crl                                                                          = $apache::default_ssl_crl,
   $ssl_crl_check                                                                    = $apache::default_ssl_crl_check,
   $ssl_certs_dir                                                                    = $apache::params::ssl_certs_dir,
+  Boolean $ssl_reload_on_change                                                     = $apache::default_ssl_reload_on_change,
   $ssl_protocol                                                                     = undef,
   $ssl_cipher                                                                       = undef,
   Variant[Boolean, Enum['on', 'On', 'off', 'Off'], Undef] $ssl_honorcipherorder     = undef,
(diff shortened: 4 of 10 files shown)


timdeluxe:
Ah well, so many hours with tests and i oversaw the basics... 🙈 Please apologize!
Will fix and squash/rebase. No need to comment about this.

However if you have feedback to the feature itself, let me know.

REFERENCE.md
```diff
@@ -145,26 +179,92 @@ Apache server's or a virtual host's listening address and port.
...
-* [`a2mod`](#a2mod): Manage Apache 2 modules
-
 ### Functions
 
-* [`apache::apache_pw_hash`](#apacheapache_pw_hash): DEPRECATED.  Use the function [`apache::pw_hash`](#apachepw_hash) instead.
-* [`apache::bool2httpd`](#apachebool2httpd): Transform a supposed boolean to On or Off. Passes all other values through.
-* [`apache::pw_hash`](#apachepw_hash): Hashes a password in a format suitable for htpasswd files read by apache.
-* [`apache_pw_hash`](#apache_pw_hash): DEPRECATED.  Use the namespaced function [`apache::pw_hash`](#apachepw_hash) instead.
-* [`bool2httpd`](#bool2httpd): DEPRECATED.  Use the namespaced function [`apache::bool2httpd`](#apachebool2httpd) instead.
+* [`stdlib::ensure`](#stdlibensure): function to cast ensure parameter to resource specific value
```
ekohl, line 184: This looks wrong. As if it created a reference for all fixtures as well.

manifests/mod/ssl.pp
```diff
@@ -174,6 +178,46 @@
     include apache::mod::socache_shmcb
   }
 
+  file { $apache::params::puppet_ssl_dir:
```
ekohl, line 181: Why refer to `apache::params` instead of `apache`?

ekohl (changes requested):

REFERENCE.md
timdeluxe: Correct, my fault. Fixed it in the new push

manifests/mod/ssl.pp
timdeluxe: Because it is defined only there? It is NOT a parameter in init.pp.
Maybe i misunderstand something about variable scoping and you can explain?

ekohl: Technically `apache` inherits from `apache::params` so you can use `$apache::puppet_ssl_dir`. Then in the future you can easily make it a parameter.

I've seen that this confuses people and isn't really obvious to people.

manifests/init.pp
```diff
@@ -91,6 +91,9 @@
 #   this parameter with your SSL key's location before deploying this server in a production 
 #   environment.
 #
+# @param ssl_reload_on_change
```
ekohl, line 94: This doesn't match the parameter name.

manifests/mod/ssl.pp
```diff
@@ -97,6 +100,7 @@
   Optional[String] $stapling_cache                          = undef,
   Optional[Boolean] $ssl_stapling_return_errors             = undef,
   $ssl_mutex                                                = undef,
+  Boolean $ssl_reload_on_change                             = false,
```
ekohl, line 103: Should this default to undef and then use `pick($ssl_reload_on_change, $apache::ssl_reload_on_change)`?

ekohl (commented):
I personally like a minimal impact on environments. Can it only manage the directory if the feature is enabled? Perhaps even purge the whole directory if it's disabled?

timdeluxe:
> I personally like a minimal impact on environments. Can it only manage the directory if the feature is enabled? Perhaps even purge the whole directory if it's disabled?

@ekohl Before i answer to each of the code remarks, lets discuss this first, because those code remark discussions depend on that:

I struggled with exactly that question, but the reload thing should be a per-vhost-based thing. One should be able to have apache reloaded if the ssl files change for a vhost, but not for the other. At least that is my opinion. I need to manage the folder in ssl.pp then, because vhost.pp is a define and managing the folder there would lead to duplicate declarations. Also even if the reload is deactivated for default vhost and for the mod::ssl module, it still can be activated for a vhost and then the folder is required.

Or do we say that reload thing is a general setting and can't be changed on a per vhost base? (In that case your requested changes could be implemented)

ekohl:
You're right that we should take it to a higher design level.

So we have a few options. The current one obviously. We can also introduce a global parameter.

Another is introducing a private class (like `apache::mod::ssl::reload`) that does the directory setup. Then in `apache::vhost` you can include it if it does need the reload.

The benefit is that `mod_ssl` doesn't need any modification. It also keeps it "lazy" in that you only get the additional directories if you do use the feature. The downside is that you can't purge/clean it up.

timdeluxe:
A private class is indeed a good idea, will try that, thanks!

However mod_ssl does need at least the file resources for the ssl files, because one could work with apache without using vhosts at all (extremly rare, but the module is widely spread, so we should not forget those people).

manifests/init.pp
timdeluxe: Correct, my fault, fixed.

manifests/mod/ssl.pp
timdeluxe: No, that makes no sense, see also our further discussions

manifests/mod/ssl.pp
timdeluxe: I am still not convinced. I've seen a lot of other places, where apache::params::* variables are used.
It still can be changed, when it shall be come a parameter of the init class.
It would rather confuse me the other way around if i would use $apache::puppet_ssl_dir and don't have a parameter for that. I don't see the need for a parameter.

DavidS: I don't have an opinion on which direction to go here, but do understand that many parts of this module are older than what anyone would call "best practice" today.

timdeluxe: @DavidS Well, i think we then need a maintainer of this module to decide.

DavidS: When in doubt, I go with what's there in the codebase and the opinion of them who does the work, unless there is a critical overriding factor.

ekohl: >  It would rather confuse me the other way around if i would use $apache::puppet_ssl_dir and don't have a parameter for that. I don't see the need for a parameter.

That's what I've also seen from others. They assume it doesn't work because there's no parameter and forget the inheritance. Implementation wise it doesn't make a difference but code clarity would be a good reason to use `::params::`. I just wanted to highlight it is an option.

timdeluxe: Thanks for highlighting. I think i leave it as it is, unless the maintainers of the module want a change.

timdeluxe:
On real world usage of my branch i found two further issues, which i now fixed: The puppet_ssl_dir was managed before apache was even installed and resulted in missing httpd_dir, so management of puppet_ssl_dir failed. Also even if the default-ssl vhost was set to absent, it still managed its certificate copy - this is also now fixed.

This PR still says @ekohl requested 1 change, however i can only find resolved conversations. Can someone help me?

REFERENCE.md
```diff
@@ -145,27 +146,6 @@ Apache server's or a virtual host's listening address and port.
 * `apache::peruser::processor`: Enables the `Peruser` module for FreeBSD only.
 * `apache::security::rule_link`: Links the activated_rules from `apache::mod::security` to the respective CRS rules on disk.
 
-### Resource types
```
ekohl, line 148: These shouldn't be deleted because they're still present. Any idea why it's doing this?

david22swan:
@timdeluxe As stated above by @ekohl it looks like some of the reference has been deleted by your changes.
Could you get that added back for us? Otherwise we won't be able to merge.

REFERENCE.md
timdeluxe: Because i used puppet strings wrongly due to fixtures, which aren't ignored by it. I think i managed it now to have a correct reference

timdeluxe:
@david22swan Sorry, missed that. Is fixed now.

manifests/init.pp
```diff
@@ -472,6 +475,7 @@
   $default_ssl_crl_path                                                 = undef,
   $default_ssl_crl                                                      = undef,
   $default_ssl_crl_check                                                = undef,
+  $default_ssl_reload_on_change                                         = false,
```
bastelfreak, line 478: could you please add a datatype for the new parameter?
```suggestion
  Boolean $default_ssl_reload_on_change                                 = false,
```

timdeluxe: @bastelfreak Done.

david22swan:
@timdeluxe Apologies, for that to disappear @ekohl must approve the PR.

manifests/vhost.pp
```diff
@@ -2707,7 +2711,64 @@
...
   # - $apache_version
-  if $ssl {
+  if $ssl and $ensure == 'present' {
+    if $ssl_reload_on_change {
+      if $ssl_cert {
+        include apache::mod::ssl::reload
+        $_ssl_cert_copy = regsubst($ssl_cert, '/', '_', 'G')
+        file { "${filename}${_ssl_cert_copy}":
+          path    => "${apache::params::puppet_ssl_dir}/${filename}${_ssl_cert_copy}",
+          source  => "file://${ssl_cert}",
```
ekohl, line 2721: Does this automatically create a dependency on that file or should it have something like:
```puppet
File <| title == $ssl_cert |> -> File["${filename}${_ssl_cert_copy}"]
```

manifests/vhost.pp
```diff
@@ -2707,7 +2711,64 @@
   # - $ssl_openssl_conf_cmd
   # - $ssl_stapling
   # - $apache_version
-  if $ssl {
+  if $ssl and $ensure == 'present' {
+    if $ssl_reload_on_change {
+      if $ssl_cert {
+        include apache::mod::ssl::reload
+        $_ssl_cert_copy = regsubst($ssl_cert, '/', '_', 'G')
+        file { "${filename}${_ssl_cert_copy}":
```
ekohl, line 2719: There's a lot of duplication for this. Can't this be captured in a define somehow and reused? Perhaps a loop?

```puppet
[$ssl_cert, $ssl_key, $ssl_ca, $ssl_chain, $ssl_crl].each |$ssl_file| {
  if $ssl_file {
    include apache::mod::ssl::reload
    $_ssl_file_copy = regsubst($ssl_file, '/', '_', 'G')
    file { "${filename}${_ssl_file_copy}":
      path    => "${apache::params::puppet_ssl_dir}/${filename}${_ssl_file_copy}",
      source  => "file://${ssl_file}",
      mode    => '0640',
      seltype => 'cert_t',
      notify  => Class['apache::service'],
    }
  }
} 
```

manifests/vhost.pp
```diff
@@ -2707,7 +2711,64 @@
   # - $ssl_openssl_conf_cmd
   # - $ssl_stapling
   # - $apache_version
-  if $ssl {
+  if $ssl and $ensure == 'present' {
+    if $ssl_reload_on_change {
```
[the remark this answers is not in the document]
timdeluxe: No it can't - see line 2788, where the concat::fragment is located, which must be handled even if the reload feature is not used. Maybe it is clearer when i shift the concat::fragment higher and the reload stuff after that?

manifests/vhost.pp
timdeluxe: We should avoid that and thats the whole point of the way i implemented it: We don't know how the cert file is managed. It may be a file resource in puppet, but does not have to be it. It could be managed completely outside of puppet or via another resource type.
If we would be sure, that the cert file is always a file resource, we could just do a notify dependency from the File to the apache::service and that would be all, but we shouldn't do that.
And yes, that "file://" locator creates a dependency: Once the content of the source file changes, also the "cert copy" file change, resulting into a notify into the apache::service class. This way it works independently of how the real cert files are managed.

manifests/vhost.pp
timdeluxe: Thats actually a good idea, will implement that.

manifests/vhost.pp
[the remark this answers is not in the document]
ekohl: I do think that would be easier to follow. Long deeply nested structures aren't that clear. Perhaps it's also not needed if https://github.com/puppetlabs/puppetlabs-apache/pull/2157#discussion_r663036253 works and can be implemented.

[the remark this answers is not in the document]
timdeluxe: Yes, you are right, but i will do the shift anyway even with the proposal of the other comment, because i find that also better.

manifests/vhost.pp
timdeluxe: Did the same in apache::mod::ssl

manifests/mod/ssl.pp
```diff
@@ -62,6 +62,9 @@
 #   - Debian/Ubuntu + Apache >= 2.4: 'default'.
 #   - Debian/Ubuntu + Apache < 2.4: 'file:${APACHE_RUN_DIR}/ssl_mutex'.
 #
+# @param ssl_reload_on_change
+#   Enable reloading of apache if the content of ssl files have changed.
```
ekohl, line 66: Perhaps this should clarify that it's only the SSL files that are configured here in mod_ssl, not those in vhosts.

manifests/init.pp
```diff
@@ -91,6 +91,9 @@
 #   this parameter with your SSL key's location before deploying this server in a production 
 #   environment.
 #
+# @param default_ssl_reload_on_change
+#   Enable reloading of apache if the content of ssl files have changed.
```
ekohl, line 95: I think this should clarify this **only** sets the default value for vhosts. If you don't have any SSL vhosts or use `mod_ssl` to configure the SSL certificates, it does nothing.

manifests/vhost.pp
```diff
@@ -2707,12 +2711,27 @@
   # - $ssl_openssl_conf_cmd
   # - $ssl_stapling
   # - $apache_version
-  if $ssl {
+  if $ssl and $ensure == 'present' {
```
ekohl, line 2714: Should it actually ensure the old certificate copies are absent if `$ensure` is `absent`?

manifests/mod/ssl.pp
```diff
@@ -174,6 +178,22 @@
...
 
+  if $ssl_reload_on_change {
+    [$ssl_cert, $ssl_key, $ssl_ca].each |$ssl_file| {
+      if $ssl_file {
+        include apache::mod::ssl::reload
+        $_ssl_file_copy = regsubst($ssl_file, '/', '_', 'G')
+        file { $_ssl_file_copy:
+          path    => "${apache::params::puppet_ssl_dir}/${_ssl_file_copy}",
+          source  => "file://${ssl_file}",
+          mode    => '0640',
```
ekohl, line 189: Should the owner and group also be managed? (same question for the vhost)

manifests/vhost.pp
```diff
@@ -2707,12 +2711,27 @@
...
     concat::fragment { "${name}-ssl":
       target  => "${priority_real}${filename}.conf",
       order   => 230,
       content => template('apache/vhost/_ssl.erb'),
     }
+    if $ssl_reload_on_change {
```
ekohl, line 2720: Nit: I like the visual separation
```suggestion

    if $ssl_reload_on_change {
```

ekohl (commented):
Thanks! This is much easier to follow.

manifests/vhost.pp
timdeluxe: Not needed, the puppet_ssl_dir is managed with purge => true, so old copies will be cleaned up by that.

manifests/init.pp
```diff
@@ -91,6 +91,12 @@
 #   this parameter with your SSL key's location before deploying this server in a production 
 #   environment.
 #
+# @param default_ssl_reload_on_change
```
ekohl, line 94: Thanks for the update of the docs. That actually highlighted something I missed: that it's only the default vhost. Should the name make that explicit?
```suggestion
# @param default_vhost_ssl_reload_on_change
```

REFERENCE.md
```diff
@@ -7918,6 +7952,14 @@ Data type: `Any`
 
 Default value: `$apache::params::ssl_certs_dir`
 
+##### <a name="ssl_reload_on_change"></a>`ssl_reload_on_change`
```
ekohl, line 7955: Somehow this type has no docs, but I don't see why. The code has a parameter.

ekohl (commented):
Perhaps some example in https://github.com/puppetlabs/puppetlabs-apache/tree/main/examples would be nice?

manifests/init.pp
timdeluxe: @ekohl Oh, well, right now i see it: All the default_* parameters in init.pp are also used in vhost.pp for non-default vhosts, as long as there is no own parameter defined for that vhost.
If you agree i would remove the complete "note" i've added to the docs, because it is not correct.

REFERENCE.md
timdeluxe: @ekohl I don't get what you mean, i see nothing wrong...?

ekohl: Normally I would expect a description. However, now I don't see any text.

timdeluxe: I still don't get it. Other parameters don't have any text as well in that section. What exactly are you missing?

timdeluxe:
> Perhaps some example in https://github.com/puppetlabs/puppetlabs-apache/tree/main/examples would be nice?

I don't think that this feature is so important, that it needs an example. The ones who need it, will find it anyway...

REFERENCE.md
timdeluxe: @ekohl Can you please give feedback here?

ekohl: I traced it down to https://github.com/puppetlabs/puppetlabs-apache/pull/2165.

timdeluxe: ekohl: Good finding. How should we proceed? Should i merge your commit into my branch here and after that regenerate REFERENCE.md (resulting in 3 commits)? 
Or is it okay, when i do the simple fix by myself and regenerate REFERENCE.md as well and put it into one commit (like i usually did here by force pushes)?

ekohl: It'd be great if you could rebase on my PR. I now regenerated REFERENCE.md in it so your diff should remain small.

ekohl (commented):
Code wise :+1: but I see a few merge commits. Could you rebase?

timdeluxe:
> Code wise 👍 but I see a few merge commits. Could you rebase?

After a fighting a bit with git logic (i always get knots in my brain) it is hopefully correctly done now.

ekohl (approved):
I started CI. I also hope someone else can review but :+1: from me. Thanks for getting this patch in good shape!

timdeluxe:
> I started CI. I also hope someone else can review but 👍 from me. Thanks for getting this patch in good shape!

I hope the same. Also thanks to you for guiding me through it and spotting a lot of little things to improve.

timdeluxe:
Tests on RedHat failed not due to my code but due to mirrors not answering correctly - don't know if it is a temporary or a general problem, which can be fixed :/

ekohl:
That's a known issue (https://github.com/puppetlabs/puppetlabs-apache/pull/2171 tries to look into it)

david22swan (approved):
LGTM

Will confer with another team member first but look's good to merge.

PR #2158, opened 2021-06-16
Update the default version of Apache for Amazon Linux 2

turnopil:
- updated `$default` variable for Amazon Linux 2

The default `httpd` package version on the Amazon Linux 2 is `2.4`
```
Name        : httpd
Arch        : x86_64
Version     : 2.4.46
Release     : 1.amzn2
Size        : 1.3 M
Repo        : amzn2-core/2/x86_64
Summary     : Apache HTTP Server
URL         : https://httpd.apache.org/
License     : ASL 2.0
Description : The Apache HTTP Server is a powerful, efficient, and extensible
            : web server.
```
Based on the above we need to update the `$default` variable to support AL2

### DIFF
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -837,7 +837,11 @@ class apache::mod::passenger (
 
   if $::osfamily == 'RedHat' and $manage_repo {
     if $::operatingsystem == 'Amazon' {
-      $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/6Server/$basearch'
+      if $::operatingsystemmajrelease == '2' {
+        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/7Server/$basearch'
+      } else {
+        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/6Server/$basearch'
+      }
     } else {
       $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/$releasever/$basearch'
     }
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -838,9 +838,9 @@ class apache::mod::passenger (
   if $::osfamily == 'RedHat' and $manage_repo {
     if $::operatingsystem == 'Amazon' {
       if $::operatingsystemmajrelease == '2' {
-        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/7Server/$basearch'
+        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/7/$basearch'
       } else {
-        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/6Server/$basearch'
+        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/6/$basearch'
       }
     } else {
       $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/$releasever/$basearch'
--- a/manifests/version.pp
+++ b/manifests/version.pp
@@ -11,8 +11,8 @@ class apache::version (
       if $scl_httpd_version {
         $default = $scl_httpd_version
       }
-      elsif ($::operatingsystem == 'Amazon') {
-        $default = '2.2'
+      elsif ($::operatingsystem == 'Amazon' and $::operatingsystemmajrelease == '2') {
+        $default = '2.4'
       } elsif ($::operatingsystem == 'Fedora' and versioncmp($facts['operatingsystemmajrelease'], '18') >= 0) or ($::operatingsystem != 'Fedora' and versioncmp($facts['operatingsystemmajrelease'], '7') >= 0) {
         $default = '2.4'
       } else {


manifests/version.pp
```diff
@@ -11,8 +11,8 @@
       if $scl_httpd_version {
         $default = $scl_httpd_version
       }
-      elsif ($::operatingsystem == 'Amazon') {
-        $default = '2.2'
+      elsif ($::operatingsystem == 'Amazon' and $::operatingsystemmajrelease == '2') {
```
ekohl, line 14: Is Amazon still considered as part of the Red Hat OS family? Do you have recent facts you can share somewhere? I've never used Amazon Linux.

turnopil: hi @ekohl 
Yep. Still a part of the RedHat
```
facter operatingsystem operatingsystemmajrelease osfamily
operatingsystem => Amazon
operatingsystemmajrelease => 2
osfamily => RedHat
```

ekohl (approved):

ekohl (changes requested):
I'm not sure why Ubuntu testing fails but that can't be related.

Maybe another place to check is here:

https://github.com/puppetlabs/puppetlabs-apache/blob/958023b094afe43273cd1cc4a1393b87f22653fc/manifests/mod/passenger.pp#L839-L843

AFAIK that is also not correct with Amazon 2 since it's based on EL7 while Amazon 1 was based on EL6.

turnopil:
> I'm not sure why Ubuntu testing fails but that can't be related.
> 
> Maybe another place to check is here:
> 
> https://github.com/puppetlabs/puppetlabs-apache/blob/958023b094afe43273cd1cc4a1393b87f22653fc/manifests/mod/passenger.pp#L839-L843
> 
> AFAIK that is also not correct with Amazon 2 since it's based on EL7 while Amazon 1 was based on EL6.

I guess the root cause of the Ubuntu failed tests are unmet dependencies:
```
docker exec litmusimage_ubuntu_16_04-2222 apt-get install -y openssh-server openssh-client
...
Some packages could not be installed.
...
The following packages have unmet dependencies:\n openssh-client : Depends: libedit2 (>= 2.11-20080614) but it is not installable\n
...
```
Maybe docker image or openssh* stuff has been updated

I'm not familiar with the `passenger` module. But it has the same problem that I'm trying to fix. Maybe it requires a separate PR?

turnopil:
btw. What I have found in docs: https://github.com/puppetlabs/puppetlabs-apache#rhelcentos-7
>RHEL/CentOS 7
The apache::mod::passenger and apache::mod::proxy_html classes are untested because the EL7 repository is missing compatible packages, which also blocks us from testing the apache::vhost defined type's rack_base_uris parameter.

NapoSky:
Any ideas when this PR will be accepted after merging is cleared ?
Appreciated, thanks!

ekohl:
> btw. What I have found in docs: https://github.com/puppetlabs/puppetlabs-apache#rhelcentos-7

That's just testing. However, just from reading the code you can see that it will try to use EL6 repos on Amazon Linux. That should also get a conditional that it's Amazon Linux 1.

> Any ideas when this PR will be accepted after merging is cleared ?

If https://github.com/puppetlabs/puppetlabs-apache/pull/2158#pullrequestreview-685407168 is addressed.

manifests/mod/passenger.pp
```diff
@@ -837,7 +837,11 @@
 
   if $::osfamily == 'RedHat' and $manage_repo {
     if $::operatingsystem == 'Amazon' {
-      $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/6Server/$basearch'
+      if $::operatingsystemmajrelease == '2' {
+        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/7Server/$basearch'
```
ekohl, line 841: This URL doesn't exist. See https://oss-binaries.phusionpassenger.com/yum/passenger/el/
```suggestion
        $baseurl = 'https://oss-binaries.phusionpassenger.com/yum/passenger/el/7/$basearch'
```

turnopil: yeah. sorry for that. `6Server` directory also doesn't exist. fixed

ekohl (approved):
This looks correct from a patch perspective. I think the modules team can give some advice on the correct commit message so it meets their standards. I'm not sure if we need to squash this etc.

david22swan:
Everything looks good from my end so I'll go ahead and merge it :)

@turnopil  Thank you for realizing the need for this fix and putting in the work to get it done, hope to see more from you in the future :)

@ekohl Thank you for reviewing this work and helping to get it across the line, great to see you taking advantage of the trusted contributors program :)

Also regarding what you stated in your last message, in regards to the PR titles the general idea is that the title should clearly reflect what the PR is attempting to accomplish so that you can tell at a glance.
In addition as a general rule if the work has a related ticket/issue then the ticket/issue number should be included within the title so as to clearly link them.
When it comes to the amount of commits that are acceptable before a squash is requested, then somewhere between 1-3 is considered to be generally acceptable so long as each commit is clearly distinct.
Though this amount can go up as the size of the overall PR increases, i.e. there are some pieces of work that need to be split up into multiple parts in order for the work done to be better understood.

Sorry if I went on for a while, wanted to be clear.
Once again thank you to the both of you.

PR #2160, opened 2021-06-22
Improve/silence warning from pull-request #2154

sircubbi:
Hi,

pull-request #2154 did try to silence the warnings about servernames and filenames to cases where a problem actually exists.

However, there are two problems with that:

1) there is a check between "name" and "normalized_servername", when it actually should check for differences between "servername" and "normalized_servername".

2) even if that check is silenced, there is still the second warning about "use_port_for_filenames" not enabled. That warning should only be issued if "use_servername_for_filenames" is enabled.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2122,7 +2122,7 @@ define apache::vhost (
     module, the $use_servername_for_filenames will be removed and log/config file names will be derived from the
     sanitized $servername parameter when not explicitly defined.'
     warning($use_servername_for_filenames_warn_msg)
-  } elsif ! $use_port_for_filenames {
+  } elsif $use_servername_for_filenames and ! $use_port_for_filenames {
     $use_port_for_filenames_warn_msg = '
     It is possible for multiple virtual hosts to be configured using the same $servername but a different port. When
     using $use_servername_for_filenames, this can lead to duplicate resource declarations.
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2111,7 +2111,7 @@ define apache::vhost (
     false => $name,
   }
 
-  if ! $use_servername_for_filenames and $name != $normalized_servername {
+  if ! $use_servername_for_filenames and $servername != $normalized_servername {
     $use_servername_for_filenames_warn_msg = '
     It is possible for the $name parameter to be defined with spaces in it. Although supported on POSIX systems, this
     can lead to cumbersome file names. The $servername attribute has stricter conditions from Apache (i.e. no spaces)


manifests/vhost.pp
```diff
@@ -2111,7 +2111,7 @@
     false => $name,
   }
 
-  if ! $use_servername_for_filenames and $name != $normalized_servername {
+  if ! $use_servername_for_filenames and $servername != $normalized_servername {
```
ekohl, line 2114: This looks correct to me. I guess in my testing I didnt' specify a servername so they happened to match.

manifests/vhost.pp
```diff
@@ -2122,7 +2122,7 @@
     module, the $use_servername_for_filenames will be removed and log/config file names will be derived from the
     sanitized $servername parameter when not explicitly defined.'
     warning($use_servername_for_filenames_warn_msg)
-  } elsif ! $use_port_for_filenames {
+  } elsif $use_servername_for_filenames and ! $use_port_for_filenames {
```
ekohl, line 2125: This doesn't look correct to me. I think the intention is to warn users ahead of time that they'll run into problems in the future.

ekohl (commented):
I must admit in reviewing this that I'm not entirely sure the whole deprecation for version 8 is that good of an idea. For example, I'm not looking forward to it. It's really a mess when upgrading.

For us the impact of removing `use_port_for_filenames` would be a lot less when it would not use the port if it's default (80 for http, 443 for https).

manifests/vhost.pp
sircubbi: I understand. However, I am not sure about the correct usage then. Most vhosts-definitions I have here are actually two definitions for the same servername (the non-ssl version, and the ssl-version).
As for know every puppet run fills the log with that warnings, and the only way to get rid of them is by enabling the $use_port_for_filenames parameter. However I am not really sure if it is a good idea to enable that parameter, since it is scheduled to be deprecated with the next releases.

What would be a proper way to declare two vhosts for the same servername without getting the warning?

manifests/vhost.pp
ekohl: Btw, I'm reading this again and what I based it on was this part:
https://github.com/puppetlabs/puppetlabs-apache/blob/337e8b00b2a54f503e2ab0fa5c16d7a3a648c302/manifests/vhost.pp#L2106-L2112

So right now it uses `$name` for `$filename`, but after it will use `$normalized_servername` which is a change. Maybe I should have done it otherwise and determined `$v8_filename` instead and compared that.

sircubbi: Sorry, as I see now the check between $name/$normalized_servername and between $servername/$normalized_servername are both bogus in that case.

While I see no problem at all to use a normalized-servername as the filename for the apache-configsnipped, I understand the reasoning in rolling that change in stages and issue warnings.

However, I really would like to have a possibility to "defuse" these warnings on a global level. While still messy, maybe we could lookup the $use_servername_for_filenames and $use_port_for_filenames from hiera? In that case it would be at least possible to enable both parameters on one place.

manifests/vhost.pp
sircubbi: As already mentioned on #2161 that would not help if you need the same servername for a ssl and a non-ssl version. At least one filename has to be prefixed or suffixed with the protocol. But then using the portname seems to be much more appropriate.

elofu17:
I'm annoyed by the logged message, logged over and over, from every puppet node with apache, for each puppet run.
All nodes are configured as standard as possible, so I don't think I need to see the warnings all the time.

```
  class { 'apache':
  }
  apache::vhost { $::fqdn:
    port    => 80,
    docroot => "/var/www/${::fqdn}",
  }
```
This result in e.g.:
```
2021-09-23 18:41:07,692 WARN  [puppetserver] Scope(Apache::Vhost[foobar.infra.gazonk.com])
    It is possible for multiple virtual hosts to be configured using the same $servername but a different port. When
    using $use_servername_for_filenames, this can lead to duplicate resource declarations.
    When $use_port_for_filenames = true, the $servername and $port parameters, sanitized, are used to construct log and
    config file names.

    From version v7.0.0 of the puppetlabs-apache module, this parameter will default to true. From version v8.0.0 of the
    module, the $use_port_for_filenames will be removed and log/config file names will be derived from the
    sanitized $servername parameter when not explicitly defined.
```


With only one port, there's no risk for duplicate declarations.
With my standard code, I should see no warnings/instructions at all -- they just confuse me.

The current defaults in v6 should be enough.
The defaults in v7 and v8 should be also be enough.
That is,. I should not need to do any changes to my code during this transition.
Therefore I should not be spammed with annoying warnings/instructions.

PR #2161, opened 2021-06-22
Change the v8 vhost filename logging logic

ekohl:
Previously the plan was to change $use_port_for_filenames to always true in v8. However, that creates a major upgrade pain. It is better to only do so if it's non-default.

This also changes the upgrade warning logic to match.

I'd also like to emphasize that this is a really painful plan. The current logging really works a lot better and I'd still like to propose to revert this whole plan for v8.

Right now this is untested so perhaps more of an RFC.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2076,13 +2076,16 @@ define apache::vhost (
     $priority_real = '25-'
   }
 
-  # https://httpd.apache.org/docs/2.4/fr/mod/core.html#servername
+  # https://httpd.apache.org/docs/2.4/en/mod/core.html#servername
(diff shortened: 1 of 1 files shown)


sircubbi:
I have to admit that I more or less adopted the naming-schema from the official example, although using a hyphen instead of a space, eg "vhost { 'my.host.com-nonssl': ... }" and "vhosts { 'my.host.com-ssl': }".

Maybe a better way would be the just check if "$name" contains some problematic characters and issue a warning if that is the case only?

I am a bit torn here, as resource-naming is always a bit tricky if used as a filename.

Also there might be an edge case, where the same servername+port combo is used several times, but on different binding-ips.

PR #2162, opened 2021-06-23
Adds mod_authz_host

bshelton:
I'd like to add the authz_host module since this is helpful for restricting traffic to apache within the configuration.


manifests/mod/authz_host.pp
```diff
@@ -0,0 +1,11 @@
+# @summary
+#   Installs `mod_authz_host`.
+# 
+# @see https://httpd.apache.org/docs/current/mod/mod_authz_host.html for additional documentation.
+#
+class apache::mod::authz_host {
+  if versioncmp($apache_version, '2.4') >= 0 {
+    include apache
+    ::apache::mod { 'authz_host': }
```
ekohl, line 9: This isn't needed since Puppet < 4 support was dropped:
```suggestion
    apache::mod { 'authz_host': }
```

manifests/mod/authz_host.pp
```diff
@@ -0,0 +1,11 @@
+# @summary
+#   Installs `mod_authz_host`.
+# 
+# @see https://httpd.apache.org/docs/current/mod/mod_authz_host.html for additional documentation.
+#
+class apache::mod::authz_host {
+  if versioncmp($apache_version, '2.4') >= 0 {
```
ekohl, line 7: Should it have an else clause with `fail()`

ekohl (changes requested):
Could you add a basic spec test so it at least is tested a bit?

sheenaajay:
@bshelton Thanks alot for your contribution. 
@ekohl Thanks alot for the valuable comments.

manifests/mod/authz_host.pp
bshelton: Like this?

manifests/mod/authz_host.pp
bshelton: resolved.

chelnak:
Lets rebase this PR to re-kick CI. If everything goes well then we can merge 😄 

Hey @ekohl! Is there any chance that you could take a look at this PR and let us know if the changes made are what you were after?

ekohl:
@chelnak I did ask for a unit test and I don't see one now. It would be fairly easy to add, so perhaps someone else can jump in and write that.

chelnak:
Thanks for the speedy reply. We will take a look in to this!

PR #2163, opened 2021-07-01
Fetch correct _module_id value in apache::mod::php

storm49152:
[CentOS, can't speak for other distros] The _module_id works fine with PHP from standard repo's. However, the PHP version in those repositories is usually old, and sometimes you need a newer PHP version, in which case one often uses the Remi repository. If you want to use e.g. php74, it will be installed without issues, but Apache will not start:

```puppet
class { 'apache::mod::php':
  package_name => 'php74',
  php_version => '74',
}
```

```
Jul 01 10:29:07 server-1 httpd[15883]: httpd: Syntax error on line 40 of /etc/httpd/conf/httpd.conf: Syntax error on line 1 of /etc/httpd/conf.modules.d/php.load: Can't locate API module structure `php_module' in file /etc/httpd/modules/libphp74.so: /etc/httpd/modules/libphp74.so: undefined symbol: php_module
```

```shell
# cat /etc/httpd/conf.modules.d/php.load
LoadModule php_module modules/libphp74.so
```

I traced this to:

```puppet
$_module_id = $_php_major ? {
  '5'     => 'php5_module',
  '7'     => 'php7_module',
  default => 'php_module',
}
```

With the example above, `$_php_major` isn't matching `7` but `74`. Therefore `$_module_id` defaults to `php_module` instead of `php7_module` and Apache will not start.
When a regex is used to match `/^5/` and `/^7/`, `$_module_id` will be set to `php7_module`, configuring the `php.load` file properly and Apache will start:

```shell
# cat /etc/httpd/conf.modules.d/php.load
LoadModule php7_module modules/libphp74.so
```

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -70,8 +70,8 @@ class apache::mod::php (
     $_lib = "${libphp_prefix}${php_version}.so"
   }
   $_module_id = $_php_major ? {
-    '5'     => 'php5_module',
-    '7'     => 'php7_module',
+    /^5/     => 'php5_module',
+    /^7/     => 'php7_module',
     default => 'php_module',
   }
 


ekohl (changes requested):
I think the problem is that you pass the version as `74` but you should really pass `7.4`. As you can see, it compares the *major* version. That's extracted using a regex on line 64.

storm49152:
You are correct. However, I did did try `7.4` and that didn't work either. I locally reverted the change I proposed and tested this again.

```puppet
class { 'apache::mod::php':
  package_name => 'php74-php',
  php_version  => '7.4',
}
```

```
# puppet agent -t
[...snip...]
Error: /Stage[main]/Apache::Service/Service[httpd]: Failed to call refresh: Systemd restart for httpd failed!
journalctl log for httpd:
-- Logs begin at Thu 2021-07-01 05:05:51 CEST, end at Mon 2021-07-05 22:52:45 CEST. --
Jul 05 22:52:44 server-1 systemd[1]: Stopping The Apache HTTP Server...
Jul 05 22:52:45 server-1 systemd[1]: httpd.service: Succeeded.
Jul 05 22:52:45 server-1 systemd[1]: Stopped The Apache HTTP Server.
Jul 05 22:52:45 server-1 systemd[1]: Starting The Apache HTTP Server...
Jul 05 22:52:45 server-1 httpd[212559]: httpd: Syntax error on line 40 of /etc/httpd/conf/httpd.conf: Syntax error on line 1 of /etc/httpd/conf.modules.d/php7.4.load: Cannot load modules/libphp7.4.so into server: /etc/httpd/modules/libphp7.4.so: cannot open shared object file: No such file or directory
Jul 05 22:52:45 server-1 systemd[1]: httpd.service: Main process exited, code=exited, status=1/FAILURE
Jul 05 22:52:45 server-1 systemd[1]: httpd.service: Failed with result 'exit-code'.
Jul 05 22:52:45 server-1 systemd[1]: Failed to start The Apache HTTP Server.
[...snip...]

# cat /etc/httpd/conf.modules.d/php7.4.load 
LoadModule php7_module modules/libphp7.4.so

# ls /etc/httpd/modules/libphp7.4.so
ls: cannot access '/etc/httpd/modules/libphp7.4.so': No such file or directory

# ls /etc/httpd/modules/libphp*
lrwxrwxrwx 1 root root 55 Jun 29 18:24 /etc/httpd/modules/libphp74.so -> /opt/remi/php74/root/usr/lib64/httpd/modules/libphp7.so
```

So, even though my proposal  seemed to work, looking back it looks like it's not for the right reason nor the correct fix for the actual problem. However, AFAICS there still needs to be some fix to correct the content of `php7.4.load` since the resulting config is not working.
Or could there something else I'm doing wrong in using the apache module..?

ekohl:
I think you're right. Here's the relevant code:
https://github.com/puppetlabs/puppetlabs-apache/blob/337e8b00b2a54f503e2ab0fa5c16d7a3a648c302/manifests/mod/php.pp#L64-L71
It looks like it should be using `$_php_version_no_dot` instead of `$php_version` there. However, I don't know if that breaks other distros.

storm49152:
So the ticket is now marked as stale.

I think I traced a problem that should be fixed. However, even though it felt trivial to do, apparently I don't know how to fix this properly. If anyone can take this from me, then that would be great. I fixed the problem locally as described. Maybe it's not the best/preferred way, but at least now we can use the module. Obviously, this local fix will be undone whenever the module is updated, so it's suboptimal.

PR #2165, opened 2021-07-13
(maint) Fix puppet-strings docs on apache::vhost

ekohl:
With this empty line between the commend and the define, puppet-strings doesn't consider the comment as a documentation string. That means all the documentation that is written doesn't show up in REFERENCE.md.

I can't regenerate REFERENCE.md due to a missing puppet-module-posix-default-r3.0 gem so somebody else has to do that.

It was found in https://github.com/puppetlabs/puppetlabs-apache/pull/2157/files#r663838497 and introduced in 7e233fe7cc108605384506f11083aeea6e3a2464.


ekohl:
For what it's worth: https://github.com/voxpupuli/puppet-lint-param-docs would have caught this.

ekohl:
I now regenerated REFERENCE.md.

ekohl:
@sheenaajay could you take a look?

bastelfreak (approved):

PR #2166, opened 2021-07-15
(MODULES-10899) Load php module with the right libphp file

sheenaajay:
Request for a review
The configuration file is updated in the required format for php_version 8 (/etc/httpd/conf.modules.d/php.load)
` LoadModule php_module modules/libphp.so`
Unit tests are updated to reflect the change.

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -67,7 +67,10 @@ class apache::mod::php (
     $_lib = "librh-php${_php_version_no_dot}-php${_php_major}.so"
   } else {
     # Controls php version and libphp prefix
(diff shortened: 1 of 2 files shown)


carabasdaniel (approved):
👍  LGTM

mxey:
This change breaks on Debian. The filename for libphp remains versioned, even for PHP 8. For example https://packages.debian.org/bookworm/amd64/libapache2-mod-php8.1/filelist contains /usr/lib/apache2/modules/libphp8.1.so, not /usr/lib/apache2/modules/libphp.so

ekohl:
Please file a new issue and/or PR. Comments on old PRs generally get lost and don't lead to a resolution.

PR #2167, opened 2021-07-16
Add an apache::vhost::proxy define

ekohl:
This builds on apache::vhost::fragment with a more specialized type for a reverse proxy setup. This reuses the same template as apache::vhost but has data types. The types are stricter, but this leads to a clearer API. For example, it doesn't accept String and Array[String] but in that case always wants an Array.

The old parameters on vhost remain and can continue to be used.

Currently a draft since this isn't really ready. I haven't even ran the tests yet but I wanted to share the concept. It was inspired by https://github.com/puppetlabs/puppetlabs-apache/pull/1659.

### DIFF
--- a/manifests/vhost/proxy.pp
+++ b/manifests/vhost/proxy.pp
@@ -1,0 +1,79 @@
+# @summary Configure a reverse reverse proxy for a vhost
+#
+# @param proxy_dest
+#   Specifies the destination address of a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) configuration.
+#
+# @param proxy_pass
+#   Specifies an array of `path => URI` values for a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) 
+#   configuration. Optionally, parameters can be added as an array.
+#   ``` puppet
+#   apache::vhost { 'site.name.fdqn':
+#     …
+#     proxy_pass => [
+#       { 'path' => '/a', 'url' => 'http://backend-a/' },
+#       { 'path' => '/b', 'url' => 'http://backend-b/' },
+#       { 'path' => '/c', 'url' => 'http://backend-a/c', 'params' => {'max'=>20, 'ttl'=>120, 'retry'=>300}},
+#       { 'path' => '/l', 'url' => 'http://backend-xy',
+#         'reverse_urls' => ['http://backend-x', 'http://backend-y'] },
+#       { 'path' => '/d', 'url' => 'http://backend-a/d',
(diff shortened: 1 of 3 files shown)


manifests/vhost/proxy.pp
```diff
@@ -0,0 +1,79 @@
+# @summary Configure a reverse reverse proxy for a vhost
+#
+# @param proxy_dest
+#   Specifies the destination address of a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) configuration.
+#
+# @param proxy_pass
+#   Specifies an array of `path => URI` values for a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) 
+#   configuration. Optionally, parameters can be added as an array.
+#   ``` puppet
```
ekohl, line 9: Note to self: this should be converted in an `@example`

manifests/vhost/proxy.pp
```diff
@@ -0,0 +1,79 @@
+# @summary Configure a reverse reverse proxy for a vhost
+#
+# @param proxy_dest
+#   Specifies the destination address of a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) configuration.
+#
+# @param proxy_pass
+#   Specifies an array of `path => URI` values for a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) 
+#   configuration. Optionally, parameters can be added as an array.
```
ekohl, line 8:
```suggestion
#   configuration.
```

manifests/vhost/proxy.pp
```diff
@@ -0,0 +1,79 @@
...
+#       { 'path' => '/f', 'url' => 'http://backend-f/',
+#         'setenv' => ['proxy-nokeepalive 1','force-proxy-request-1.0 1']},
+#       { 'path' => '/g', 'url' => 'http://backend-g/',
+#         'reverse_cookies' => [{'path' => '/g', 'url' => 'http://backend-g/',}, {'domain' => 'http://backend-g', 'url' => 'http:://backend-g',},], },
+#       { 'path' => '/h', 'url' => 'http://backend-h/h',
+#         'no_proxy_uris' => ['/h/admin', '/h/server-status'] },
+#     ],
+#   }
+#   ```
+#   * `reverse_urls`. *Optional.* This setting is useful when used with `mod_proxy_balancer`. Values: an array or string.
```
ekohl, line 31: These should probably be moved into the specific type.

ekohl:
Replaced by https://github.com/puppetlabs/puppetlabs-apache/pull/2169. Thanks @wbclark!

PR #2169, opened 2021-07-20
Add an apache::vhost::proxy define

wbclark:
This builds on apache::vhost::fragment with a more specialized type for
a reverse proxy setup. This reuses the same template as apache::vhost
but has data types. The types are stricter, but this leads to a clearer
API. For example, it doesn't accept String and Array[String] but in that
case always wants an Array.

The old parameters on vhost remain and can continue to be used.

### DIFF
--- a/manifests/vhost/proxy.pp
+++ b/manifests/vhost/proxy.pp
@@ -106,7 +106,7 @@
 #
 define apache::vhost::proxy (
   String[1] $vhost,
-  $priority = undef,
+  Optional[Variant[Integer,String,Boolean]] $priority = undef,
   Integer[0] $order = 170,
   Optional[Stdlib::Port] $port = undef,
   Optional[String[1]] $proxy_dest = undef,
--- a/manifests/vhost/proxy.pp
+++ b/manifests/vhost/proxy.pp
@@ -1,0 +1,81 @@
+# @summary Configure a reverse reverse proxy for a vhost
+#
+# @param proxy_dest
+#   Specifies the destination address of a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass) configuration.
+#
+# @param proxy_pass
+#   Specifies an array of `path => URI` values for a [ProxyPass](https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass)
+#   configuration. Optionally, parameters can be added as an array.
+#   ``` puppet
+#   apache::vhost { 'site.name.fdqn':
+#     …
+#     proxy_pass => [
+#       { 'path' => '/a', 'url' => 'http://backend-a/' },
+#       { 'path' => '/b', 'url' => 'http://backend-b/' },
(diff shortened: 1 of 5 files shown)


wbclark:
Discussed with @ekohl and agreed to take over https://github.com/puppetlabs/puppetlabs-apache/pull/2167

This PR therefore contains my latest work on it to get the tests into a working state.

manifests/vhost/proxy.pp
```diff
@@ -0,0 +1,81 @@
...
+#         'keywords' => ['nocanon', 'interpolate'] },
+#       { 'path' => '/f', 'url' => 'http://backend-f/',
+#         'setenv' => ['proxy-nokeepalive 1','force-proxy-request-1.0 1']},
+#       { 'path' => '/g', 'url' => 'http://backend-g/',
+#         'reverse_cookies' => [{'path' => '/g', 'url' => 'http://backend-g/',}, {'domain' => 'http://backend-g', 'url' => 'http:://backend-g',},], },
+#       { 'path' => '/h', 'url' => 'http://backend-h/h',
+#         'no_proxy_uris' => ['/h/admin', '/h/server-status'] },
+#     ],
+#   }
+#   ```
```
ekohl, line 30: This should be updated. I initially copied it from `apache::vhost`. I'd now also make it a separate `@example` block.

types/vhost/proxypass.pp
```diff
@@ -0,0 +1,16 @@
+# @summary A proxy pass definition for in an Apache vhost
```
ekohl, line 1: In the define itself there's more documentation. That probably makes sense in this file as well (or only here?)

david22swan:
@wbclark Apologies for the wait on a review.
Anyway, after taking a look over this pr it for the most part looks good to me, however would be more comfortable if you responded to ekohls comments before I merged it in.

wbclark:
Thanks very much @ekohl and @david22swan for the feedback, and sorry for the delay in resuming work on this.

I've updated this with major improvements to the documentation for the data type and the defined resource type, in order to address the provided feedback.

wbclark:
@david22swan thanks, I've made the requested updates.

I did these in separate commits for easier reviewing, I assume you'll want to squash before merging. Let me know if you prefer me to do that and force push, or if you'd rather do the squashing on your end during merge. Thanks!

david22swan:
@wbclark One more quick change sorry, but to the best of my knowledge, $priority should be `Variant[Integer,String,Boolean]`
Integer and String for passing in input and Boolean for if you wish to disable it and set it to false

david22swan:
@wbclark Thanks for putting in the work and getting this so clean. It LGTM now so I'm gonna go ahead and merge.

Also, to clarify about the accepted values, while the apache config only accepts integer values being given, the original examples and default variable both indicate it being passed through as a string, thus necessitating that we accept input in that manner.

Anyway, thanks again for putting in the work and getting it done so fast :)
Looking forward to seeing more

david22swan (approved):
LGTM

wbclark:
Thanks @david22swan and @ekohl ! It's been a pleasure working with you on it.

PR #2063, opened 2020-09-04
Configure default shared lib path for mod_wsgi on RHEL8

nbarrientos:
RHEL8:
```
  # rpm -ql python3-mod_wsgi | grep \.so$
  /usr/lib64/httpd/modules/mod_wsgi_python3.so
```

RHEL7:
```
  # rpm -ql mod_wsgi | grep \.so$
  /usr/lib64/httpd/modules/mod_wsgi.so
```

Tests might need to be changed. The package name selection has already been patched in commit [d346b76d1aab954b642c55942d3ff8cf6cb578ac](https://github.com/puppetlabs/puppetlabs-apache/commit/d346b76d1aab954b642c55942d3ff8cf6cb578ac).

PR #2174, opened 2021-07-22
(MODULES-11075) Improve future version handling for RHEL

mwhahaha:
This change updates the version handling to better future proof the
logic when newer versions are shipped. This should prevent having to add
things like #2063, #2021, #2038, etc.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -175,9 +175,8 @@ class apache::params inherits ::apache::version {
       $mod_dir            = "${httpd_dir}/conf.modules.d"
     } else {
       $mod_dir              = $facts['operatingsystemmajrelease'] ? {
-        '7'     => "${httpd_dir}/conf.modules.d",
-        '8'     => "${httpd_dir}/conf.modules.d",
-        default => "${httpd_dir}/conf.d",
+        '6'     => "${httpd_dir}/conf.d",
+        default => "${httpd_dir}/conf.modules.d",
       }
     }
     $mod_enable_dir       = undef
@@ -214,9 +213,8 @@ class apache::params inherits ::apache::version {
       'auth_mellon'           => 'mod_auth_mellon',
       'auth_openidc'          => 'mod_auth_openidc',
       'authnz_ldap'           => $facts['operatingsystemmajrelease'] ? {
-        '7'     => 'mod_ldap',
-        '8'     => 'mod_ldap',
-        default => 'mod_authz_ldap',
+        '6'     => 'mod_authz_ldap',
+        default => 'mod_ldap',
       },
       'authnz_pam'            => 'mod_authnz_pam',
       'fastcgi'               => $facts['operatingsystemmajrelease'] ? {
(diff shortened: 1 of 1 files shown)


spec/spec_helper_local.rb
```diff
@@ -57,6 +57,10 @@
   let(:facts) { on_supported_os['redhat-8-x86_64'] }
 end
 
+shared_context 'RedHat 9' do
+  let(:facts) { on_supported_os['redhat-9-x86_64'] }
+end
```
ekohl, line 62: I don't think there are facts for this yet. This feels a bit premature. Also, it doesn't look like any test uses this so I'd prefer to leave it out until it's needed.

ekohl (commented):
All changes to params.pp look correct. Perhaps we can add CentOS Stream 9 facts to https://github.com/voxpupuli/facterdb and get some early testing set up.

spec/spec_helper_local.rb
mwhahaha: ok

david22swan:
@mwhahaha Apologies, but this isn't something that can be merged right now. Support/compatibility for RHEL and CentOS 9 in our modules can only be added after they are supported in the agent.

Unfortunately this means I will have to close this PR, but I would first like to say thank you for putting in the thought and the work that you have and I hope to see more from you in the future.

ekohl:
I think disagree with not merging it altogether. IMHO we can change params.pp to be more future compatible. We wouldn't claim it in metadata.json, but there's no harm in the params.pp changes.

manifests/params.pp
```diff
@@ -175,9 +175,8 @@
       $mod_dir            = "${httpd_dir}/conf.modules.d"
     } else {
       $mod_dir              = $facts['operatingsystemmajrelease'] ? {
-        '7'     => "${httpd_dir}/conf.modules.d",
-        '8'     => "${httpd_dir}/conf.modules.d",
-        default => "${httpd_dir}/conf.d",
+        '6'     => "${httpd_dir}/conf.d",
```
ekohl, line 178: Where are we today with EL5 support? Is it really dead or should we keep some code for it?

mwhahaha:
We need this landed so we can get EL9 started using the community version. I would appreciate not holding on to this because it blocks our EL9 work.

manifests/params.pp
mwhahaha: It's no longer in the metadata.json and 5 has been EOL since November 30, 2020 per https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates.  I can add it back in but I think 6+ is the current support list?

mwhahaha:
And for the record I'm core on the openstack-* modules which is where we want to consume these changes as we work on getting openstack spun up on CentOS9

manifests/params.pp
ekohl: I agree and I'd love to see it die, but the support policy of Puppetlabs' modules is not 100% matching metadata.json (which always confuses me). For example, there's still conditionals left in the codebase. That always makes me cautious with removing it.

mwhahaha:
ok i can go back and remove this optimization

ekohl:
I'd like someone from the modules team to weigh in, but sadly I can't ping them.

david22swan:
@mwhahaha @ekohl 
After talking it over further with some other members of the team, the conclusion we came to is that while the params optimizations look to be on the right track, as previously stated, we cannot commit to any RHEL 9 support so specific references to it within the code, such as the commit title and the metadata and $php_version changes, cannot be merged in.

If you wish to remove these parts of the pr however and push the changes up, either here or in a fresh pr, they look good to merge in my eyes.

Once again apologies that we are not able to merge your full changes at this time.

mwhahaha:
K i'll update this one and just frame it as a future proofing/optimization

mwhahaha:
@david22swan Will this suffice?

ekohl (approved):

david22swan (approved):
LGTM
Thank you for getting back so fast on the changes, and once again I am sorry we couldn't get in all that you wanted.

Hope to see more from you in the future :)

PR #2181, opened 2021-08-23
Use correct PHP version for Debian Bullseye

kienanstewart:
Debian Bullseye uses PHP 7.4 by default. Without adding it in params.pp, PHP 7.2 is used and will fail to install.


ekohl:
There is already support for Bullseye in the latest git commit. See #2180. However, it's unreleased right now so please use a git version until there is a release. If you do have issues with that commit included, please let us know.

PR #2192, opened 2021-09-12
Add support for setting UserDir in Virual Hosts

smortex:
We can only set this setting at the VirtualHost level by relying on the
`custom_fragment` parameter, which does not scale when we want to enable
UserDir for a single VHost because we have to manually disable it on all
other Vhosts.

With this change, one can ensure UserDir is active only on specific
VHost with something like:

```
Apache::Vhost {
  userdir => 'disabled',
}

apache::vhost { 'example.com':
  # [...]
  userdir => 'enabled',
}
```

A specific `userdir` parameter avoid clases that would prevent this
technique to be effective with `custom_fragment`.


smortex:
Hum, just realized that `apache::mod::userdir` has a `custom_fragment` parameter, allowing:

```puppet
class { 'apache::mod::userdir':
  custom_fragment => @(CONFIG),
      UserDir disabled
    | CONFIG
}

apache::vhost { 'example.com':
  # [...]
custom_fragment => @(CONFIG),
      UserDir enabled
    | CONFIG
}
```

I think this change allow a more readable configuration (as described above) so maybe this change still make sense?

sheenaajay:
Thanks @smortex for submitting the PR. Ohh you are right. We can use `apache::mod::userdir` for specifying `UserDir`. 
Just wondering that will lead two ways to use the same parameter. When it comes to maintenance we need to modify it in multiple places.

david22swan:
@smortex Hey sorry for the wait on a review on this.
Anyway was wondering if it was still something that you wished to have merged, considering the discussion between you and Sheena above?
If it is then could you please update the documentation to reflect the change and we can see about getting it merged in.

smortex:
@david22swan Yeah, I see this PR from time to time and have mixed feelings about it: relying on `custom_fragments` indeed works but is quite hackish and I do not like it much (I trend to consider `custom_fragments` as a good way to use new features in apache that have not been integrated in the module while waiting for them to be added).  This PR allows a more readable implementation in end-users control repo but at the cost of quite a lot of code in the module in regard of the added value :shrug:.

If you think it is a waste of time, feel free to close this PR, otherwise I will be a happy user of this feature one it is merged :wink:

david22swan:
@smortex Talked it over and we're good with this change so gonna go ahead and merge.
Thank's for putting in the work :)

PR #2123, opened 2021-02-15
Drop support for Debian < 9

ekohl:
This is a draft PR to push my local code. It includes https://github.com/puppetlabs/puppetlabs-apache/pull/2110. The idea is to keep the codebase smaller. Maybe this is too soon, but I'd like some feedback.

Alternatives are to drop Debian < 8 and Ubuntu < 16.04.

PR #2194, opened 2021-09-15
pdksync - (IAC-1598) - Remove Support for Debian 8

david22swan:
(IAC-1598) - Remove Support for Debian 8
pdk version: `2.1.0`

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -51,7 +51,6 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
-        "8",
         "9",
         "10",
         "11"


ekohl:
Personally I always think that you should also remove the code for it rather than just the metadata. Thoughts on merging https://github.com/puppetlabs/puppetlabs-apache/pull/2123 instead?

david22swan:
@ekohl Would be happy to, is there any more work need done on it or is it good to go?

ekohl:
Previously it was not acceptable due to support policy, but code wise I think it's ready. I just rebased it and took it out of draft. Let's move the review there.

PR #2207, opened 2021-11-29
Add multiple directories to userdir.conf

hahnx117:
This allows the option for multiple directories to be used for UserDirs.


david22swan:
@hahnx117 Sorry for the wait on review
Anyway, could you update the tests to reflect this new functionality?
Something to help prove that it works and catch any future failures.

PR #2209, opened 2021-12-20
Show `allow_override` key in docs.

Tamerz:
For the main virtual host configuration, there is a parameter `override` for this: https://forge.puppet.com/modules/puppetlabs/apache/reference#override

However in the directories parameter, it needs to be set to `allow_override`.

It is confusing to have two different names for the same thing, but changing it at this point would probably break people's settings. At least we can document its usage in the example.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -10157,6 +10157,11 @@ apache::vhost { 'files.example.net':
       'provider' => 'files',
       'deny'     => 'from all',
     },
+    { 'path'           => '/var/www/html',
+      'provider'       => 'directory',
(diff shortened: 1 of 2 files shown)


Tamerz:
I tried to update REFERENCE.md following the instructions but I get an error:

```
$ puppet strings generate --format markdown --out REFERENCE.md
Error: Is a directory @ io_fread - readmes
Error: Try 'puppet help strings generate' for usage
```

ekohl (approved):
Thanks! For reference, I always use `REFERENCE.md` using `rake strings:generate:reference`.

PR #2210, opened 2021-12-21
Allow adding 'ProxyPreserveHost' to vhost location

Tamerz:
I needed to be able to add the `ProxyPreserveHost` setting on a per-location basis in my virtualhost. This PR does just that.

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -493,6 +493,9 @@
     ProxyPassReverse <%= reverse_url %>
           <%- end -%>
         <%- end -%>
+        <%- if proxy['preserve_host'] -%>
+    ProxyPreserveHost On
+        <%- end -%>
         <%- if proxy['setenv'] -%>
           <%- Array(proxy['setenv']).each do |setenv_var| -%>
     SetEnv <%= setenv_var %>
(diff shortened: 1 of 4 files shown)


templates/vhost/_directories.erb
```diff
@@ -493,6 +493,9 @@
     ProxyPassReverse <%= reverse_url %>
           <%- end -%>
         <%- end -%>
+        <%- if proxy['preserve_host'] -%>
```
ekohl, line 496: Should it deal with (in Puppet types) `Optional[Boolean]`? In other words: `undef` omits it, `true` is `On` and `false` is `Off`.

Tamerz: According to the Apache docs, the default value is "Off". So if you don't put anything at all, it is off.

https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypreservehost

Because of that, I thought this would keep things more simple.

ekohl: What if you set `ProxyPreserveHost On` at the vhost level?

Tamerz: That is a very good point. I'll update this tomorrow, thanks for catching it.

Tamerz:
I still am looking to get this added.

LukasAud:
Moving this back to active PRs. Thanks for letting us know. We are currently a bit busy but we will try to take a look at this as soon as possible. Thank you for your patience.

david22swan:
@Tamerz Sorry for the wait on review.
Anyway while the change look's ok to my eyes I have a small concern that the change you agreed with @ekohl  does not seem to have been made? Or at least not from what I can tell.
If you could get back to me on this it would be great

ekohl (changes requested):
@Tamerz any chance you can get back to this?

Tamerz:
I apologize but I think it is beyond me. The module is just too complicated and tests take too long to run. If I do it, I want to do it right and I know I don't have the bandwidth. I'll close this out and put in a feature request instead.

PR #2217, opened 2022-03-10
Add vhost "enabled" option to enable or disable it

BDelacour:
Hello,

I need to manually handle if a vhost is enabled and I don't want Puppet to overwrite it.
Coupled with `apache::purge_vhost_dir: false`, it works great !

Thank you !


david22swan:
@BDelacour Hey, sorry it's taken so long to comment on this
Anyway the change look's ok to me, but would prefer it to have a more descriptive name, i.e. `vhost_enabled`, and some documentation added to it at the top alongside what already exists for the other variables, so that it is represented in the documentation.

BDelacour:
Hey ! Sorry I left this PR stale for some time. I'll update it soon

PR #2221, opened 2022-03-24
Support ProxyPassMatch within a LocationMatch container

skylar2-uw:
Apache httpd supports placing `ProxyPass` and `ProxyPassMatch` directives either at virtual host scope, or inside a `Location`/`LocationMatch` container. When placed inside a container, the path parameter is omitted and inherited from the container's path. For `Location` containers, this is a simple URI, while `LocationMatch` allows the path to be regex. While `ProxyPass` can go into a `LocationMatch` container, the inherited path is interpreted as a string rather than a regex, which limits its utility.

While the puppetlabs-apache module (via its `apache::vhost` type) supports both `ProxyPass` and `ProxyPassMatch` at virtual host scope, currently it only supports `ProxyPass` inside a `Location`/`LocationMatch` container, which means that it isn't possible to use the regex-aware `LocationMatch` container type. This pull request enhances the `vhost/_directories.erb` template to support the same `proxy_pass_match` dictionary at `Location`/`LocationMatch` scope that `apache::vhost` type supports at virtual host scope.

I couldn't find any documentation for the existing `ProxyPass` support in `apache::vhost`'s `directories` array, so I added a short blurb to the `apache::vhost` Puppet Strings.

I added what I thought would be a decent simple test case, though I confess I'm not a rspec expert so it might bear special review.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1351,6 +1351,8 @@
 #    Values: `directory`, `files`, `proxy`, `location`, `directorymatch`, `filesmatch`, 
 #   `proxymatch` or `locationmatch`. If you set `provider` to `directorymatch`, it 
 #   uses the keyword `DirectoryMatch` in the Apache config file.<br />
+#   proxy_pass and proxy_pass_match are supported like their parameters to apache::vhost, and will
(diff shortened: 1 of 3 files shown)


skylar2-uw:
Will close and resubmit w/ correct branch name

PR #2226, opened 2022-03-29
Drop Apache 2.0 compatibility code

ekohl:
It's safe to assume the version is >= 2.2 so the conditional can be dropped.

### DIFF
--- a/manifests/mod/proxy_connect.pp
+++ b/manifests/mod/proxy_connect.pp
@@ -1,18 +1,9 @@
 # @summary
 #   Installs `mod_proxy_connect`.
-# 
-# @param apache_version
-#   Used to verify that the Apache version you have requested is compatible with the module.
-# 
 # @see https://httpd.apache.org/docs/current/mod/mod_proxy_connect.html for additional documentation.
 #
-class apache::mod::proxy_connect (
-  $apache_version  = undef,
-) {
+class apache::mod::proxy_connect {
   include apache
-  $_apache_version = pick($apache_version, $apache::apache_version)
-  if versioncmp($_apache_version, '2.2') >= 0 {
-    Class['::apache::mod::proxy'] -> Class['::apache::mod::proxy_connect']
-    ::apache::mod { 'proxy_connect': }
-  }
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

manifests/mod/proxy_connect.pp
```diff
@@ -2,17 +2,14 @@
 #   Installs `mod_proxy_connect`.
 # 
 # @param apache_version
-#   Used to verify that the Apache version you have requested is compatible with the module.
+#   Deprecated and unused. Remains here for compatibility.
 # 
 # @see https://httpd.apache.org/docs/current/mod/mod_proxy_connect.html for additional documentation.
 #
 class apache::mod::proxy_connect (
   $apache_version  = undef,
```
ekohl, line 10: I kept this variable in to not make it a breaking change. Silently ignoring it is safe, but it should be removed with the next major release.

bastelfreak: should we open a PR that drops it? I'm quite sure otherwise we will forget it.

bastelfreak (dismissed):

manifests/mod/proxy_connect.pp
ekohl: I thought about the same thing.

smortex (approved):

chelnak:
@david22swan This looks good to go to me but I'm wondering if we should include the breaking change mentioned by @ekohl above. There is no real reason we can't do a major release.

What do you think?

david22swan:
@ekohl @chelnak Seeing as we have a breaking change already waiting for a release with the removal of OS compatibility code I'd say go for it and remove it.

chelnak:
I'm down with that.

@ekohl Feel free to remove the variable in this PR 😄

ekohl:
Updated to drop the parameter.

bastelfreak (approved):

david22swan (approved):

david22swan:
@ekohl Thanks again for the work :)

PR #2227, opened 2022-03-29
Restructure MPM disabling

ekohl:
This takes the approach of listing the disabling steps once rather than repeating it for every mpm.

I wasn't quite sure on the prefork disabling. Does prefork need to stay enabled for itk?

### DIFF
--- a/manifests/mpm.pp
+++ b/manifests/mpm.pp
@@ -85,9 +85,6 @@ define apache::mpm (
       }
 
       if $mpm == 'itk' {
-        include apache::mpm::disable_mpm_event
-        include apache::mpm::disable_mpm_worker
-
         package { 'libapache2-mpm-itk':
           ensure => present,
           before => [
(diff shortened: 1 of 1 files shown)


david22swan:
Not sure about this change, while it cuts down on repetition it may also make the code harder for new users to read through. In addition, look's like it's causing a mass failure in the spec tests.

ekohl:
> Not sure about this change, while it cuts down on repetition it may also make the code harder for new users to read through

Funny, I had great problems with the duplication and think this is much easier to follow. My logic is:
```
If MPM is used
  configure it
else
  remove it
end
```

If there is no configuring it ends up being
```
if not MPM is used
  remove it
end
```

And the mass failures can easily be explained by my syntax error. Due to https://tickets.puppetlabs.com/browse/MODULES-11161 I can't run PDK module tests locally so I have to rely on CI. Sadly it makes contributing to Puppetlabs modules really hard for me. Normally I ignore any PDK module but puppetlabs modules are hard to ignore.

manifests/mpm.pp
```diff
@@ -97,14 +94,16 @@
         }
       }
 
-      if $mpm == 'prefork' {
-        include apache::mpm::disable_mpm_event
+      unless $mpm in ['itk', 'prefork'] {
```
david22swan, line 97: Is this right?
Your replacing the _if_ `$mpm` equals with an _unless_ `$mpm` equals.

Doesn't this now mean that `include apache::mpm::disable_mpm_prefork` won't be run if $mpm equals either of the two, where before it was run.

david22swan:
Looking back at this, and after confusing myself and then realising mistake. This looks like a good change.
Thank's for putting in the work :)

PR #2231, opened 2022-04-01
pdksync - (GH-cat-8) Move CentOS 8 support to CentOS Stream 8

david22swan:
(GH-cat-8) Move CentOS 8 support to CentOS Stream 8
pdk version: `2.3.0`

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -29,8 +29,8 @@
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
-        "7",
-        "8"
+        "Stream8",
+        "7"
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


metadata.json
```diff
@@ -29,8 +29,8 @@
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
-        "7",
-        "8"
+        "Stream8",
```
ekohl, line 32: For what it's worth, in Vox Pupuli we simply use `8`. This is because Facter also uses `8` as the fact `operatingsystemrelease` and CentOS Linux 8 is EOL anyway.

ekohl (changes requested):

metadata.json
david22swan: You've raised a valid point. Likely was a little to hasty of me to start this work. Will halt unless I hear something from higher up on either way.

ekohl: I think it's still good to move to Stream 8 but what we did was replace the interpretation of 8 to mean stream:
* https://github.com/voxpupuli/beaker-hostgenerator/commit/a4bdd3b3c51ccc65d6e75a9fa245a2f240b0d12d
* https://github.com/voxpupuli/facterdb/issues/164 is open to do the same for FacterDB

PR #2233, opened 2022-04-04
Add support for PassengerPreloadBundler

smortex:
The latest passenger include a new option to load ruby bundles earlier,
working around deployment issues.  See:
* https://github.com/phusion/passenger/issues/2410
* https://github.com/phusion/passenger/issues/2409

### DIFF
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -120,6 +120,9 @@
 # @param passenger_load_shell_envvars
 #   Enables or disables the loading of shell environment variables before spawning the application.
 #
+# @param passenger_preload_bundler
+#   Enables or disables loading bundler before loading your Ruby app.
+#
 # @param passenger_log_file
 #   File path to log file. By default Passenger log messages are written to the Apache global error log.
 #
@@ -343,6 +346,7 @@ class apache::mod::passenger (
   $passenger_installed_version                                                               = undef,
   $passenger_instance_registry_dir                                                           = undef,
   $passenger_load_shell_envvars                                                              = undef,
+  Optional[Boolean] $passenger_preload_bundler                                               = undef,
   Optional[Stdlib::Absolutepath] $passenger_log_file                                         = undef,
   $passenger_log_level                                                                       = undef,
   $passenger_lve_min_uid                                                                     = undef,
@@ -544,6 +548,11 @@ class apache::mod::passenger (
     if $passenger_load_shell_envvars {
       if (versioncmp($passenger_installed_version, '4.0.20') < 0) {
         fail("Passenger config option :: passenger_load_shell_envvars is not introduced until version 4.0.20 :: ${passenger_installed_version} is the version reported")
+      }
+    }
+    if $passenger_preload_bundler {
+      if (versioncmp($passenger_installed_version, '6.0.13') < 0) {
+        fail("Passenger config option :: passenger_preload_bundler is not introduced until version 6.0.13 :: ${passenger_installed_version} is the version reported")
       }
     }
     if $passenger_log_file {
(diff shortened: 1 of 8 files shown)


smortex:
@chelnak not sure about what "community-labeller" is supposed to check (failed test in first commit)… I labelled the issue and re-run the test but the outcome was the same :shrug: Then I saw CI was broken, so I added a commit to fix this and that "community-labeller" issue did not trigger.  I see you authored this GHA maybe you are interested in this info?

ekohl:
@smortex I think https://github.com/puppetlabs/puppetlabs-apache/pull/2232 is supposed to fix it. If this commit is included in your PR then I think we can conclude https://github.com/puppetlabs/puppetlabs-apache/pull/2232 didn't fix it.

chelnak:
That's the one @ekohl !

@smortex It turns out getting the right scope for this sort of thing is harder than I thought. However it should be fixed in the commit mentioned above.

Thanks for pointing it out 👍🙂

manifests/mod/passenger.pp
```diff
@@ -546,6 +550,11 @@
         fail("Passenger config option :: passenger_load_shell_envvars is not introduced until version 4.0.20 :: ${passenger_installed_version} is the version reported")
       }
     }
+    if $passenger_preload_bundler {
+      if (versioncmp($passenger_preload_bundler, '6.0.13') < 0) {
```
david22swan, line 554: Shouldn't this line be `passenger_installed_version` not `passenger_preload_bundler`?

david22swan (changes requested):
Other than the above, this look's good to me.

manifests/mod/passenger.pp
smortex: Meh… completion operator error :facepalm: 

Thanks, I amended the commit to fix this!

spec/acceptance/mod_php_spec.rb
```diff
@@ -44,7 +44,7 @@ class { 'apache::mod::php': }
       describe file("#{apache_hash['mod_dir']}/php7.4.conf") do
         it { is_expected.to contain 'DirectoryIndex index.php' }
       end
-    elsif os[:family] == 'redhat' && os[:release] =~ %r{^8\.}
+    elsif os[:family] == 'redhat' && os[:release] =~ %r{^8($|\.)}
```
bastelfreak, line 47: looks a bit complicated. `%r{^8}` should be fine?

smortex: Quite agree, at the current rate of RedHat releases I think we can ignore the issues that might arise the release following RedHat 79 :grin: …

smortex: `/^8\b/` should also do it as expected

david22swan (approved):
LGTM

manifests/mod/passenger.pp
```diff
@@ -343,6 +346,7 @@
   $passenger_installed_version                                                               = undef,
   $passenger_instance_registry_dir                                                           = undef,
   $passenger_load_shell_envvars                                                              = undef,
+  Optional[Boolean] $passenger_preload_bundler                                               = undef,
```
bastelfreak, line 349: Wouldn't false be the better default?
```suggestion
  Boolean $passenger_preload_bundler                                                         = false,
```

ekohl: No. Right now `undef` means "don't output it at all" and the boolean translates to `On`/`Off`. This is the correct behavior IMHO, especially since older Passenger versions don't support this so it'd fail on an unexpected statement.

ekohl (approved):
Not sure what's up with the CLA check.

david22swan:
@smortex Thanks for putting the work in :)

As for the CLA test, they do that sometimes

PR #2235, opened 2022-05-17
(GH-cat-9) Update module to match current syntax standard

david22swan:
Code now compliant with the following rules:
- manifest_whitespace_closing_bracket_after
- relative_classname_inclusion
- relative_classname_reference
- version_conparison
- legacy_facts
- top_scope_facts
- parameter_documentation
- parameter_types


spec/spec_helper_local.rb
```diff
@@ -60,117 +60,226 @@
 shared_context 'Fedora 17' do
   let :facts do
     {
+      # Legacy
       id: 'root',
```
ekohl, line 64: Where is this even used? Can't this fact (and the modern version of it) simply be dropped instead?

spec/spec_helper_local.rb
```diff
@@ -60,117 +60,226 @@
 shared_context 'Fedora 17' do
```
ekohl, line 60: I'd consider dropping this and opened https://github.com/puppetlabs/puppetlabs-apache/pull/2238. You may want to merge that first and rebase on it.

spec/classes/mod/jk_spec.rb
```diff
@@ -67,7 +67,7 @@
     include_examples 'Debian 11'
 
     context 'with only required facts and default parameters' do
-      let(:facts) { super().merge('ipaddress' => default_ip) }
+      let(:facts) { super().merge(networking: { ip: default_ip }) }
```
ekohl, line 70: This breaks things and you need to do a deep merge. In VP we have a helper for this: `override_facts()`. https://github.com/voxpupuli/voxpupuli-test#fact-handling describes the reasoning. Ideally this would be easier (maybe something in rspec-puppet?), but your current approach remotes all other networking facts (like FQDN).

david22swan: No other facts are used but I see your point, will update the code to account

spec/spec_helper_local.rb
david22swan: Will likely remove at the end, have left it in place as I complete the rest of the code changes

spec/classes/mod/proxy_html_spec.rb
```diff
@@ -20,7 +20,7 @@
 
     context 'on i386' do
       let(:facts) do
-        super().merge(hardwaremodel: 'i686',
+        super().merge(os: { family: 'Debian', name: 'Debian', release: { full: '11', major: '11' }, hardware: 'i686', },
```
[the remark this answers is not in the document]
ekohl: I think it would be good to fix https://github.com/voxpupuli/rspec-puppet-facts/issues/112 and solve it at that level. It's already pulled in and it's probably the right place to fix it (or rspec-puppet itself). IMHO that would be best for the ecosystem. Fixing it in voxpupuli-test was just my way of making quick progress.

I can't promise to have time for it, but if someone contributes the PR I'd be happy.

spec/spec_helper_local.rb
david22swan: Legacy facts have now been removed.

spec/classes/mod/proxy_html_spec.rb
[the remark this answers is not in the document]
david22swan: Was unable to include the voxpupuli-test gem added due to conflicts with existing gems so have simply added the override_facts code to the spec_helper_local.rb for now if you don't mind me copying your code.

david22swan:
@chelnak @ekohl @smortex Hoping to get a bunch of eyes on this as while none of the changes in it are very big, there are a lot of them.

manifests/mod/auth_openidc.pp
```diff
@@ -4,7 +4,7 @@
 # @see https://github.com/zmartzone/mod_auth_openidc for additional documentation.
 #
 class apache::mod::auth_openidc (
-) inherits ::apache::params {
+) inherits apache::params {
```
chelnak, line 7: Should this be on line 6 given that we are not passing in anything to the constructor?

manifests/custom_config.pp
```diff
@@ -49,18 +49,18 @@
...
-  Boolean $verify_config            = true,
-  $filename                         = undef,
-  $owner                            = undef,
-  $group                            = undef,
-  $file_mode                        = undef,
-  Boolean $show_diff                = true,
+  Enum['absent', 'present'] $ensure         = 'present',
+  String $confdir                           = $apache::confd_dir,
+  Optional[String] $content                 = undef,
+  Variant[Integer,String,Boolean] $priority = '25',
```
chelnak, line 55: In what case could priority be a boolean?

manifests/mod/event.pp
```diff
@@ -47,18 +47,18 @@
...
-  $threadsperchild        = '25',
-  $maxrequestsperchild    = '0',
-  $maxconnectionsperchild = undef,
-  $serverlimit            = '25',
-  $apache_version         = undef,
-  $threadlimit            = '64',
-  $listenbacklog          = '511',
+  Variant[String,Boolean] $startservers                       = '2',
+  Variant[String,Boolean] $maxclients                         = '150',
+  Optional[Variant[String,Boolean]] $maxrequestworkers        = undef,
```
chelnak, line 52: In what case would `maxrequestworkers` be a boolean? Would it make more sense to be `Variant[String, Int]`?

manifests/mod/event.pp
```diff
@@ -47,18 +47,18 @@
...
+  Variant[String,Boolean] $minsparethreads                    = '25',
+  Variant[String,Boolean] $maxsparethreads                    = '75',
+  Variant[String,Boolean] $threadsperchild                    = '25',
+  Variant[String,Boolean] $maxrequestsperchild                = '0',
+  Optional[Variant[String,Boolean]] $maxconnectionsperchild   = undef,
```
chelnak, line 57: In what case would `maxconnectionsperchild` be a boolean? Would it make more sense to be `Variant[String, Int]`?

manifests/mod/event.pp
```diff
@@ -47,18 +47,18 @@
...
+  Variant[String,Boolean] $serverlimit                        = '25',
```
chelnak, line 58: In what case would `serverlimit` be a boolean? Would it make more sense to be `Variant[String, Int]`?

manifests/mod/event.pp
```diff
@@ -47,18 +47,18 @@
...
+  Optional[String] $apache_version                            = undef,
+  Variant[String,Boolean]  $threadlimit                       = '64',
+  Variant[String,Boolean]  $listenbacklog                     = '511',
```
chelnak, line 61: Same question re Bool types

manifests/service.pp
```diff
@@ -3,11 +3,11 @@
...
 # @api private
 class apache::service (
-  $service_name           = $apache::params::service_name,
-  Boolean $service_enable = true,
-  $service_ensure         = 'running',
-  Boolean $service_manage = true,
-  $service_restart        = undef
+  String $service_name                    = $apache::params::service_name,
+  Boolean $service_enable                 = true,
+  Variant[Boolean,String] $service_ensure = 'running',
```
chelnak, line 8: What are the possible values here? Should it just be a string? Rather than the variant?

chelnak (commented):
Looks great! Just a few questions on typing.. but they might have reasonable answers!

manifests/custom_config.pp
david22swan: It can be set as boolean false to disable it and set the input to an empty string.

manifests/mod/auth_openidc.pp
david22swan: Think it reads better this way though, looks neater in my eyes at least.

manifests/custom_config.pp
david22swan: That kind of option seems to be pretty common in the code

manifests/mod/event.pp
david22swan: as above

manifests/mod/event.pp
david22swan: as above

manifests/service.pp
david22swan: service_ensure can be set too true or false in the code. There are examples of each in the tests

chelnak: https://github.com/puppetlabs/puppetlabs-apache/blob/babc21bb524d2bae88066fe8abeeb7e77ace960a/manifests/service.pp#L16

Oh i see here... that doesn't seem right to me but it's a conversation that doesn't fall in the scope of this PR so I'm happy. That goes for the other questions too!

chelnak (dismissed):

chelnak (approved):

PR #2241, opened 2022-06-09
timeout: switch from String to Integer datatype

cocker-cc:
Put Parameter $apache::timeout on Datatype Integer instead of String.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -491,7 +491,7 @@ class apache (
   Optional[Enum['On', 'Off', 'on', 'off']] $ldap_verify_server_cert     = undef,
   Optional[String] $ldap_trusted_mode                                   = undef,
   Boolean $error_documents                                              = false,
-  String $timeout                                                       = '60',
+  Variant[Integer[0], Pattern[/^[0-9]+$/]] $timeout                     = '60',
   String $httpd_dir                                                     = $apache::params::httpd_dir,
   String $server_root                                                   = $apache::params::server_root,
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -491,7 +491,7 @@
   Optional[Enum['On', 'Off', 'on', 'off']] $ldap_verify_server_cert     = undef,
   Optional[String] $ldap_trusted_mode                                   = undef,
   Boolean $error_documents                                              = false,
-  String $timeout                                                       = '60',
+  Integer $timeout                                                      = 60,
```
ekohl, line 494: IMHO an integer would be better, but it is a breaking change. Would it be better to make it
```suggestion
  Variant[Integer[0], String[1]] $timeout                                                      = 60,
```

(and then properly align it, that's hard in GH)

cocker-cc: What do you think of `Variant[Integer[0], Pattern[/^[0-9]+$/]]`?

bastelfreak: FYI: The next release will be a major one, so if we want to break things, now is a good time

cocker-cc: > FYI: The next release will be a major one, so if we want to break things, now is a good time

Adding Datatypes to the Parameters actually was a good Idea, but IMHO it was done only half-hearted.  This should be improved for the next Major-Release.

ekohl:
I'm not sure I agree with this being closed. It actually narrows the data types further to disallow misconfigurations.

PR #2243, opened 2022-06-09
Acceptance tests: correct EPEL usage

ekohl:
Migrates from stahnma/epel to puppet/epel (after it was transferred) and also stops using EPEL 7 on EL8. That is unlikely to work and can easily breaking things. Perhaps it's why the EL8 tests are often timing out in CI.


bastelfreak (approved):

ekohl:
Ah, this fails because `puppet/epel` dropped EL6 support. Needs a version pin.

david22swan (approved):

david22swan:
@ekohl Thx for getting a fix up so fast. Wasn't sure when I would have gotten around to it.

PR #2248, opened 2022-06-13
[Issue 2247] fix datatype of vhost:headers from String to Array

chillinger:
Fix wrong introduction of Dataype String to headers parameter
The code requires the parameter to be Array[String] to work properly

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1853,7 +1853,7 @@ define apache::vhost (
   Optional[Variant[Array[String],String]] $redirectmatch_status                       = undef,
   Optional[Variant[Array[String],String]] $redirectmatch_regexp                       = undef,
   Optional[Variant[Array[String],String]] $redirectmatch_dest                         = undef,
-  Optional[String] $headers                                                           = undef,
+  Array[String] $headers                                                              = [],
   Optional[Array[String]] $request_headers                                            = undef,
   Optional[Array[String]] $filters                                                    = undef,
   Optional[Array] $rewrites                                                           = undef,
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2260,7 +2260,7 @@ define apache::vhost (
   }
 
   # Check if mod_headers is required to process $headers/$request_headers
-  if ($headers and ! empty($headers)) or ( $request_headers and ! empty($request_headers)){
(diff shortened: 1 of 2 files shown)


david22swan:
From what I can see this looks to be correct, but could you also update the example in the description so that it matches

chillinger:
@david22swan thanks for the review, description is updated

david22swan:
@chillinger Look's like your getting some unit test failures.

manifests/vhost.pp
```diff
@@ -1853,7 +1853,7 @@
   Optional[Variant[Array[String],String]] $redirectmatch_status                       = undef,
   Optional[Variant[Array[String],String]] $redirectmatch_regexp                       = undef,
   Optional[Variant[Array[String],String]] $redirectmatch_dest                         = undef,
-  Optional[String] $headers                                                           = undef,
+  Optional[Array[String]] $headers                                                    = undef,
```
ekohl, line 1856: If you're changing it to an array, I'd say drop the `Optional` part and make it an empty array by default. I'd also narrow it to non-empty strings:
```suggestion
  Array[String[1]] $headers                                                    = [],
```

Then you can simplify `templates/vhost/_header.erb` a lot too.

You can also drop the `if $headers` and stick to `! empty($headers)` on line 2263.

ekohl (changes requested):

manifests/vhost.pp
chillinger: @ekohl  I thought about that, but realised the patch the way I did for the following reasons:
- I only have the time to actually fix the bug which breaks our puppet runs, I did not want to rework more than really necessary
- every other array type parameter in the module, including $request_headers is an Optional[Array[String]]. I wanted to stay consistent

For those reasons I would like this fix to be accepted. If the datatype should be changed to Array with default [], I would like this seen as an independent rework issue

david22swan:
@chillinger Apologies, but after some discussion within the team we have decided to go ahead with the changes proposed by @ekohl rather than your own, as they are more comprehensive and represent a larger improvement to the module.
From what I can see they should still solve your issue and we hope to get them merged in soon.
I am sorry for any inconvenience this has caused you, and while we could not merge your work at this time we hope to see more from you in the future ad thank you for bringing this issue to our attention.

chillinger:
@david22swan @ekohl  no problem at all, I'm happy an even better solution arose from my initial proposal!

PR #2250, opened 2022-06-15
Stricter data types and better module inclusion in vhost.pp

ekohl:
There is no need to call defined() before calling include since include handles it already.

It also consistently checks if $ensure is set to present before including modules.


ekohl:
So the scope of this PR expanded a bit. It continues the trend that #2251 started. I've tried to keep it reviewable by having small commits.

david22swan:
@ekohl Look's like the error is coming from `aliases`

```
  2267) apache::vhost os-independent items on debian-11-x86_64  set everything! is expected to contain Concat::Fragment[rspec.example.com-apache-header] with content =~ /^MDomain example\.com example\.net auto$/
        Failure/Error:
          is_expected.to contain_concat__fragment('rspec.example.com-apache-header').with(
            content: %r{^MDomain example\.com example\.net auto$},
          )
        Puppet::PreformattedError:
          Evaluation Error: Error while evaluating a Resource Statement, Apache::Vhost[rspec.example.com]: parameter 'aliases' index 0 expects a Hash value, got String (line: 3) on node fv-az180-264.aibneqh1mxpuhl3tnnuy2ifbce.bx.internal.cloudapp.net
        # /opt/hostedtoolcache/Ruby/2.5.9/x64/lib/ruby/2.5.0/benchmark.rb:308:in `realtime'
        # ./spec/defines/vhost_spec.rb:1715:in `block (6 levels) in <top (required)>'
        # ./vendor/bundle/ruby/2.5.0/bin/rspec:23:in `load'
        # ./vendor/bundle/ruby/2.5.0/bin/rspec:23:in `<top (required)>'
        # ./vendor/bundle/ruby/2.5.0/bin/bundle:23:in `load'
        # ./vendor/bundle/ruby/2.5.0/bin/bundle:23:in `<main>'
        # ------------------
        # --- Caused by: ---
        # Puppet::ParseError:
        #   Apache::Vhost[rspec.example.com]: parameter 'aliases' index 0 expects a Hash value, got String
        #   /opt/hostedtoolcache/Ruby/2.5.9/x64/lib/ruby/2.5.0/benchmark.rb:308:in `realtime'
```
Think this is the only one, just seems to trigger a lot of tests

ekohl:
Yes, I had some trouble finding all instances. I do wonder how it ever worked. It was introduced in f1d64a0a0b71af0102d10da69e333aa9ac5a15f5, but it doesn't look like it ever produced anything valid by passing a plain string. Let's see if this version is better. I also had to fix a test related to directories.

ekohl:
The rubocop violation looks weird to me. Possibly an issue with Rubocop? I don't quite see it.

david22swan:
@ekohl In regards to the acceptance test....... it should probably be removed now yes if the behaviour it is expected to catch will no longer occur with your improvement in place.
Though would like a more descriptive title and comment for what what seems a substantial change in behaviour.

In regards to the Rubocop error........ Rubocop is just being annoying. Took a look and if you move aliases up so that it is not immediately followed by `directories` and there is another variable between them, the error goes away.
Don't know why, guess Rubocop doesn't like two arrays formatted in that manner, one after another?

ekohl:
I think I'll pull out the last 2 patches to their own PR. Would it make sense to even make it 3 PRs? The first commit, then a series of patches to localize the imports and then a third for the last 2 commits?

david22swan:
If you're willing to separate them out into three then that does sound good to me yes.
This PR does seem to be doing multiple different things and that did throw me for a bit when I first looked at it, so dividing it up and given each piece of work its own pr may be for the best.

ekohl:
Setting this back to a draft. I've submitted https://github.com/puppetlabs/puppetlabs-apache/pull/2253 as a first PR. Once it's merged I'll submit the next.

ekohl:
Part 2: https://github.com/puppetlabs/puppetlabs-apache/pull/2254

ekohl:
And the last part: https://github.com/puppetlabs/puppetlabs-apache/pull/2255

PR #2253, opened 2022-06-23
Use a stricter data type on apache::vhost::aliases

ekohl:
This also implements a real test on the rendered template. It appears it was passing in something invalid ever since it was introduced in f1d64a0a0b71af0102d10da69e333aa9ac5a15f5.


david22swan:
@ekohl Look's like the rubocop failure is still occurring, swapping the position of `aliases` and putting a another variable, or a few, between it and `directories` should fix it.

ekohl:
Looks like it passes that now. Let's see for the real unit tests.

david22swan (approved):
LGTM

david22swan:
Got one transient failure and one setup failure, but happy to merge :)
Looking forward to getting the next two in as well :)

PR #2254, opened 2022-06-23
Clean up includes and templates in vhost.pp

ekohl:
This localizes the includes in vhost.pp which removes some duplicated logic. Sometimes the logic was also slightly different. This keeps it consistent.

It also only includes modules if ensure is set to present. This allows users to toggle the ensure easily and only having the minimal amount of modules present.


ekohl:
Acceptance is still running, but this looks like it'll be green.

david22swan (approved):
LGTM

david22swan:
And that's another one in, sorry for the wait on merge, was off yesterday

ekohl:
No worries, thanks for the merge.

PR #2255, opened 2022-06-28
Automatically include modules used in vhost directories

ekohl:
The data passed to the directories parameter of apache::vhost often implies a module should be present. This change inspects the passed data and automatically includes modules if needed. This makes it easier to use via Hiera while disabling all default mods.

Because of this change one acceptance test becomes obsolete: it no longer fails and just works.

To make this easier the data type is simplified. This means users need to be a bit stricter about their input, but they need to worry less about including the right modules.

This replaces https://github.com/puppetlabs/puppetlabs-apache/pull/2250.

### DIFF
--- a/manifests/default_mods.pp
+++ b/manifests/default_mods.pp
@@ -157,8 +157,7 @@ class apache::default_mods (
     }
 
     include apache::mod::authz_user
-
-    ::apache::mod { 'authz_groupfile': }
+    include apache::mod::authz_groupfile
     include apache::mod::env
   } elsif $mods {
     ::apache::default_mods::load { $mods: }
--- a/manifests/mod/authz_groupfile.pp
+++ b/manifests/mod/authz_groupfile.pp
@@ -1,0 +1,9 @@
+# @summary
+#   Installs `mod_authz_groupfile`
+#
+# @see https://httpd.apache.org/docs/current/mod/mod_authz_groupfile.html for additional documentation.
+#
+class apache::mod::authz_groupfile {
+  include apache
+  apache::mod { 'authz_groupfile': }
+}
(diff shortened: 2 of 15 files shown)


david22swan:
@ekohl Look's like another few spec failures have shown their heads, almost there 🤞

ekohl:
Tests have been running for a long time. I suspect there is a failure in many cases which means it generates a ton of errors. I pushed a commit that massively reduces the spec tests so hopefully I get a faster result.

ekohl:
Ah, it doesn't run the vhost tests on non-Debian. That's surprising since other code suggests it does run on that. Will update again.

ekohl:
This is much easier to iterate on.

ekohl:
Well, that's green. Removing the temporary commit to see if it all fits together.

david22swan:
@ekohl Seems to be a few spec test failures left. All coming from the same section:
```
  1) apache::vhost os-independent items on scientific-6-x86_64  not everything can be set together... is expected to contain Concat::Fragment[rspec.example.com-directories] with content =~ /^\s+Allow from 127\.0\.0\.1$/
     Failure/Error:
       is_expected.to contain_concat__fragment('rspec.example.com-directories').with(
         content: %r{^\s+Allow from 127\.0\.0\.1$},
       )
       expected that the catalogue would contain Concat::Fragment[rspec.example.com-directories]
     # ./spec/defines/vhost_spec.rb:2140:in `block (6 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.7.0/bin/rspec:23:in `load'
     # ./vendor/bundle/ruby/2.7.0/bin/rspec:23:in `<top (required)>'
     # /opt/hostedtoolcache/Ruby/2.7.6/x64/bin/bundle:23:in `load'
     # /opt/hostedtoolcache/Ruby/2.7.6/x64/bin/bundle:23:in `<main>'
```

david22swan:
No need to apologize, I'm liking all the changes

ekohl:
Rebased to resolve the conflicts. Let's see if the tests pass.

ekohl:
I'm not quite sure why GH actions don't run.

david22swan (approved):
LGTM

david22swan:
Ok, everything's gone green so I'm happy to merge.

PR #2256, opened 2022-06-29
Group assertions in vhost spec

ekohl:
This uses chaining to remove a lot of separate expectations. The big benefit is that it's a lot quicker to run and if one thing fails (like the whole catalog fails to compile) there is just a single error for the fragment rather than one for each test. It's also a lot shorter.

I'm submitting this as a separate PR so I'm testing on a clean slate. I do want this in https://github.com/puppetlabs/puppetlabs-apache/pull/2255 because it makes moving things easier.

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -602,26 +602,12 @@ describe 'apache::vhost', type: :define do
                                                                                     'path' => "/etc/#{apache_name}/sites-enabled/30-rspec.example.com.conf")
             }
           end
-          it { is_expected.to contain_concat__fragment('rspec.example.com-apache-header') }
           it {
-            is_expected.to contain_concat__fragment('rspec.example.com-apache-header').with(
-              content: %r{^\s+LimitRequestFieldSize 8190$},
-            )
-          }
-          it {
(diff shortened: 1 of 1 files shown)


david22swan (dismissed):
Aside from the rubocop issue this looks good to me.

A little concerned over the way that failures that come after the first one, within the same block, may be hidden, but think the trade off is worth it.

ekohl:
> A little concerned over the way that failures that come after the first one, within the same block, may be hidden, but think the trade off is worth it.

Yes, worst case you may need multiple iterations where you fix it but IMHO it's also worth the trade off.

bastelfreak (approved):

ekohl:
I must say that I wrote this on top of #2255 and then cherry picked it. I hope I didn't miss anything in the rebase. It will also be interesting to see the timing difference between `main` and this PR.

ekohl:
https://github.com/puppetlabs/puppetlabs-apache/runs/7110407048?check_suite_focus=true was 16 minutes vs https://github.com/puppetlabs/puppetlabs-apache/runs/7103340284?check_suite_focus=true with 20 minutes. Very unscientific, but it shaves off a few minutes, or about 20%. That's better than I was hoping for.

david22swan (approved):
LGTM

PR #2265, opened 2022-07-27
Allow multiple scopes for Scope in Apache::OIDCSettings

### LINKED ISSUE #2262: Allow multiple scopes for Scope in Apache::OIDCSettings
## Use Case
Code in question: https://github.com/puppetlabs/puppetlabs-apache/blob/babc21bb524d2bae88066fe8abeeb7e77ace960a/types/oidcsettings.pp#L20

The OIDCScope setting can be used with multiple scopes.
Per mod_auth_openidc:
```
# Define the OpenID Connect scope that is requested from the OP (eg. "openid email profile").
# When not defined, the bare minimal scope "openid" is used.
# NB: multiple scope values must be enclosed in a single pair of double quotes 
# NB: this can be overridden on a per-OP basis in the .conf file using the key: scope
#OIDCScope "<scope(s)-separated-by-spaces-and-enclosed-in-double-quotes>"
```
However, quoted strings are not currently supported in this Puppet module (per the regex pattern), nor does it automatically enclose the value in quotes in the output config.

## Describe the Solution You Would Like
Change the regex pattern to match quoted strings.

## Describe Alternatives You've Considered
Enclose the output string in quotes.

## Additional Context
N/A

jjackzhn:
Fixes #2262


jjackzhn:
Just rebased.

david22swan (approved):
LGTM

Sorry for taking so long to get back on the updated changes but this looks good to merge :)
Thanks for putting in the work.

PR #2266, opened 2022-07-28
Define apache_name via a let block

ekohl:
Defining variables in rspec is dangerous since they end up being defined on classes. That means that in a for loop it may end up being defined as the last running version. A let block provides proper variable scoping.

This wasn't a problem until fb0d0782d8221741474f8170dacf522f6ce32b69 where it ran spec tests on all operating systems. Then it became dependent on the order of the supported operating systems.

Fixes: fb0d0782d8221741474f8170dacf522f6ce32b69

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -5,14 +5,7 @@ require 'spec_helper'
 describe 'apache::vhost', type: :define do
   describe 'os-independent items' do
     on_supported_os.each do |os, facts|
-      apache_name = case facts[:os]['family']
-                    when 'RedHat'
-                      'httpd'
-                    when 'Debian'
-                      'apache2'
-                    else
-                      'apache2'
-                    end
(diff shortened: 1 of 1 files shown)


bastelfreak (dismissed):

david22swan:
Change look's good to me and it look's like the original failures have been fixed.
A few more new ones have popped up however.

ekohl:
Hmm, those are odd. I pushed a change that may solve it.

bastelfreak (approved):

ekohl:
While reading through it I did wonder why we have OS specific conditions. It may be better to get rid of them so I opened https://github.com/puppetlabs/puppetlabs-apache/pull/2267 to drop mod_fastcgi.

ekohl:
> Hmm, those are odd. I pushed a change that may solve it.

Looks green now.

david22swan (approved):
LGTM

PR #2267, opened 2022-07-28
Drop mod_fastcgi support

ekohl:
mod_fastcgi isn't part of EL7 and beyond, nor in any current Debian/Ubuntu release. That's because upstream is dead. It also isn't part of RHEL itself and EPEL 6 is dead. Removing it from the codebase simplifies it.


bastelfreak (approved):

david22swan:
While this look's like a good change to me and is one I would love to get in I'm not sure we can merge it.
As a RedHat partner we are obligated to continue to support RHEL 6, and by extension Apache 2.2, until such time as it is definitively EOL.
If I could remove it I would, I have an entire branch saved away somewhere waiting for the day it can be rebased and all that excess code can be removed from the module.

ekohl:
> As a RedHat partner we are obligated to continue to support RHEL 6, and by extension Apache 2.2, until such time as it is definitively EOL.

But as far as I could see, mod_fastcgi is not in EL6 itself. At least, https://vault.centos.org/centos/6/os/x86_64/Packages/ doesn't list it (don't have a real RHEL 6 on hand). It looks like it was previously provided through external repositories. However, https://archives.fedoraproject.org/pub/archive/epel/6/x86_64/Packages/m/ also only lists mod_fcgid, not most_fastcgi. So really, where did it come from?

6c1ed75df6988dee291041c1bc4251f0f9525be9 stopped installing it on EL7, but I can't see where it came from on EL6. 48fb00e52a48ebb54363546dcab14dd9c32d707d added it, but also doesn't describe anything. https://github.com/puppetlabs/puppetlabs-apache/pull/295 does describe that it once came from official archives, but even in 2013 it was abandoned for 2 years.

I'd say that if it isn't in RHEL 6 itself, it's safe to drop. EPEL6 is EOL and not supported by Red Hat. Disclaimer: while I have my own employer provided Red Hat, I don't know the exact scope of partnerships.

david22swan:
From my perspective at least, it's more of that the alternative to it, `mod_proxy_fastcgi` is only available in 2.3 and upwards.
So if we remove it then anyone still using RHEL 6 and Apache 2.2, which I'm told there is, are unable to utilise this functionality or anything similar.

ekohl:
Generally speaking Red Hat is very strict in "if we don't provide the bits, we don't support it". I found https://actsupport.com/apache-php-fpm-mod-fastcgi/ which states:
> mod_fastcgi will not be in default CentOS repo, so please install rpmforge repo.

Considering rpmforge is dead and even no longer online, I really wonder who is going to use it. Worse, how are you going to support it if you can't even get it to run because you don't have the bits?

david22swan (approved):
Ok, you've convinced us.
Hope you don't mind waiting for the narrow PR to be merged first though.

ekohl:
No, that's a good idea. Rebasing this is trivial.

ekohl:
Rebased

david22swan (approved):
LGTM
Test failure is from a CI issue and not due to the code.

david22swan:
Thank's for the work.
This should be released on next Friday, since the release has so many changes have put a message out in this weeks blog and giving it some time before cutting it and pushing it out.

raybellis:
Please put this back in!   `mod_fastcgi` is still a supported package in FreeBSD!

ekohl:
Interesting. https://www.freshports.org/www/mod_fastcgi/ does have more information, like https://github.com/FastCGI-Archives which I wasn't able to find. Since I'm not able to test it, I think you'll have to come up with a PR @raybellis.

raybellis:
I don't have the means or head space to build a CI test, but I can assure you that the code that was removed in these commits was working perfectly well for us on FreeBSD. 

I do have a workaround using custom config blobs, so this is not critical, but please do consider that major Linux distros are not the only place people run Puppet.

ekohl:
My biggest concern is that mod_fastcgi is dead upstream and no development happens on it. I took that as a sign nobody would bother shipping it and I'm really surprised FreeBSD still does.

raybellis:
I took it as a sign that it doesn't need any more development on it ;-)

PR #2270, opened 2022-07-29
Allow configuring CRS paranoia level

Vincevrp:


Vincevrp:
@david22swan I've added some tests for the paranoia level parameters. Please let me know if this does not suffice.

david22swan (approved):
LGTM
Thanks for getting back so soon on the test's.
Once your other PRs are updated the same they should be good to go also.
Next release is planned for this Friday, so your changes should be out then.

PR #2272, opened 2022-07-29
 Allow configuring SecRequestBodyAccess and SecResponseBodyAccess

Vincevrp:


Vincevrp:
@david22swan tests have been added :)

spec/classes/mod/security_spec.rb
```diff
@@ -79,13 +79,17 @@
                 audit_log_type: 'Concurrent',
                 audit_log_storage_dir: '/var/log/httpd/audit',
                 secdefaultaction: 'deny,status:406,nolog,auditlog',
+                secrequestbodyaccess: 'On',
```
david22swan, line 82: Minor change, but for the test's could you set the values to something other than the default?

david22swan (changes requested):
LGTM except for on small thing

spec/classes/mod/security_spec.rb
Vincevrp: Changed the values and rebased

david22swan (approved):
LGTM
Thanks for getting back on the change :)

PR #2273, opened 2022-07-29
Allow overriding CRS anomaly threshold per vhost

Vincevrp:


Vincevrp:
@david22swan This one should also be good to go. Please let me know if you require any other changes.

david22swan (approved):
LGTM

On small comment on the new types though, the `Default` is only needed when declaring less than or equal to a given number. If you only gives a single value it assumes `Default` as the second one.

PR #2275, opened 2022-08-01
Allow vhosts to have a string priority again

ekohl:
In f41251e336fa3e7c31c19968ccee74121cf71e47 the type was narrowed to no longer allow strings, but this can cause problems.

Sorting is alphabetical and you need to format it for the correct sorting. So to make sure 2 loads before 10 you need to format it as 02.  While the vhost can do some printf style magic to change 2 to 02, it must then also know what the highest number is. Otherwise 100 is sorted before 20. By allowing strings, you allow the caller to fix this.

A type alias is introduced to reduce duplication and make it easier to track.

Fixes: f41251e336fa3e7c31c19968ccee74121cf71e47

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1730,7 +1730,7 @@ define apache::vhost (
   Optional[Integer] $ssl_stapling_timeout                                             = undef,
   Optional[Enum['on', 'off']] $ssl_stapling_return_errors                             = undef,
   Optional[String] $ssl_user_name                                                     = undef,
-  Optional[Variant[Integer, Boolean]] $priority                                       = undef,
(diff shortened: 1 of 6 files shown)


bastelfreak (approved):

ekohl:
Looks like I was not paying attention yesterday and failed to write proper tests for the type. Updated now.

ekohl:
@david22swan this is blocking our CI pipelines. I'd appreciate it if this could get a look.

david22swan (approved):
LGTM

david22swan:
Quick and easy merge.

ekohl:
It certainly was quick :)

PR #2277, opened 2022-08-04
Disable mod_php on EL9

ekohl:
EL 9 doesn't ship mod_php, so this is uninstallable. By making it a required parameter (which effectively was already the case) and setting the version to undef the catalog fails to compile by default. This still allows users to set it if they build a mod_php somehow.

EL 9 users should use php-fpm instead.

A draft right now to see which tests fail.

Fixes: edbdb65b06418779aa30ebe6cf0948007cc51a29

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -21,10 +21,11 @@
 #   UNIX group of the root user
 #
 # @param php_version
-#   The php version
+#   The php version. This is a required parameter, but optional allows showing a clear error message
 #
 # @param libphp_prefix
 #
+# @note Unsupported platforms: RedHat: 9
 class apache::mod::php (
   Optional[String] $package_name = undef,
   String $package_ensure         = 'present',
@@ -37,6 +38,10 @@ class apache::mod::php (
   Optional[String] $php_version  = $apache::params::php_version,
   String $libphp_prefix          = 'libphp'
 ) inherits apache::params {
+  unless $php_version {
+    fail("${facts['os']['name']} ${facts['os']['release']['major']} does not support mod_php")
+  }
+
   include apache
   if (versioncmp($php_version, '8') < 0) {
     $mod = "php${php_version}"
--- a/spec/acceptance/mod_php_spec.rb
+++ b/spec/acceptance/mod_php_spec.rb
@@ -3,101 +3,94 @@
 require 'spec_helper_acceptance'
 apache_hash = apache_settings_hash
 
-# Temporary exclusion pending resolution of https://github.com/puppetlabs/puppetlabs-apache/issues/2261
-unless os[:family] == 'redhat' && os[:release].to_i == 9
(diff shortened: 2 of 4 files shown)


spec/acceptance/mod_php_spec.rb
```diff
@@ -3,101 +3,98 @@
...
+      end
+    elsif os[:family] == 'ubuntu' && os[:release] == '18.04'
+      describe file("#{apache_hash['mod_dir']}/php7.2.conf") do
+        it { is_expected.to contain 'DirectoryIndex index.php' }
+      end
+    elsif os[:family] == 'ubuntu' && os[:release] == '20.04'
+      describe file("#{apache_hash['mod_dir']}/php7.4.conf") do
+        it { is_expected.to contain 'DirectoryIndex index.php' }
+      end
+    elsif os[:family] == 'redhat' && os[:release] =~ %r{^(9)\b}
```
david22swan, line 47: Since RHEL 9 is no longer being tested this should probably be removed.

manifests/mod/php.pp
```diff
@@ -34,7 +35,7 @@
   String $template               = 'apache/mod/php.conf.erb',
   Optional[String] $source       = undef,
   Optional[String] $root_group   = $apache::params::root_group,
-  Optional[String] $php_version  = $apache::params::php_version,
+  String[1] $php_version         = $apache::params::php_version,
```
david22swan, line 38: Rather than this would it not be better to put a check in to detect when its running on RHEL 9 and fail with a message stating that `php-fpm` should be used instead?

i.e.:
```
  if $facts['os']['name'] == 'Redhat' and $facts['os']['release']['major'] == '9' {
    fail("RHEL 9 does not support 'mod_php' and so 'php-fpm' must be used instead.")
  }
```

david22swan (changes requested):

manifests/mod/php.pp
david22swan: Make the reasoning and cause explicit?

ekohl: How about this solution: keeping it Optional, but treat `undef` as unsupported?

david22swan: Sound's good to me

david22swan (approved):
LGTM

Soon as the tests pass will merge in and restart release

david22swan (changes requested):
Ah, look's like you need to update `./spec/classes/mod/php_spec.rb:98` to expect a failure when run on Redhat 9.
Which is weird since it shouldn't ben running against it anymore?

ekohl:
> Which is weird since it shouldn't ben running against it anymore?

Unit tests run on all supported OSes (according to metadata.json)

bastelfreak (approved):

david22swan:
Ah right, the unsupported comment works on the acceptance test

david22swan (approved):
LGTM

david22swan:
:)
Thanks again for the work.
Time for the Release to go out!

PR #2280, opened 2022-08-05
Parameterize CRS DOS protection

Vincevrp:


templates/mod/security_crs.conf.erb
```diff
@@ -714,16 +714,17 @@ SecAction \
 #
 # Uncomment this rule to use this feature:
 #
+<% if @enable_dos_protection -%>
```
david22swan, line 717: Since this was previously included uncommented should the @enable_dos_protection not be set to `True` by default?

Vincevrp: @david22swan It depends. This module enabled it by default. However, it's optional in the default CRS configuration.

david22swan: From what I can tell, this has been enabled by default since the module was created.
As such I feel like it should continue onward as the same, where this to be merged in as it is the protection may be removed from people who actively want it in place without them noticing or being notified.

Vincevrp: I've changed the default to true.

david22swan (approved):
LGTM

PR #2281, opened 2022-08-05
Remove auth_kerb and nss from Debian Bullseye

ekohl:
libapache2-mod-auth-kerb is unavailable in Debian Bullseye[1]
libapache2-mod-auth-kerb was removed in Debian Buster[2]

[1]: https://packages.debian.org/buster/libapache2-mod-auth-kerb
[2]: https://packages.debian.org/stretch/libapache2-mod-nss

### DIFF
--- a/spec/spec_helper_local.rb
+++ b/spec/spec_helper_local.rb
@@ -37,6 +37,10 @@ shared_context 'a mod class, without including apache' do
   let(:facts) { on_supported_os['debian-10-x86_64'] }
 end
 
+shared_context 'Debian 10' do
+  let(:facts) { on_supported_os['debian-10-x86_64'] }
+end
+
 shared_context 'Debian 11' do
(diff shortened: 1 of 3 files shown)


spec/spec_helper_local.rb
```diff
@@ -34,6 +34,10 @@
 end
 
 shared_context 'a mod class, without including apache' do
```
ekohl, line 36: I think we should get rid of this, but it's out of scope for this PR.

david22swan (approved):
LGTM

Took a look and yeh, look's like the packages aren't included so happy to merge

david22swan (changes requested):
Once the syntax is fixed I mean

bastelfreak (approved):

ekohl:
I think the failure is unrelated. @david22swan mind having another look?

david22swan (approved):
LGTM

PR #2283, opened 2022-08-09
Manage DNF module for mod_auth_openidc

ekohl:
On EL 8 mod_auth_openidc is in a DNF module that must be enabled. Otherwise the package is uninstallable. This is verified by adding an acceptance test for the class.

The inheritance on apache::params is removed since it was redundant. That is only needed if a class parameter uses apache::params.

$oidc_settings on apache::vhost is changed to have a default. The template expects one and With that it's impossible to miscompile. The alternative is to implement a fail() inside the code if it is empty, but this provides some safety.

### DIFF
--- a/manifests/mod/auth_openidc.pp
+++ b/manifests/mod/auth_openidc.pp
@@ -5,6 +5,7 @@
 #
 class apache::mod::auth_openidc inherits apache::params {
   include apache
+  include apache::mod::authn_core
   include apache::mod::authz_user
   apache::mod { 'auth_openidc': }
 }
--- a/manifests/mod/auth_openidc.pp
+++ b/manifests/mod/auth_openidc.pp
@@ -1,11 +1,30 @@
 # @summary
 #   Installs and configures `mod_auth_openidc`.
-# 
-# @see https://github.com/zmartzone/mod_auth_openidc for additional documentation.
 #
-class apache::mod::auth_openidc inherits apache::params {
+# @param manage_dnf_module Whether to manage the DNF module
+# @param dnf_module_ensure The DNF module name to ensure. Only relevant if manage_dnf_module is set to true.
+# @param dnf_module_name The DNF module name to manage. Only relevant if manage_dnf_module is set to true.
+#
+# @see https://github.com/zmartzone/mod_auth_openidc for additional documentation.
+# @note Unsupported platforms: OracleLinux: 6; RedHat: 6; Scientific: 6; SLES: all
+#
+class apache::mod::auth_openidc (
+  Boolean $manage_dnf_module = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8',
+  String[1] $dnf_module_ensure = 'present',
+  String[1] $dnf_module_name = 'mod_auth_openidc',
+) {
   include apache
   include apache::mod::authn_core
   include apache::mod::authz_user
+
   apache::mod { 'auth_openidc': }
+
+  if $manage_dnf_module {
+    package { 'dnf-module-mod_auth_openidc':
+      ensure   => $dnf_module_ensure,
+      name     => $dnf_module_name,
+      provider => 'dnfmodule',
+      before   => Apache::Mod['auth_openidc'],
+    }
+  }
 }
--- a/manifests/mod/auth_basic.pp
+++ b/manifests/mod/auth_basic.pp
@@ -4,5 +4,6 @@
 # @see https://httpd.apache.org/docs/current/mod/mod_auth_basic.html for additional documentation.
 #
 class apache::mod::auth_basic {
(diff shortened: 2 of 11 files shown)


ekohl:
Odd thing: why is it running acceptance tests on Scientific 6 (which is EOL since 2020) but not on Red Hat 6?

ekohl:
I think this is now ready. It's starting to look greener and greener in the checks.

bastelfreak (approved):

manifests/mod/auth_openidc.pp
```diff
@@ -1,10 +1,29 @@
...
+# @param manage_dnf_module Whether to manage the DNF module
+# @param dnf_module_ensure The DNF module name to ensure. Only relevant if manage_dnf_module is set to true.
+# @param dnf_module_name The DNF module name to manage. Only relevant if manage_dnf_module is set to true.
+#
 # @see https://github.com/zmartzone/mod_auth_openidc for additional documentation.
+# @note Unsupported platforms: OracleLinux: 6; RedHat: 6; Scientific: 6; SLES: all
 #
-class apache::mod::auth_openidc inherits apache::params {
+class apache::mod::auth_openidc (
+  Boolean $manage_dnf_module = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8',
```
ekohl, line 12: Note that EL9 doesn't have this in a module, so it's limited to EL8.

evgeni, line 12: This is correct, because while EL9 has module support, `mod_auth_oidc` is not modular there.
(:exploding_head:)

evgeni (dismissed):

ekohl:
I just pushed an update that added a unit test to ensure the DNF module isn't present on non-EL8.

ekohl:
The RHEL 9 failure looks like an intermittent failure on provisioning the machine. It passed previously and Puppet 7 is green. Shouldn't be blocking.

ekohl:
Something @evgeni and I were discussing, does this module also need an include on `authn_core` so you can use `AuthType` if you disable default mods? `mod_auth_gssapi` does that at least:
https://github.com/puppetlabs/puppetlabs-apache/blob/92865b3d81dfd0895c60668b6f5db3e915a4e030/manifests/mod/auth_gssapi.pp#L8

@asieraguado you added the class initially, perhaps you can weigh in?

asieraguado:
> Something @evgeni and I were discussing, does this module also need an include on `authn_core` so you can use `AuthType` if you disable default mods? `mod_auth_gssapi` does that at least:
> 
> https://github.com/puppetlabs/puppetlabs-apache/blob/92865b3d81dfd0895c60668b6f5db3e915a4e030/manifests/mod/auth_gssapi.pp#L8
> 
> @asieraguado you added the class initially, perhaps you can weigh in?

Sorry, I didn't check if the module works without `authn_core`, but probably it's better to have the include. Some other auth modules don't have it, for example [auth_mellon](https://github.com/puppetlabs/puppetlabs-apache/blob/main/manifests/mod/auth_mellon.pp). Maybe it is needed in all of them?

evgeni:
Yeah, I think it's needed for all of them, as without `authn_core` there is no `AuthType` directive in the config, so even if the specific auth module itself is loadable (which I think it is), it can't be selected to be used for authentication.

ekohl:
> I think it's needed for all of them

Are we talkin about `manifests/mod/auth_*.pp` now?

evgeni:
> > I think it's needed for all of them
> 
> Are we talkin about `manifests/mod/auth_*.pp` now?

Technically `manifests/mod/auth_*.pp`,  `manifests/mod/authn_*.pp`,  `manifests/mod/authnz_*.pp`?

ekohl:
@david22swan thoughts? Ideally we would also get a release out soon with this.

Also, you don't do stable branches, right? Ideally we'd get this into a 7.x release as well.

david22swan (approved):
LGTM

Changes look goo and I don't see anything that stands out as needing a fix so would be happy to merge.

ekohl:
> Technically `manifests/mod/auth_*.pp`, `manifests/mod/authn_*.pp`, `manifests/mod/authnz_*.pp`?

I'd prefer to keep that for a future PR.

Can we merge this then? Ideally also get a release out (with the other fixes for which PRs are open now).

evgeni:
> > Technically `manifests/mod/auth_*.pp`, `manifests/mod/authn_*.pp`, `manifests/mod/authnz_*.pp`?
> 
> I'd prefer to keep that for a future PR.

Sure, that's mostly unrelated to the changes in this PR, just something that I started thinking of *because* of this PR ;)

david22swan:
Ok, everyone sounds happy so gonna go ahead and merge

PR #2288, opened 2022-08-10
Use more mod classes in default_mods

ekohl:
The goal is to remove duplication and make it easy to use mod classes where needed. This makes using the module without default mods easier. It also removes duplication.

This is a draft because apache::mod::expires creates a config by default, which is probably not what we want out of the box. I'd like to see a discusion on how to best resolve this.

### DIFF
--- a/manifests/default_mods.pp
+++ b/manifests/default_mods.pp
@@ -124,9 +124,6 @@ class apache::default_mods (
     include apache::mod::setenvif
     include apache::mod::auth_basic
 
-    # filter is needed by mod_deflate
-    include apache::mod::filter
-
     # authz_core is needed for 'Require' directive
     ::apache::mod { 'authz_core':
       id => 'authz_core_module',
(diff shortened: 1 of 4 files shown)


ekohl:
> This is a draft because apache::mod::expires creates a config by default, which is probably not what we want out of the box. I'd like to see a discusion on how to best resolve this.

@david22swan any thoughts on this?

ekohl:
I can't ping @puppetlabs/modules directly, but I just ran into this again.

ekohl:
@david22swan this isn't ready to merge, but here I could use some opinion. Especially if we're going to do a major release that may be a good time to change the default.

david22swan:
@ekohl Apologies for the wait on reply, from a general viewpoint this looks good to me to go forward though unsure, of the top of my head, how to resolve the issue with expires

ekohl:
It would be a breaking change to change it. I had hoped we could for 9.0.0. Shall I rebase it for now without the expires change and add a comment inline why we don't?

david22swan:
@ekohl Yeh, that sounds good.
Could also open a backwards incompatible issue to highlight it.
Make sure it's not forgotten when the next major release comes.

LukasAud:
Well, this is unfortunate. It looks like this PR has gone under our radar while working on our latest major release. We will need to revisit this PR again and find a suitable time to merge it before the next major release.

amitkarsale:
Hello @ekohl , I just happen to browse through this PR. Is this PR in progress or do you want to move it in review or you want it to be closed.

ekohl:
@amitkarsale I don't intend to work on this anymore. While I still think it's good, I just don't have the time for it. Also with Perforce making Puppet closed source I no more interest to contribute back.

Issue #2285, opened 2022-08-09
Apache:Balancermember configured data type no longer allows previously valid settings


canth1:
## Describe the Bug
Before v8.0.0, the Apache module would allow me to configure an apache::balancermember with URLs other than http://. For example the URL fcgi:// or unix:/ are both valid previously. However with commit https://github.com/puppetlabs/puppetlabs-apache/commit/f41251e336fa3e7c31c19968ccee74121cf71e47 the data type changed from string to Stdlib::HTTPUrl, which doesn't consider the previously mentioned URLs as valid.

## Expected Behavior
URLs such as fcgi:// and/or unix:/ are also valid, and should be allowed.

## Steps to Reproduce
1. Install v8.0.0 of the apache module
2. Configure a balancermember with the following settings:

apache::balancermember { 'FCGI':
	balancer_cluster	=>	'fcgi',
	url				=>	'fcgi://127.0.0.1:9001/',
}

or

apache::balancermember { 'UNIX':
	balancer_cluster	=>	'unix',
	url				=>	"unix:/run/some_socket.sock|fcgi://run/some_socket.sock",
}

## Environment
 - Version 8.0.0
 - Platform Ubuntu 20.04

ekohl:
Could you come up with a patch to change it to `String[1]` instead? Unless you can come up with a better pattern of course.

canth1:
> Could you come up with a patch to change it to `String[1]` instead? Unless you can come up with a better pattern of course.

Certainly, I'll see if I can come up with a good selection of matches, and looking at the docs, there will be several:

https://httpd.apache.org/docs/trunk/mod/mod_proxy.html

AJP13 (Apache JServe Protocol version 1.3) - ajp://
CONNECT (for SSL) - https://
FastCGI - fcgi://
ftp - ftp://
HTTP/0.9, HTTP/1.0, and HTTP/1.1 - http:// or https://
HTTP/2.0 - h2:// or h2c://
SCGI - scgi://
UWSGI - uwsgi://
WS and WSS (Web-sockets) - ws:// or wss://

And of course unix:/

I'm thinking the Pattern data type would be a good fit for this?

PR #2289, opened 2022-08-10
Add support for all proxy schemes, not just https://

canth1:
This commit fixes https://github.com/puppetlabs/puppetlabs-apache/issues/2285 by supporting all mod_proxy schemes according to the documentation here:

https://httpd.apache.org/docs/2.4/mod/mod_proxy.html


ekohl:
Looks like the tests fail on this.

I also wonder if this should be a defined type.

david22swan:
> Looks like the tests fail on this.
> 
> I also wonder if this should be a type alias.

I would agree on this

david22swan (changes requested):
@canth1 
Hey, look's like you've got a few failing tests

Also, think that @ekohl suggestion is a good one and that this would be a good use of a type alias to keep the code neater

types/modproxy.pp
```diff
@@ -0,0 +1,3 @@
+# @summary Supported protocols / schemes by mod_proxy
+# @see https://httpd.apache.org/docs/2.4/mod/mod_proxy.html
+type Apache::ModProxyProtocol = Pattern[/(\A(ajp|fcgi|ftp|h2|h2c|http|https|scgi|uwsgi|ws|wss):\/\/.*\z)/, /(\Aunix:\/([^\n\/\0]+\/*)*\z)/]
```
ekohl, line 3: Is it valid to not specify anything or should it be
```suggestion
type Apache::ModProxyProtocol = Pattern[/(\A(ajp|fcgi|ftp|h2|h2c|http|https|scgi|uwsgi|ws|wss):\/\/.+\z)/, /(\Aunix:\/([^\n\/\0]+\/*)*\z)/]
```

ekohl (approved):
Small detail, but this looks good.

ekohl (changes requested):
Looks like the tests are failing.

Also, it's possible to add tests for a type alias: https://rspec-puppet.com/documentation/type_aliases/

canth1:
> Small detail, but this looks good.

I don't think that's correct, because I took my regex directly from Stdlib:HTTPSUrl

https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/types/httpsurl.pp

Which is:

type Stdlib::HTTPSUrl = Pattern[/(?i:\Ahttps:\/\/.*\z)/]

canth1:
> Looks like the tests are failing.
> 
> Also, it's possible to add tests for a type alias: https://rspec-puppet.com/documentation/type_aliases/

Ok, I'll add the tests for this new type alias.

ekohl:
> > Small detail, but this looks good.
> 
> I don't think that's correct, because I took my regex directly from Stdlib:HTTPSUrl
> 
> https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/types/httpsurl.pp
> 
> Which is:
> 
> type Stdlib::HTTPSUrl = Pattern[/(?i:\Ahttps://.*\z)/]

Interesting. That implies it also accepts `https://` which doesn't make sense to me since I don't consider it an HTTP(S) URL.

types/modproxy.pp
```diff
@@ -0,0 +1,3 @@
+# @summary Supported protocols / schemes by mod_proxy
+# @see https://httpd.apache.org/docs/2.4/mod/mod_proxy.html
+type Apache::ModProxyProtocol = Pattern[/(\A(ajp|fcgi|ftp|h2c?|https?|scgi|uwsgi|wss?):\/\/.*\z)/, /(\Aunix:\/([^\n\/\0]+\/*)*\z)/]
```
ekohl, line 3: I don't think this is correct. The filename must match the type name.

canth1:
> > > Small detail, but this looks good.
> > 
> > 
> > I don't think that's correct, because I took my regex directly from Stdlib:HTTPSUrl
> > https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/types/httpsurl.pp
> > Which is:
> > type Stdlib::HTTPSUrl = Pattern[/(?i:\Ahttps://.*\z)/]
> 
> Interesting. That implies it also accepts `https://` which doesn't make sense to me since I don't consider it an HTTP(S) URL.

Actually, you're correct. We shouldn't allow an empty value after ://

 I'll fix that with your suggestion.

It looks like the Stdlib:HTTPSUrl allows that for some reason?

ekohl:
> It looks like the Stdlib:HTTPSUrl allows that for some reason?

Possibly an oversight and nobody ran into problems because of it.

canth1:
The spec tests are failing, but I can't figure out why?

Resource type not found: Apache::ModProxyProtocol

Not sure why it's giving me that error?

ekohl:
Because of https://github.com/puppetlabs/puppetlabs-apache/pull/2289#discussion_r948196017. You must name it according to the conventions so the loader finds it, so `types/mod_proxy_protocol.pp`.

ekohl (approved):
This looks good to me. @david22swan you requested changes, mind taking another look? Also, IMHO we should squash this on merge.

david22swan (approved):
LGTM

Everything look's good to me, only small comment is that the variables aren't aligned, but since there are only two of them it's not that big of a deal in my eyes.
Will go ahead with a squash merge.

Issue #2286, opened 2022-08-10
timeout param issue in mod::proxy


robertc99:
Version 8 has set the type of apache::mod::proxy::proxy_timeout as undef or string. It should be undef or integer.
proxy_iobuffersize should probably also be integer.

ekohl:
So this is:
https://github.com/puppetlabs/puppetlabs-apache/blob/92865b3d81dfd0895c60668b6f5db3e915a4e030/manifests/mod/proxy.pp#L33
The docs are https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxytimeout and those say it is in seconds, so `Integer` makes sense. Probably even `Integer[0]` since negative values are unsupported AFAIK.

Could you come up with a PR?

ekohl:
Fixed in #2294

PR #2290, opened 2022-08-11
change type of mod::proxy::proxytimeout to integer

robertc99:
as per https://github.com/puppetlabs/puppetlabs-apache/issues/2286

### DIFF
--- a/manifests/mod/proxy.pp
+++ b/manifests/mod/proxy.pp
@@ -30,7 +30,7 @@ class apache::mod::proxy (
   Optional[String] $apache_version     = undef,
   Optional[String] $package_name       = undef,
   String $proxy_via                    = 'On',
-  Optional[String] $proxy_timeout      = undef,
+  Optional[Integer[0]] $proxy_timeout      = undef,
   Optional[String] $proxy_iobuffersize = undef,
 ) {
   include apache


manifests/mod/proxy.pp
```diff
@@ -30,7 +30,7 @@
   Optional[String] $apache_version     = undef,
   Optional[String] $package_name       = undef,
   String $proxy_via                    = 'On',
-  Optional[String] $proxy_timeout      = undef,
+  Optional[Integer[0]] $proxy_timeout      = undef,
```
ekohl, line 33: Nit: ideally this would be aligned witht he other parameters

ekohl (commented):
Technically this would be a breaking change. I wonder if we should go the compatibility road and change it to `Variant[Undef, Integer[0], String]`.

@david22swan thoughts?

manifests/mod/proxy.pp
david22swan: @robertc99 Think it would be best to do as @ekohl suggested and make this change backwards compatible so we don't have to immediately cut a major release.
Have created a ticket to track this and any other similar issues so that we can correct them when the next major release is made:
https://github.com/puppetlabs/puppetlabs-apache/issues/2291

david22swan:
@robertc99 Think it would be best to do as @ekohl suggested and make this change backwards compatible so we don't have to immediately cut a major release.
Have created a ticket to track this and any other similar issues so that we can correct them when the next major release is made:
https://github.com/puppetlabs/puppetlabs-apache/issues/2291

traylenator:
Should be closed now that https://github.com/puppetlabs/puppetlabs-apache/pull/2294 is merged.

ekohl:
Thanks everyone involved, but #2294 was indeed merged and also tacked another class.

PR #2292, opened 2022-08-11
Allow setting icons_path to false so no alias will be set for it

Zarne:
with the commit f41251e336fa3e7c31c19968ccee74121cf71e47, icons_path set to false does not work.
This commit fixes that.


ekohl (approved):
Thanks. I took the liberty of shortening the summary a bit.

The OracleLinux test failure looks unrelated, others are still running so not merging yet.

david22swan (approved):
LGTM

PR #2293, opened 2022-08-11
Fix deprecation warning about performing a regex comparison on a hash

smokris:
I'm getting warnings like this for each `puppet agent` run:
```
templates/vhost/_require.erb:3: warning: deprecated Object#=~ is called on Hash; it always returns nil
```
This PR changes the logic to only performing the regex comparison when [the `requires` value](https://github.com/puppetlabs/puppetlabs-apache/blob/main/REFERENCE.md#requires) is a string.

### DIFF
--- a/templates/vhost/_require.erb
+++ b/templates/vhost/_require.erb
@@ -1,6 +1,6 @@
 <%- _item = scope.lookupvar('_template_scope')[:item] -%>
 <%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
-  <%- if _item['require'] && _item['require'] != '' && _item['require'] !~ /unmanaged/i -%>
+  <%- if _item['require'] && !(_item['require'].is_a?(String) && (_item['require'] == '' || _item['require'] =~ /unmanaged/i)) -%>
     <%- if _item['require'].is_a?(Hash) -%>
       <%- case _item['require']['enforce'].downcase
           when 'all','none','any' then -%>


ekohl (commented):
Don't you have the same problem on line 24?

smokris:
> Don't you have the same problem on line 24?

Hmm, good question.  Here's line 24:
```erb
<%- if !(_item['require'] && _item['require'] != '') && _item['require'] !~ /unmanaged/i && !(_item['auth_require']) -%>
```

When `_item['require']` is a hash, `!(_item['require'] && _item['require'] != '')` evaluates to false, so the other clauses are not evaluated, and thus it doesn't emit the "deprecated Object#=~ is called on Hash" warning.

david22swan:
@ekohl Are you good with the above explanation?? ^^^

ekohl (approved):
I think you're right and the whole `_item['require'] !~ /unmanaged/i` is redundant. It's complex.

If I'm reading it right, this is equal:
```
<%- if (!_item['require'] || _item['require'] == '') && _item['require'] !~ /unmanaged/i && !(_item['auth_require']) -%>
```

Now if you read that form, it must be either `undef` or `''`. Both of those are never `/unmanaged/i` so it's redundant, thus equal to
```
<%- if (!_item['require'] || _item['require'] == '') && !(_item['auth_require']) -%>
```

Once you read that, it can actually become an `elsif` for the block above so it becomes:
```erb
  <%- if _item['auth_require'] -%>
    Require <%= _item['auth_require'] %>
  <%- elsif !_item['require'] || _item['require'] == '' -%>
    Require all granted
  <%- end -%>
```

If my logic there is correct, I think we don't need to modify it now.

ekohl:
> If my logic there is correct, I think we don't need to modify it now.

Just to clarify: if `_require` is `nil` then it does `nil !~ /unmanaged/i`, but that doesn't raise any deprecation warning.

Submitted my suggestion as a separate PR: https://github.com/puppetlabs/puppetlabs-apache/pull/2303

david22swan (approved):

david22swan:
Ok, everything looks good to merge then
Thanks for putting in the work

PR #2301, opened 2022-09-06
Allow RewriteInherit with empty rewrites

martin-koerner:
There are cases where you want to exclusively inherit RewriteRules from e.g. global scope without specifying vHost-specific rewrites - which is now allowed again.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2514,11 +2514,12 @@ define apache::vhost (
 
   # Template uses:
   # - $rewrites
+  # - $rewrite_inherit
   # - $rewrite_base
   # - $rewrite_rule
   # - $rewrite_cond
   # - $rewrite_map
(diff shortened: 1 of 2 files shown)


ekohl (commented):
Could you add a test case for this?

martin-koerner:
> Could you add a test case for this?

Sure, I updated the existing one for empty rewrites with rewrite_inherit = false and added a second one - hope that's sufficient:
```
        empty rewrites
          is expected not to contain Concat::Fragment[rspec.example.com-rewrite]
        empty rewrites_with_rewrite_inherit
          is expected to contain Concat::Fragment[rspec.example.com-rewrite] with content =~ /^\s+RewriteOptions Inherit$/
```

david22swan (approved):
LGTM

david22swan:
@martin-koerner This change looks good to me and the test coverage seems sufficient. Thanks for putting in the work :)

PR #2307, opened 2022-09-07
adding boolean and variant

ajm-ex-dev:


david22swan:
@ajm-ex-dev Ah, looking at the test the problem shows quickly, look's like `Boolean[False]` can't be a type unfortunately.
Sorry for suggesting it, not sure why I thought it was a thing.
Apologies, but if you could swap it to just `Boolean` then this should be good to merge

david22swan:
@ajm-ex-dev Hey, have a pr up with the changes suggested in the thread if you could try it out?
https://github.com/puppetlabs/puppetlabs-apache/pull/2309

Issue #2291, opened 2022-08-11
Small changes for next major release


david22swan:
A list of small backwards incompatible changes that should be included when we next do a major release, any changes listed here are from bugs that have been given compatible fixes for now:

- `$proxy_timeout` in manifest/mod/proxy should have the type `Optional[Integer[0]]`
https://github.com/puppetlabs/puppetlabs-apache/blob/main/manifests/mod/proxy.pp#L33, also `cas_timeout` and `cas_idle_timeout`

traylenator:
Some more.

* `apache::vhost::request_headers` was supported as string as well as an array however only documented as an array - so could be ignored.

david22swan:
Linking: https://github.com/puppetlabs/puppetlabs-apache/pull/2350

PR #2309, opened 2022-09-09
(maint) Add variable manage_vhost_enable_dir

david22swan:
Additional variable added at request of community to allow for `vhost_enable_dir` to be disabled on Debian systems.
Due to the way in which aid variable inherits from the params manifest, simply setting it to `undef` does not work as when this is done it's logic assumes it should inherit from the default params value.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -439,8 +439,11 @@
 #   Whether the additional config files in `/etc/apache2/conf-enabled` should be managed.
 # 
 # @param vhost_enable_dir
-#   Set's whether the vhost definitions will be stored in sites-availible and if
+#   Set's the vhost definitions which will be stored in sites-availible and if
 #   they will be symlinked to and from sites-enabled.
+# 
+# @param manage_vhost_enable_dir
+#   Overides the vhost_enable_dir inherited parameters and allows it to be disabled
 #
 # @param mod_enable_dir
 #   Set's whether the mods-enabled directory should be managed.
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -500,6 +503,7 @@
   Optional[Stdlib::Absolutepath] $conf_enabled                               = $apache::params::conf_enabled,
   Stdlib::Absolutepath $vhost_dir                                            = $apache::params::vhost_dir,
   Optional[Stdlib::Absolutepath] $vhost_enable_dir                           = $apache::params::vhost_enable_dir,
```
chelnak, line 505: Do you think we could change this to `vhost_dir` to make the parameters less ambiguous? 

I'm also assuming that `vhost_enable_dir` in this instance is referring to the document root in the apache config?

david22swan: That would make this a backwards incompatible change and require a major release done.
We could add it as something to be done for the next major release though in the list pr:
https://github.com/puppetlabs/puppetlabs-apache/issues/2291

david22swan: Also, there already is a vhost_dir

chelnak: I think it would certainly need cleaning up yeah.

To me, `vhost_enable_dir` and `enable_vhost_enable_dir` is very confusing.

What is stopping us from cutting a major release?

chelnak (dismissed):

LukasAud (approved):

PR #2312, opened 2022-09-13
fix directory empty options if an empty array is being used

bovy89:
puppetlabs-apache 6.5.1 with `options => false`

```
  <Directory "/var/www/html/myapp">
    AllowOverride None
    Require all granted
  </Directory>
```

puppetlabs-apache latest with `options => []`

```
   <Directory "/var/www/html/myapp">
     Options
     AllowOverride None
     Require all granted
   </Directory>
```

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -37,7 +37,7 @@
     <%- if ! directory['geoip_enable'].nil? -%>
     GeoIPEnable <%= scope.call_function('apache::bool2httpd', [directory['geoip_enable']]) %>
     <%- end -%>
-    <%- if directory['options'] -%>
+    <%- if directory['options'] && Array(directory['options']).any? -%>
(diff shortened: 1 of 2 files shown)


templates/vhost/_directories.erb
```diff
@@ -37,7 +37,7 @@
     <%- if ! directory['geoip_enable'].nil? -%>
     GeoIPEnable <%= scope.call_function('apache::bool2httpd', [directory['geoip_enable']]) %>
     <%- end -%>
-    <%- if directory['options'] -%>
+    <%- if Array(directory['options']).length > 0 -%>
```
ekohl, line 40: I think you need to check if the option is set at all to be safe. Then it's also better to use `any?`
```suggestion
    <%- if directory['options'] && Array(directory['options']).any? -%>
```

ekohl (approved):

david22swan (approved):
LGTM

PR #2313, opened 2022-09-13
(maint) Codebase Hardening

david22swan:
Changes made to ensure that no malformed commands are passed through to the system.

Includes an update the the $verify_command variable type
The unless variable in an exec manifest, which $verify_command is directly passed too, accepts either an Array or and Array of Arrays as input.
Updating the variables type and default values to match this.
In the case where it is passed a String it proceeds to wrap it in an Array and then treat it as such.

### DIFF
--- a/manifests/custom_config.pp
+++ b/manifests/custom_config.pp
@@ -32,6 +32,9 @@
 #   This parameter is used only if the `verify_config` parameter's value is `true`. If the 
 #   `verify_command` fails, the Puppet run deletes the configuration file and raises an error, 
 #   but does not notify the Apache service.
+#   Command can be passed through as either a String, i.e. `'/usr/sbin/apache2ctl -t'`
+#   An array, i.e. `['/usr/sbin/apache2ctl', '-t']`
+#   Or an array of arrays with each one having to pass succesfully, i.e. `[['/usr/sbin/apache2ctl', '-t'], '/usr/sbin/apache2ctl -t']`
 #
 # @param verify_config
 #   Specifies whether to validate the configuration file before notifying the Apache service.
@@ -49,18 +52,18 @@
 #   show_diff property for configuration file resource
 #
 define apache::custom_config (
-  Enum['absent', 'present'] $ensure     = 'present',
-  Stdlib::Absolutepath $confdir         = $apache::confd_dir,
-  Optional[String] $content             = undef,
-  Apache::Vhost::Priority $priority     = 25,
-  Optional[String] $source              = undef,
-  String $verify_command                = $apache::params::verify_command,
-  Boolean $verify_config                = true,
-  Optional[String] $filename            = undef,
-  Optional[String] $owner               = undef,
-  Optional[String] $group               = undef,
-  Optional[Stdlib::Filemode] $file_mode = undef,
-  Boolean $show_diff                    = true,
+  Enum['absent', 'present'] $ensure                                    = 'present',
+  Stdlib::Absolutepath $confdir                                        = $apache::confd_dir,
+  Optional[String] $content                                            = undef,
+  Apache::Vhost::Priority $priority                                    = 25,
+  Optional[String] $source                                             = undef,
+  Variant[String, Array[String], Array[Array[String]]] $verify_command = $apache::params::verify_command,
+  Boolean $verify_config                                               = true,
+  Optional[String] $filename                                           = undef,
+  Optional[String] $owner                                              = undef,
+  Optional[String] $group                                              = undef,
+  Optional[Stdlib::Filemode] $file_mode                                = undef,
+  Boolean $show_diff                                                   = true,
 ) {
   if $content and $source {
     fail('Only one of $content and $source can be specified.')
--- a/manifests/custom_config.pp
+++ b/manifests/custom_config.pp
@@ -90,11 +90,12 @@ define apache::custom_config (
     $notifies = undef
   }
 
+  $_file_path = "${confdir}/${_filename}"
   $_file_mode = pick($file_mode, $apache::file_mode)
 
   file { "apache_${name}":
     ensure    => $ensure,
-    path      => "${confdir}/${_filename}",
+    path      => $_file_path,
     owner     => $owner,
     group     => $group,
     mode      => $_file_mode,
@@ -115,7 +116,7 @@ define apache::custom_config (
       require     => Anchor['::apache::modules_set_up'],
     }
 
-    $remove_command = ['/bin/rm', shell_escape(join([$confdir, $_filename], '/'))]
+    $remove_command = ['/bin/rm', $_file_path]
     exec { "remove ${name} if invalid":
       command     => $remove_command,
       unless      => [$verify_command],
--- a/manifests/custom_config.pp
+++ b/manifests/custom_config.pp
@@ -115,9 +115,10 @@ define apache::custom_config (
       require     => Anchor['::apache::modules_set_up'],
     }
 
+    $remove_command = ['/bin/rm', shell_escape(join([$confdir, $_filename], '/'))]
     exec { "remove ${name} if invalid":
-      command     => "/bin/rm ${confdir}/${_filename}",
-      unless      => $verify_command,
+      command     => $remove_command,
+      unless      => [$verify_command],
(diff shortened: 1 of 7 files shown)


manifests/custom_config.pp
```diff
@@ -115,9 +115,10 @@
       require     => Anchor['::apache::modules_set_up'],
     }
 
+    $remove_command = ['/bin/rm', shell_escape(join([$confdir, $_filename], '/'))]
```
ekohl, line 118: `shell_escape` shouldn't be needed when passing an array
```suggestion
    $remove_command = ['/bin/rm', join([$confdir, $_filename], '/')]
```

manifests/init.pp
```diff
@@ -635,7 +635,9 @@
     path => '/bin:/sbin:/usr/bin:/usr/sbin',
   }
 
+  $confd_command = ['mkdir', $confd_dir]
   exec { "mkdir ${confd_dir}":
+    command => $confd_command,
```
ekohl, line 640: Do you need to define an additional variable for this? I'd think this is the same
```suggestion
  exec { "mkdir ${confd_dir}":
    command =>  ['mkdir', $confd_dir],
```

manifests/mpm/disable_mpm_event.pp
```diff
@@ -1,19 +1,28 @@
 # @summary disable Apache-Module event
 class apache::mpm::disable_mpm_event {
+  $event_command = ['/usr/sbin/a2dismod', 'event']
+  $event_onlyif = ['/usr/bin/test', '-e', shell_escape(join([$apache::mod_enable_dir, 'event.load'],'/'))]
```
ekohl, line 4: `onlyif` should be a nested array since it can take an array of commands.

ekohl (changes requested):
Comments apply multiple times

manifests/custom_config.pp
david22swan: My thought was that since I'm joining it and passing it through that way it may need to be escaped where it wouldn't if I was just passing it through.

manifests/init.pp
david22swan: It's mainly a stylistic choice to keep everything consistent.
While this instance and a few others would read fine included like you suggested there are others that read significantly better separated out and so I have them all separated out for consistency.

manifests/mpm/disable_mpm_event.pp
david22swan: changed

manifests/custom_config.pp
david22swan: This does seem wrong now......
Have updated the code

manifests/custom_config.pp
```diff
@@ -115,9 +115,10 @@
       require     => Anchor['::apache::modules_set_up'],
     }
 
+    $remove_command = ['/bin/rm', join([$confdir, $_filename], '/')]
```
ekohl, line 118: Given that the same concat happens on line 97, should a dedicated variable be introduced for the filename?

And a whole different thought: `file` has a `validate_cmd` option. However, I don't think it can be used in this case since you can't validate the config file by itself. It needs the whole apache config. That's probably why this rollback mechanism was invented here. It does make you wonder if the `file` type should have a `validate_mode` where it can work in a mode where it replaces the real file rather than a temporary file and rolls back if needed.

david22swan: To the first point, it does look better and removes the chance of them getting out of sync, so have implemented.

As to the second.... it sounds like a reasonable idea to me.
Could be something to raise.

manifests/custom_config.pp
```diff
@@ -90,11 +90,12 @@
     $notifies = undef
   }
 
+  $_file_path = join([$confdir, $_filename], '/')
```
ekohl, line 93: I do wonder what the value of join here is over simple string interpolation. It's not like you win anything in compatibility (unlike OS specific join for Windows/Unix) and IMHO readability is also worse.

manifests/custom_config.pp
```diff
@@ -115,9 +116,10 @@
       require     => Anchor['::apache::modules_set_up'],
     }
 
+    $remove_command = ['/bin/rm', $_file_path, '/')]
```
ekohl, line 119:
```suggestion
    $remove_command = ['/bin/rm', $_file_path]
```

manifests/custom_config.pp
david22swan: Had used the join as I had simply copied from he remove_command array, but it does read easier as a string

chelnak (changes requested):
Looking good! Just one question regarding `$verify_command`.

ekohl:
Just a thought: can we develop a puppet-lint plugin that detects string interpolation in `exec`'s command/unless attributes? It will probably be easy to fool (doing string interpolation and store that in a variable will be much harder to detect), but simple cases could be caught.

chelnak:
> Just a thought: can we develop a puppet-lint plugin that detects string interpolation in `exec`'s command/unless attributes? It will probably be easy to fool (doing string interpolation and store that in a variable will be much harder to detect), but simple cases could be caught.

Yes! This is actually something that we were planning when we had our team meet up last week. Should be a nice piece of work for someone on the team to run with.

manifests/custom_config.pp
```diff
@@ -115,9 +116,10 @@
       require     => Anchor['::apache::modules_set_up'],
     }
 
+    $remove_command = ['/bin/rm', $_file_path]
     exec { "remove ${name} if invalid":
-      command     => "/bin/rm ${confdir}/${_filename}",
-      unless      => $verify_command,
+      command     => $remove_command,
+      unless      => [$verify_command],
```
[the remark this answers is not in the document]
david22swan: Bit iffy on doing this.
The verify command can be manually set and we know that splitting the command can cause errors in certain situations, i.e. the use of pipes
Not sure I like the idea of splitting the command when I don't know what it may be.
Probably other thinking it, but still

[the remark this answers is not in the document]
chelnak: Yes that is a good point.. but to me bolsters the fact that we should be guiding people to properly structure their commands.

`unless` takes an array of arrays. Each inner array is interpreted as a command by the provider. When multiple commands are passed each one needs to evaluate as true for the overall success of `unless.

Though the above would be a breaking change for some which adds to the complication.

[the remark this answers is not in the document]
david22swan: Could split the default command into an array. Then set verify_command to be: `Variant[String, Array[String], Array[Array[String]]`
It's a bit messy, but is technically what it should accept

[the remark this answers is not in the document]
david22swan: Then maybe remove the string in the next major release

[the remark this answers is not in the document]
chelnak: Yeah I think that is a nice compromise actually given that the exec type defines `unless` like [this](https://github.com/puppetlabs/puppet/blob/04cf2ae3ac98d7ac628663f4ef0b652b6b013041/lib/puppet/type/exec.rb#L451).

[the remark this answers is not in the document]
david22swan: pushed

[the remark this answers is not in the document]
ekohl: > Could split the default command into an array. Then set verify_command to be: Variant[String, Array[String], Array[Array[String]]
> It's a bit messy, but is technically what it should accept

I like this, but would it make sense to (also) introduce a `Stdlib::Exec::Unless` data type? While we couldn't use it immediately, it does signal to the user how it should be used. Just like there's `Stdlib::Ensure::*`.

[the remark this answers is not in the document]
david22swan: also rebased onto main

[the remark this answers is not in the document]
david22swan: and squashed down all my corrective changes into one commit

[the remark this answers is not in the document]
chelnak: @ekohl that would be nice yes. I'm a fan of anything we can do to help users on their journey.

chelnak (approved):

chelnak (dismissed):
Looking great @david22swan 

Your branch could do with a rebase before merge though please 👍

chelnak (approved):

PR #2314, opened 2022-09-14
(MAINT) Removing support for CentOS 8

jordanbreenpuppet:
Prior to this commit the metadata.json file for this module listed OSs that are not supported by puppet agent as supported. This commit aims to refactor the metadata.json file to only list OSs that are supported by puppet agent.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -30,8 +30,7 @@
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
-        "7",
-        "8"
+        "7"
       ]
     },
     {


metadata.json
```diff
@@ -66,7 +65,6 @@
     {
       "operatingsystem": "Ubuntu",
       "operatingsystemrelease": [
-        "18.04",
```
ekohl, line 69: This surprises me. Ubuntu 18.04 is not yet EOL.

GSPatton (approved):
Looks good

PR #2327, opened 2022-10-09
Update vhost_proxypass.pp

akondasif:
HTTPS instead of HTTP


ekohl:
Thank you for your submission, but I don't think this adds anything. Very commonly port 8080 is HTTP and it's on localhost. It doesn't really enhance the example.

Issue #2318, opened 2022-09-20
@param rewrite_rule ignored in apache::vhost due to changed default value of @param rewrite


janit42:
## Describe the Bug
a rewrite rule for the `apache::vhost` type defined in `$rewrite_rule` is ignored since [7b22dae](https://github.com/puppetlabs/puppetlabs-apache/commit/7b22dae24ff4a04f9258362f90bf6a2151c24a23) and not written to the vhost file if `$rewrites` is not explicitly set to `undef`.
 
## Expected Behavior
rewrite rules defined in $rewrite_rule of the apache::vhost type should be written to the vhost file with the default value of `$rewrites`

## Additional Context
In https://github.com/puppetlabs/puppetlabs-apache/commit/7b22dae24ff4a04f9258362f90bf6a2151c24a23 the default value for `$rewrites` was changed from `undef` to `[]` in [manifests/vhost.pp](https://github.com/puppetlabs/puppetlabs-apache/blob/main/manifests/vhost.pp#L1819). This results in [templates_vhost/_rewrite.erb](https://github.com/puppetlabs/puppetlabs-apache/blob/main/templates/vhost/_rewrite.erb#L41) always evaluating to false (boolean check on an empty array variable is still `true`) unless `$rewrites` is explicitly set to `undef`. This results in the content of `$rewrite_rule` being ignored.

I'm not sure if this is the intended behavior.

A possible solution would be to change the [default for `$rewrites`](https://github.com/puppetlabs/puppetlabs-apache/blob/main/manifests/vhost.pp#L1819) back to `undef` or to extend the checks in [/templates/vhost/_rewrite.erb#L1](https://github.com/puppetlabs/puppetlabs-apache/blob/main/templates/vhost/_rewrite.erb#L1) and [/templates/vhost/_rewrite.erb#L41](https://github.com/puppetlabs/puppetlabs-apache/blob/main/templates/vhost/_rewrite.erb#L41) to also test for an empty array.

PR #2330, opened 2022-10-12
fix rewrite rules being ignored

trefzer:
In 7b22dae the default value for $rewrites was changed from undef to [] in manifests/vhost.pp. This results in templates_vhost/_rewrite.erb always evaluating to false (boolean check on an empty array variable is still true) unless $rewrites is explicitly set to undef. This results in the content of $rewrite_rule being ignored.

closes: #2318

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -738,7 +738,6 @@ describe 'apache::vhost', type: :define do
               .with_content(%r{^\s+RewriteOptions Inherit$})
               .with_content(%r{^\s+RewriteBase /})
               .with_content(%r{^\s+RewriteRule ^index\.html$ welcome.html$})
-            )
           }
           it { is_expected.to contain_concat__fragment('rspec.example.com-scriptalias') }
           it { is_expected.to contain_concat__fragment('rspec.example.com-serveralias') }
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -379,7 +379,7 @@ describe 'apache::vhost', type: :define do
               'request_headers'             => ['append MirrorID "mirror 12"'],
               'rewrites'                    => [
                 {
-                  'rewrite_rule' => ['^index\.html$ rewrites.html'],
+                  'rewrite_rule' => ['^index.html$ rewrites.html'],
                 },
               ],
               'filters' => [
@@ -392,7 +392,7 @@ describe 'apache::vhost', type: :define do
                 'FilterProtocol COMPRESS  DEFLATE change=yes;byteranges=no',
               ],
               'rewrite_base'                => '/',
-              'rewrite_rule'                => '^index\.html$ welcome.html',
+              'rewrite_rule'                => '^index.html$ welcome.html',
               'rewrite_cond'                => ['%{HTTP_USER_AGENT} ^MSIE'],
               'rewrite_inherit'             => true,
               'setenv'                      => ['FOO=/bin/true'],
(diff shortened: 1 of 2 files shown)


trefzer:
@ekohl
I agree, but since the logic is in a template we would need to check the final content in the file which I have no idea how to do with spec test. (anyone who nows please let me know !)

ekohl:
I would probably take this example:
https://github.com/puppetlabs/puppetlabs-apache/blob/77255d652212466ae5dbc81b1381936794c50dbf/spec/defines/vhost_spec.rb#L1723-L1736
Then copy it and pass some parameters that triggers the behavior you want to test.

Note you can also test the whole file:
```ruby
it {
  is_expected.to contain_concat__fragment('rspec.example.com-rewrite').with_content <<~CONTENT
    Here you can write the full content
    And use multiple lines
  CONTENT
}
```

spec/defines/vhost_spec.rb
```diff
@@ -741,7 +741,10 @@
           it { is_expected.to contain_concat__fragment('rspec.example.com-redirect') }
           it {
             is_expected.to contain_concat__fragment('rspec.example.com-rewrite').with(
+              content: %r{^\s+RewriteEngine On$},
```
ekohl, line 744: You can't test this way (as the action indicates), but you can chain it: `.with(content: ...).with(content: ...)`

david22swan:
@trefzer Hey sorry to bother you, just checking in on when you'll have time to finish with the sped test?

spec/defines/vhost_spec.rb
```diff
@@ -740,8 +740,11 @@
           }
           it { is_expected.to contain_concat__fragment('rspec.example.com-redirect') }
           it {
-            is_expected.to contain_concat__fragment('rspec.example.com-rewrite').with(
-              content: %r{^\s+RewriteOptions Inherit$},
+            is_expected.to contain_concat__fragment('rspec.example.com-rewrite')
+              .with_content(%r{^\s+RewriteEngine On$})
+              .with_content(%r{^\s+RewriteOptions Inherit$}),
+              .with_content(%r{^\s+RewriteBase /}),
+              .with_content(%r{^\s+RewriteRule ^index\.html$ welcome.html$}),
```
ekohl, line 747:
```suggestion
              .with_content(%r{^\s+RewriteOptions Inherit$})
              .with_content(%r{^\s+RewriteBase /})
              .with_content(%r{^\s+RewriteRule ^index\.html$ welcome.html$})
```

templates/vhost/_rewrite.erb
```diff
@@ -1,4 +1,4 @@
-<%- if @rewrites -%>
+<%- if @rewrite_inherit or not @rewrites.empty? -%>
```
ekohl, line 1:
```suggestion
<%- if @rewrite_inherit or !@rewrites.empty? -%>
```

ekohl (commented):
These changes should make the test suite work.

spec/defines/vhost_spec.rb
trefzer: ups :) missed them.

david22swan (approved):
LGTM

@ekohl Just checking you're good with the changes you requested?

If so will merge in and release

ekohl (approved):
Yes, I'm good. All my comments were there to make the test suite pass (we'll ignore mend for now) so :+1: to merging this.

Thanks @trefzer for your efforts! And it was great to collaborate on a PR in person again. That was a long time ago for me

david22swan:
Good good good

Should be released by end of day

PR #2331, opened 2022-10-12
add maxrequestworkers parameter for mpm_worker module

trefzer:


trefzer:
check seems to fail provisioning SLES-15 which ist not a problem of this commit.

ekohl (changes requested):
In other places (prefork/event mpm) we prefer `MaxRequestWorkers` and don't output `MaxClients` if it's set. Could you implement the same here?

trefzer:
check seems to fail provisioning RedHat 7 and 9, not the change request.

david22swan:
@ekohl Are you happy with the changes made?

spec/classes/mod/worker_spec.rb
```diff
@@ -126,7 +126,8 @@
       it { is_expected.to contain_file('/etc/httpd/conf.modules.d/worker.conf').with(content: %r{^<IfModule mpm_worker_module>$}) }
       it { is_expected.to contain_file('/etc/httpd/conf.modules.d/worker.conf').with(content: %r{^\s+ServerLimit\s+10$}) }
       it { is_expected.to contain_file('/etc/httpd/conf.modules.d/worker.conf').with(content: %r{^\s+StartServers\s+11$}) }
-      it { is_expected.to contain_file('/etc/httpd/conf.modules.d/worker.conf').with(content: %r{^\s+MaxClients\s+12$}) }
+      it { is_expected.not_to contain_file('/etc/httpd/conf.modules.d/worker.conf').with(content: %r{^\s+MaxClients}) }
```
ekohl, line 129: It's better to expect a file to be there without certain content. The big reason is that your example will also pass if the file isn't present in the catalog at all.
```suggestion
      it { is_expected.to contain_file('/etc/httpd/conf.modules.d/worker.conf').without(content: %r{MaxClients}) }
```

On a side note: this whole file should be rewritten IMHO to have a single `it` statement with one long `content` block that has the full generated template, but that's not part of this PR.

manifests/mod/worker.pp
```diff
@@ -8,6 +8,7 @@
 #   The max number of simultaneous requests that will be served.
 #   This is the old name and is still supported. The new name is
 #   MaxRequestWorkers as of 2.3.13.
+#   If maxreuestworkers is set, this value is ignored.
```
ekohl, line 11:
```suggestion
#   If maxrequestworkers is set, this value is ignored.
```

spec/classes/mod/worker_spec.rb
trefzer: Ack done

david22swan (approved):
Little concerned with maxclients no longer having a test where it is explicitly set, but given it has a default value covered in the other tests this should be fine.

LGTM

@ekohl Are you happy that the changes that you requested have been implemented?

manifests/mod/worker.pp
```diff
@@ -40,6 +41,10 @@
 # @param apache_version
 #   Used to verify that the Apache version you have requested is compatible with the module.
 #
+# @param maxrequestworkers
+#   Maximum number of connections that will be processed simultaneously
```
ekohl, line 45: I think this will show up slightly better in the rendered strings documentation
```suggestion
#   Maximum number of connections that will be processed simultaneously.
```

ekohl (approved):
One tiny comment, but generally :+1:

david22swan:
Gonna go ahead and merge.
@trefzer Thanks for putting in the work.
@ekohl Thanks for helping to review

PR #2333, opened 2022-10-12
Fix RedHat + PHP 8 libphp file

### LINKED ISSUE #2332: Comparison error in apache::mod::php when RedHat and PHP 8
## Describe the Bug
When `$php_version` is set to `8.0` and the OS is RedHat, the wrong php library is being loaded due to an incorrect comparison.

It should load `libphp.so`, but instead it tries to load `libphp8.0.so`.

The problem is in line 96:

https://github.com/puppetlabs/puppetlabs-apache/blob/6bc10727f32032133337ed5022ef5158b0358847/manifests/mod/php.pp#L92-L101

Line 92 returns a string (so `$_php_major` is a string), but in line 96 it is comparing it to an integer.

A fix could be replacing line 96 with (adding single quotes to `8`:
`} elsif ($facts['os']['family'] == 'RedHat') and ($_php_major == '8') { `
or even replace it with:
`} elsif ($facts['os']['family'] == 'RedHat') and (versioncmp($php_version, '8') >= 0) {`
like it is done on line 47:
https://github.com/puppetlabs/puppetlabs-apache/blob/6bc10727f32032133337ed5022ef5158b0358847/manifests/mod/php.pp#L47

I can create a PR if the bug is accepted as a bug.

## Expected Behavior
The correct php library (`libphp.so`) should be loaded when using RedHat + PHP 8.

## Steps to Reproduce
1. Set `$php_version` to `8.0` in a RHEL environment.

## Environment
 - Puppetlabs/apache 8.1.0
 - RHEL 8.6

polatsinan:
Fixes loading correct libphp file when RedHat AND PHP 8.
Fixes #2332

### DIFF
--- a/spec/classes/mod/php_spec.rb
+++ b/spec/classes/mod/php_spec.rb
@@ -117,8 +117,8 @@ describe 'apache::mod::php', type: :class do
                     { php_version: '8.0' }
                   end
 
-                it { is_expected.to contain_apache__mod('php') }
-                it { is_expected.to contain_file('php.load').with(content: "LoadModule php_module modules/libphp.so\n") }
+                  it { is_expected.to contain_apache__mod('php') }
+                  it { is_expected.to contain_file('php.load').with(content: "LoadModule php_module modules/libphp.so\n") }
                 end
               elsif facts[:os]['release']['major'].to_i >= 9
                 it { is_expected.to contain_apache__mod('php') }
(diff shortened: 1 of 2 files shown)


ekohl (commented):
This matches line 96 to line 47 so :+1: from that perspective. I'd really like some tests in `spec/classes/mod/php_spec.rb` for this. Do you think you could write some?

ekohl (approved):
@polatsinan could you have a look at writing a test?

ekohl (dismissed):
Thanks!

spec/classes/mod/php_spec.rb
```diff
@@ -119,7 +119,7 @@
 
                 it { is_expected.to contain_apache__mod('php') }
                 it { is_expected.to contain_file('php.load').with(content: "LoadModule php_module modules/libphp.so\n") }
```
ekohl, line 121: I think this is also needed to please Rubocop
```suggestion
                  it { is_expected.to contain_apache__mod('php') }
                  it { is_expected.to contain_file('php.load').with(content: "LoadModule php_module modules/libphp.so\n") }
```

david22swan (approved):
LGTM

Look's like a good change to me :)
Thanks for putting in the work

PR #2329, opened 2022-10-10
Drop Apache 2.2 support

ekohl:
This is likely incomplete and untested at the moment. However, it shows how big the cleanup is.

PR #2334, opened 2022-10-13
(CONT-173) - Updating deprecated facter instances

jordanbreen28:
Prior to this commit, this module contained instances of Facter::Util::Resolution.exec and Facter::Util::Resolution.which, which are deprecated. This commit aims to replace these exec helpers with their supported Facter::Core::Execution counterparts.

This commit:
- Replaced all Facter::Util::Resolution instances with corresponding Facter::Core::Execution exec helpers

### DIFF
--- a/lib/facter/apache_version.rb
+++ b/lib/facter/apache_version.rb
@@ -5,17 +5,17 @@ Facter.add(:apache_version) do
   setcode do
     apache_version = nil
 
-    if Facter::Util::Resolution.which('httpd')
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I'll just note that in https://github.com/puppetlabs/puppetlabs-apache/pull/2329 I actually suggest to remove this fact altogether. Not sure if that will be accepted.

chelnak (approved):
Nice work!

Please can you update your PR description though? It references commit, and it should be PR.

PR #2335, opened 2022-10-13
Support setting SSLProxyCipherSuite on mod_ssl

ekohl:
While it is a parameter for apache::vhost, this allows setting it server wide just like SSLCipherSuite.

PR #2336, opened 2022-10-13
Update EL8+ and Debian SSL defaults

ekohl:
It has been a long time since the defaults were updated. This results in a safer out-of-the-box deployment, matching what the vendors do.

For reference, EL7 has `SSLCipherSuite HIGH:3DES:!aNULL:!MD5:!SEED:!IDEA` and `SSLProtocol all -SSLv2 -SSLv3`. The 3DES looks to be a mistake and I'd expect them to have meant `!3DES`. I have not updated that default nor checked SLES.


ekohl:
Includes https://github.com/puppetlabs/puppetlabs-apache/pull/2335 since EL8+ sets `SSLProxyCipherSuite` by default.

ekohl:
@david22swan @chelnak it would help me a **lot** if this was reviewed and merged this week. Currently the tests are still running, hoping they'll pass but I'd appreciate if you take a look at the general design.

ekohl:
I do not know why mod_md has started to fail. I've also never used mod_md myself so not sure what's wrong. Sadly the error log is not very useful. I've pushed some changes that I hope will help.

ekohl:
I decided to dive into how one would do this properly in rspec. https://github.com/voxpupuli/beaker-rspec/pull/115 is the result. I'll have a stab at porting this to Litmus, which should be very straight forward. That should allow me to provide sufficient debug info.

spec/spec_helper_acceptance_local.rb
```diff
@@ -191,3 +191,30 @@ def mod_supported_on_platform?(mod)
   return false if ENV['DISABLE_MOD_TEST_EXCLUSION']
   ApacheModTestFilterHelper.instance.mod_supported_on_platform?(mod)
 end
+
+RSpec::Matchers.define :apply do
```
ekohl, line 195: IMHO this matcher should have been implemented in beaker-rspec and then in puppet_litmus. It provide a much more native RSpec experience in writing tests.

ekohl:
Now we're getting somewhere:
> AH02311: Fatal error initialising mod_ssl, exiting. See /var/log/httpd/example.com_error_ssl.log for more information

ekohl:
So it fails with:
```
Wed Oct 26 10:48:29.313335 2022] [ssl:warn] [pid 13085] AH10085: Init: example.com:443 will respond with '503 Service Unavailable' for now. There are no SSL certificates configured and no other module contributed any.
[Wed Oct 26 10:48:29.313646 2022] [ssl:emerg] [pid 13085] AH01898: Unable to configure permitted SSL ciphers
[Wed Oct 26 10:48:29.313661 2022] [ssl:emerg] [pid 13085] SSL Library Error: error:1410D0B9:SSL routines:SSL_CTX_set_cipher_list:no cipher match
```

So it looks like mod_md and using `PROFILE=system` is incompatible.

I'm debating options now. One option is to modify the test to explicitly set a cipher list. I'll try to set up a testing system to verify.

ekohl:
> So it looks like mod_md and using `PROFILE=system` is incompatible.

Turns out it was much simpler than I thought. It only showed up with `mod_md` since that's the only HTTPS vhost. That means the ciphers aren't evaluated elsewhere and our mod_ssl acceptance test happily accepts it since it's syntactically valid. It's actually `PROFILE=SYSTEM`, not `PROFILE=system`.

ekohl:
@david22swan @chelnak test failures are now provisioning errors: it never even got to running the test suite. Can this be merged?

bastelfreak (dismissed):

ekohl:
No change, just forcing a new CI run.

david22swan (approved):
LGTM
Thank's for getting this in

PR #2338, opened 2022-10-20
Make serveradmin an optional parameter and use it

ekohl:
Since fc6cced03e60c5a78bceb006dd4267fc37f54959 the parameter no longer ends up in the main configuration. It was still used in apache::mod::shib until 049fd54b57e8a4f28ceeea4888456eafe1509fa1. Since then it's been unused.

This changes it to an optional non-empty string which means it doesn't change the default config, but still allows users to set the ServerAdmin in the main Apache config.


david22swan (approved):
LGTM

Had to take a few look's but this seem's good and the acceptance tests are showing no problems so I'm happy to merge

PR #2340, opened 2022-10-26
Extract acceptance node setup to a separate file

ekohl:
This makes it easier to maintain. You get proper syntax highlighting in your editor and can even lint it. This also converts some shell lines to pure Puppet. This reduces the need for large local spec_helper_acceptance modifications. The pattern is copied from voxpupuli-acceptance where the beaker setup automatically detects if the file is present and applies it.

It does drop EL6 EPEL support, following 4922544709313e0b7d176013e6f8118b85fdf96a.

### DIFF
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -1,0 +1,36 @@
+# needed by tests
+package { 'curl':
+  ensure => 'latest',
+}
+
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

david22swan (approved):
LGTM

Seem's a lot cleaner and neater now

PR #2342, opened 2022-11-14
Declare minimum Puppet version to be 6.24.0

### LINKED ISSUE #2328: Incorrect Type Passed to confd_command in init.pp
https://github.com/puppetlabs/puppetlabs-apache/blob/3b4bae8dbac8081e241d3049a520cc953705ceb5/manifests/init.pp#L640

 Error: Parameter command failed on Exec[mkdir /opt/rh/httpd24/root/etc/httpd/conf.d]: Command must be a String, got value of class Array (file: /etc/puppetlabs/code/environments/production/modules/apache/manifests/init.pp, line: 639)

ekohl:
In a565ea97609979f7ccbc407534a553def0e865d8 the commands are passed as an array, but this feature was only [introduced in Puppet 6.24.0](https://puppet.com/docs/puppet/6/release_notes_puppet.html#enhancements_puppet_x-7-9-0-PUP-5704). This raises the minimum version to match, since it's no longer possible to use the module on anything older.

Fixes #2328

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -83,7 +83,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 6.0.0 < 8.0.0"
+      "version_requirement": ">= 6.24.0 < 8.0.0"
     }
   ],
   "description": "Module for Apache configuration",


trefzer:
I could not figure out if  puppet minimal requirement declared in the metadata file is for the version of puppetserver or the puppetclient (anyone knows ?). 
The exec version changed in the puppet client. I think we should probably add a fail message in the puppet code if client is below 6.24 to lead people in the correct direction what they need to update !
(I know several installations with puppetserver > 6 but still clients <6 !!)

david22swan:
Agree on the change and would like it merged in, however will need to wait as there is still some discussion on our side and other work that we would want in alongside this as part of a major release.

ekohl:
> I could not figure out if puppet minimal requirement declared in the metadata file is for the version of puppetserver or the puppetclient (anyone knows ?).

I'd say at least the agent since it's perfectly valid to apply a module locally without a Puppetserver present. Additionally, it relies on types and providers supplied by the agent. As a last point, it has been a recommendation that Puppetserver >= agent so this does imply the server needs to be reasonably new.

However, AFAIK nothing formally checks this requirement programmatically. This is why [Kafo](https://github.com/theforeman/kafo) (the library driving Foreman's installer) [implements it](https://github.com/theforeman/kafo/blob/master/modules/kafo_configure/manifests/puppet_version_semver.pp)

> Agree on the change and would like it merged in, however will need to wait as there is still some discussion on our side and other work that we would want in alongside this as part of a major release.

I disagree this is breaking. It is just being explicit what was already implicit. If you think this is breaking, we should revert https://github.com/puppetlabs/puppetlabs-apache/commit/a565ea97609979f7ccbc407534a553def0e865d8 now and reintroduce it in a new major version.

chelnak:
FWIW I don't want to revert changes that are already in main and released.

I'd much rather us fix forward.

david22swan:
Yeah, we'll go ahead with a non major release.
It's the least bad option of a few bad options.

david22swan:
@ekohl Thanks for putting up the PR

david22swan (approved):

PR #2349, opened 2022-12-06
disable::mpm_event: Fix module deactivation

bastelfreak:
there were a few things broken:
* a2dismod was called with the wrong module name. at least on apache 2.4.54 on Debian 11 it is named mpm_event, not event
* the paths to the files were wrong
* a2dismod already purges the .load and .conf file, the exec resources
 are not required (even then a file resource would probably be cleaner
 to purge them)
* Apache needs to be reloaded, so just running this priot to
 apache2::service isn't enough, it needs to be a notify

This initially worked but was changed from mpm_event to event in https://github.com/puppetlabs/puppetlabs-apache/pull/1961. Then it was fixed in https://github.com/puppetlabs/puppetlabs-apache/pull/2034 by adding the exec resources, but the actual a2dismod was still broken.

### DIFF
--- a/manifests/mpm/disable_mpm_event.pp
+++ b/manifests/mpm/disable_mpm_event.pp
@@ -1,29 +1,11 @@
 # @summary disable Apache-Module event
 class apache::mpm::disable_mpm_event {
-  $event_command = ['/usr/sbin/a2dismod', 'event']
-  $event_onlyif = [['/usr/bin/test', '-e', join([$apache::mod_enable_dir, 'event.load'],'/')]]
-  exec { '/usr/sbin/a2dismod event':
+  $event_command = ['/usr/sbin/a2dismod', 'mpm_event']
+  $event_onlyif = [['/usr/bin/test', '-e', join([$apache::mod_enable_dir, 'mpm_event.load'],'/')]]
+  exec { '/usr/sbin/a2dismod mpm_event':
     command => $event_command,
     onlyif  => $event_onlyif,
     require => Package['httpd'],
-    before  => Class['apache::service'],
-  }
-
-  $event_load_command = ['/bin/rm', join([$apache::mod_enable_dir, 'event_event.load'],'/')]
-  $event_load_onlyif = [['/usr/bin/test', '-e', join([$apache::mod_enable_dir, 'event_event.load'],'/')]]
-  exec { 'remove distribution event load file':
-    command => $event_load_command,
-    onlyif  => $event_load_onlyif,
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I was wondering about differences in disable steps. We already have 2 identical disable steps: worker and prefork:
```console
$ diff -Nut <(sed 's/worker/prefork/g' disable_mpm_worker.pp) disable_mpm_prefork.pp 
$
```

With this change there is a third:
```console
$ diff -Nut <(sed 's/worker/event/g' disable_mpm_worker.pp) disable_mpm_event.pp 
--- /dev/fd/63	2022-12-06 11:53:19.373572653 +0100
+++ disable_mpm_event.pp	2022-12-06 11:53:00.571376870 +0100
@@ -1,11 +1,11 @@
 # @summary disable Apache-Module event
 class apache::mpm::disable_mpm_event {
-  $event_command = ['/usr/sbin/a2dismod', 'event']
-  $event_onlyif = [['/usr/bin/test', '-e', join([$apache::mod_enable_dir, 'event.load'],'/')]]
-  exec { '/usr/sbin/a2dismod event':
+  $event_command = ['/usr/sbin/a2dismod', 'mpm_event']
+  $event_onlyif = [['/usr/bin/test', '-e', join([$apache::mod_enable_dir, 'mpm_event.load'],'/')]]
+  exec { '/usr/sbin/a2dismod mpm_event':
     command => $event_command,
     onlyif  => $event_onlyif,
     require => Package['httpd'],
-    before  => Class['apache::service'],
+    notify  => Class['apache::service'],
   }
 }
```

There are 2 changes: first is that my replacement isn't quite correct, but really it would be if you introduced `$mpm_name`. The second is `before` vs `notify`. Looking at this file it was `before`, but `notify` looks more correct to me.

They're also (IMHO) private classes that are only used once:
https://github.com/puppetlabs/puppetlabs-apache/blob/1d2302da73286fbbf6f91103cb7f7268c416d57f/manifests/mpm.pp#L97-L108

Perhaps we should introduce a `define` and call it with the right module name?

bastelfreak:
I also noticed this and will try to clean it up. I updated the commit description with some history of this.

ekohl (dismissed):
There is an [a2mod type](https://github.com/puppetlabs/puppetlabs-apache/blob/main/lib/puppet/type/a2mod.rb). I wonder that should be used. However, in the short term I'll take this cleanup.

david22swan (approved):
LGTM

PR #2350, opened 2022-12-06
Drop deprecated a2mod type/providers

bastelfreak:
This type is deprecated, so it should be removed. I raise this PR so we won't forget to remove it. Please wait with merging until a new major release is planned.


ekohl:
In https://github.com/puppetlabs/puppetlabs-apache/issues/2291 we gather that. Perhaps it should have been a milestone, but at least link it up.

david22swan (approved):
LGTM
According to the changelog these where deprecated 9 years ago, so think it's about time we took them out.
Thanks for putting in the work :)

PR #2357, opened 2022-12-14
(CONT-262) Removal of Apache 2.2 Compatibility

david22swan:
A rebase of an earlier piece of work that was prepared by a former Puppet employee for the purpose of removing Apache 2.2, but was never merged due to push back from community and customers at the time.

Originally authored by former Puppet employee @HelenCampbell


manifests/mod/event.pp
```diff
@@ -99,16 +99,9 @@
...
-    'redhat': {
-      if versioncmp($_apache_version, '2.4') >= 0 {
-        apache::mpm { 'event':
-          apache_version => $_apache_version,
-        }
-      }
-    }
-    'debian', 'freebsd' : {
-      apache::mpm { 'event':
+  case $::osfamily {
```
bastelfreak, line 102: please don't reference facts as top scope variables als also please don't use legacy facts:
```suggestion
  case $facts['os']['family'] {
```

bastelfreak: when this case has no just a single match, maybe `if $facts['os']['family'] in ['Debian', 'FreeeBSD', 'RedHat']`

manifests/mod/event.pp
```diff
@@ -99,16 +99,9 @@
...
+    'debian','freebsd','redhat' : {
```
bastelfreak, line 103: They need to be capitalized?

manifests/mod/ssl.pp
```diff
@@ -118,27 +118,12 @@
...
-        $_ssl_mutex = 'default'
-      }
-      'freebsd': {
-        $_ssl_mutex = 'default'
-      }
-      'gentoo': {
-        $_ssl_mutex = 'default'
-      }
-      'Suse': {
+    case $::osfamily {
```
bastelfreak, line 125: same as above, please don't replace structed facts with a legacy fact as topscope variable.

manifests/mpm.pp
```diff
@@ -127,20 +110,18 @@
...
-        file { "${apache::mod_enable_dir}/${mpm}.load":
-          ensure  => link,
-          target  => "${apache::mod_dir}/${mpm}.load",
-          require => Exec["mkdir ${apache::mod_enable_dir}"],
-          before  => File[$apache::mod_enable_dir],
-          notify  => Class['apache::service'],
-        }
+      file { "${::apache::mod_enable_dir}/${mpm}.load":
+        ensure  => link,
+        target  => "${::apache::mod_dir}/${mpm}.load",
```
bastelfreak, line 115:
```suggestion
        target  => "${apache::mod_dir}/${mpm}.load",
```

spec/classes/mod/itk_spec.rb
```diff
@@ -70,35 +47,27 @@
   end
 
   context 'on a RedHat OS' do
+<<<<<<< refs/remotes/upstream/main
```
bastelfreak, line 50: some rebase leftovers

templates/vhost/_directories.erb
```diff
@@ -2,19 +2,17 @@
...
+      <%- scope.function_warning(["Apache::Vhost: Using allow is deprecated in your Apache version"]) -%>
+    <%- end -%>
+    <%- if directory['deny'] and ! [ false, 'false', '' ].include?(directory['deny']) -%>
+      <%- scope.function_warning(["Apache::Vhost: Using deny is deprecated in your Apache version"]) -%>
+    <%- end -%>
+    <%- if directory['order'] and ! [ false, 'false', '' ].include?(directory['order']) -%>
+      <%- scope.function_warning(["Apache::Vhost: Using order is deprecated in your Apache version"]) -%>
+    <%- end -%>
+    <%- if directory['satisfy'] and ! [ false, 'false', '' ].include?(directory['satisfy']) -%>
+      <%- scope.function_warning(["Apache::Vhost: Using satisfy is deprecated in your Apache version"]) -%>
```
ekohl, line 15: In my PR I went further and simply killed the deprecated parameters. We're going for a major version bump: let's make it clean.

spec/classes/mod/mime_magic_spec.rb
```diff
@@ -44,7 +44,21 @@ def general_mime_magic_specs
   end
 
   context 'on a RedHat OS with default params' do
+<<<<<<< refs/remotes/upstream/main
```
bastelfreak, line 47: rebase leftovers

spec/defines/vhost_spec.rb
```diff
@@ -7,15 +7,63 @@
...
+    describe 'basic assumptions' do
+      let(:params) { default_params }
+
+      it { is_expected.to contain_class('apache') }
+      it { is_expected.to contain_class('apache::params') }
+      it { is_expected.to contain_apache__listen(params[:port]) }
+    end
+    context 'set everything!' do
+      let :params do
+>>>>>>> Modification of version code and apache_spec fixes
```
ekohl, line 66: Conflicts here.

ekohl (changes requested):
I still think https://github.com/puppetlabs/puppetlabs-apache/pull/2329 is clear. That actually removes lines and this adds lines.

Though it's probably because it adds all parameters back to README.md instead of using `REFERENCE.md`. So please start with removing those again and then we can see how much remains.

Personally I'd start with my PR and then rebase this work on top of it to see if there's anything left.

manifests/mod/event.pp
david22swan: Yeh, only just noticed this. Missed that it was overwritten by the rebasse

manifests/mod/event.pp
david22swan: don't think so

ekohl: Puppet does case insensitive comparisons on a case but the fact is capitalized. So if you change it to `if $::osfamily == 'redhat'` it will fail. That must be `if $::osfamily == 'RedHat`. Because of that I always try to keep the capitalization correct, even if it doesn't matter.

bastelfreak:
> Personally I'd start with my PR and then rebase this work on top of it to see if there's anything left.
I agree with @ekohl

PR #2362, opened 2022-12-20
(CONT-406) Fix for RHEL 7 compatibility

david22swan:
A previous PR had updated the security_crs.comf.erb file to a more modern version, however this has rendered it incompatible with RHEL 7. This PR introduces and if to solve the issue.

Also includes a quick fix to remove legacy facts from a template file.


templates/mod/security.conf.erb
```diff
@@ -54,13 +54,13 @@
     <%- end -%>
     SecArgumentSeparator &
     SecCookieFormat 0
-<%- if scope.lookupvar('::osfamily') == 'Debian' -%>
+<%- if $facts['os']['family'] == 'Debian' -%>
```
ekohl, line 57: This is ERB, not EPP

ekohl (changes requested):
I wouldn't mind converting to EPP eventually, but we're not there today.

templates/mod/security.conf.erb
david22swan: crap, read the wrong guide

GSPatton (approved):
looks good to me

PR #2365, opened 2023-01-10
vhost: Make ProxyAddHeaders configureable

bastelfreak:
It's a valid usecase to set ProxyAddHeaders to Off. That's currently only possible when other proxy settings are configured as well. It needs to be independant.


ekohl (approved):

david22swan (approved):
LGTM

PR #2380, opened 2023-02-02
Adding mod_version module to be activated by default

Q-Storm:
re adding mod_version which was accidentally removed with the drop of "apache 2.2 Support" in #2329.

### DIFF
--- a/manifests/mod/version.pp
+++ b/manifests/mod/version.pp
@@ -1,0 +1,12 @@
+# @summary
+#   Installs `mod_version`.
+#
+# @see https://httpd.apache.org/docs/current/mod/mod_version.html for additional documentation.
+#
+class apache::mod::version {
+  if $facts['os']['family'] == 'debian' {
+    warning("${module_name}: module version_module is built-in and can't be loaded")
+  } else {
+    ::apache::mod { 'version': }
+  }
+}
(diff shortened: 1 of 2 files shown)


jnk0:
As a user of puppetlabs-apache I strongly support this PR as it re-adds the accidentally removed module version, which is an essential module and needs to be part of default_mods too. Just think about .htaccess files distributed with various software or software addons.

Q-Storm:
Thanks @jnk0 I added the the module to the default_mods.pp

david22swan (approved):
LGTM

Mend error is unrelated

manifests/mod/version.pp
```diff
@@ -0,0 +1,12 @@
+# @summary
+#   Installs `mod_version`.
+#
+# @see https://httpd.apache.org/docs/current/mod/mod_version.html for additional documentation.
+#
+class apache::mod::version {
+  if $facts['os']['family'] == 'debian' {
```
ekohl, line 7: `Debian` , if you want to get the casing right. Though I see it's also wrong in many other places.

manifests/mod/version.pp
```diff
@@ -0,0 +1,12 @@
...
+    warning("${module_name}: module version_module is built-in and can't be loaded")
+  } else {
+    ::apache::mod { 'version': }
```
ekohl, line 10: The leading `::` is not needed. AFAIK it never was for defines, but certainly since we dropped Puppet < 4 it's no longer needed.

ekohl (commented):
It would have been nice to at least add a unit test.

PR #2381, opened 2023-02-02
(BugFix) Update OS Family comparison to correctly match

david22swan:
OS Family comparison should be against `RedHat` not `Redhat`

### DIFF
--- a/spec/classes/mod/security_spec.rb
+++ b/spec/classes/mod/security_spec.rb
@@ -159,24 +159,42 @@ describe 'apache::mod::security', type: :class do
               }
             end
 
-            it {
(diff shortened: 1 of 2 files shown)


ekohl (dismissed):

spec/classes/mod/security_spec.rb
```diff
@@ -158,25 +158,41 @@
                 dos_block_timeout: 300,
               }
             end
```
ekohl, line 160: Looks like the RuboCop annotations made it in, now that puppetlabs_spec_helper is released.

david22swan: fixed

pmcmaw (approved):

PR #2384, opened 2023-02-08
Fix example code for apache::vhost::php_values

gcoxmoz:
The `php_values` parameter requires a Hash, but gives examples from the ~v7 era when other values were allowed.


david22swan (approved):
LGTM

Issue #2352, opened 2022-12-07
mod proxy class does not allow array of IP addresses in param allow_from


jmcnatt:
## Describe the Bug
Class `apache::mod::proxy` parameter `allow_from` expects a string, but should accept a list (array) of IP addresses.

Error:
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Class[Apache::Mod::Proxy]: parameter 'allow_from' expects a value of type Undef or String, got Tuple
```

## Expected Behavior
An array of IPs should be accepted, either through hiera or through resource-style declaration.

## Steps to Reproduce
1. Add the following to a profile:
```puppet
include apache

class { 'apache::mod::proxy':
  allow_from => [
    '192.168.0.1',
    '192.168.0.2',
  ],
}
```

## Environment
 - Module version:  8.4.0
 - Platform:  CentOS 7

## Additional Context
The issue surfaced with [this commit](https://github.com/puppetlabs/puppetlabs-apache/commit/ae56cc9a6dbcd3846ccbcb9f8a7b9302d8b2a996#diff-a022756b9c37d25d2b9d8ea3403a1e08d7b3e044b448bc5b4c8342bf66f56b29) that specifically typed the params.

The template [proxy.conf.erb](https://github.com/puppetlabs/puppetlabs-apache/blob/0b74e1dc22f28fd660a45191adb48199cea3c5a3/templates/mod/proxy.conf.erb#L14) even references an array to split.

ekohl:
Sounds like a regression indeed. The parameter documentation also says "list of IPs". Could you come up with a patch?

pebtron:
I can confirm this is broken on RHEL7 + module version 9.0.1. `templates/mod/proxy.conf.erb` assumes `allow_from` is an array when `manifests/mod/proxy.pp` specifies it to be of type `Stdlib::IP::Address`.

david22swan:
Fix has been merged in, should be release by end of day

PR #2385, opened 2023-02-08
Fixes mod::proxy allow_from parameter inconsistency #2352

pebtron:
apache::mod::proxy allowed Stdlib::IP::Address as a class parameter, but proxy.conf.erb assumed it was an array.


david22swan (approved):
LGTM

Thank's for catching this.
Looks like a good change to me

PR #2387, opened 2023-02-09
(BUGFIX) Update to ensure correct facter comparisons

david22swan:
Ensuring that the correct values are compared against facter, there seem to be many cases where lowercase has been used in place of uppercase.

### DIFF
--- a/manifests/default_confd_files.pp
+++ b/manifests/default_confd_files.pp
@@ -8,7 +8,7 @@ class apache::default_confd_files (
   # The rest of the conf.d/* files only get loaded if we want them
   if $all {
     case $facts['os']['family'] {
-      'freebsd': {
+      'FreeBSD': {
         include apache::confd::no_accf
       }
       default: {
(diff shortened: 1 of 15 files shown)


ekohl (dismissed):

pmcmaw (approved):

ekohl:
Out of interest: how did you find this? I'm personally in the camp of "there should be one way and one way only", which implies always using the same casing so big :+1: to this. I just wonder, did you ran into this because of some bug/tooling issue?

david22swan:
@ekohl Sorry for the late reply, came up in regards to the fix for the templates in my other PR: https://github.com/puppetlabs/puppetlabs-apache/pull/2381/files
Seem's that the checks in `.erb` files are case sensitive.
After getting that merged in just decided to double check everything else and when I saw the lack of consistency decided to do a quick fix just in case.

ekohl:
Oh yes, ERB is Ruby so case sensitive comparisons. I dislike that Puppet does case insensitive comparisons.

Issue #2319, opened 2022-09-21
apache::mod::security is not RHEL 8 compliant.


d0n0van:
## Describe the Bug
apache::mod::security is not RHEL 8 compliant.
mod_security_crs-3 incompatible, different path and names.

## Expected Behavior
.

## Steps to Reproduce
install on rhel8

## Environment
 - RHEL 8

## Additional Context
new mod_security_crs-3 in on:
$modsec_crs_path = '/usr/share/mod_modsecurity_crs'

activated_rules => [
"$modsec_crs_path/rules/crawlers-user-agents.data",
"$modsec_crs_path/rules/iis-errors.data",
"$modsec_crs_path/rules/java-classes.data",
"$modsec_crs_path/rules/java-code-leakages.data",
"$modsec_crs_path/rules/java-errors.data",
"$modsec_crs_path/rules/lfi-os-files.data",
"$modsec_crs_path/rules/php-config-directives.data",
"$modsec_crs_path/rules/php-errors.data",
"$modsec_crs_path/rules/php-function-names-933150.data",
"$modsec_crs_path/rules/php-function-names-933151.data",
"$modsec_crs_path/rules/php-variables.data",
"$modsec_crs_path/rules/REQUEST-901-INITIALIZATION.conf",
"$modsec_crs_path/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf",
"$modsec_crs_path/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf",
"$modsec_crs_path/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf",
"$modsec_crs_path/rules/REQUEST-903.9004-DOKUWIKI-EXCLUSION-RULES.conf",
"$modsec_crs_path/rules/REQUEST-903.9005-CPANEL-EXCLUSION-RULES.conf",
"$modsec_crs_path/rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf",
"$modsec_crs_path/rules/REQUEST-905-COMMON-EXCEPTIONS.conf",
"$modsec_crs_path/rules/REQUEST-910-IP-REPUTATION.conf",
"$modsec_crs_path/rules/REQUEST-911-METHOD-ENFORCEMENT.conf",
"$modsec_crs_path/rules/REQUEST-912-DOS-PROTECTION.conf",
[...]

PR #2398, opened 2023-03-27
Fix mod_security rules on RHEL 8+

Tamerz:
As mentioned here, the rules have moved in RHEL 8 and newer: https://github.com/puppetlabs/puppetlabs-apache/issues/2319

This updates the rules to match what is in the newer packages.


Tamerz:
I don't have time to work on this.

PR #2404, opened 2023-04-14
Add the missing mod_authnz_ldap parameters

chutzimir:
This is the full list from https://httpd.apache.org/docs/2.4/mod/mod_authnz_ldap.html

Following the current style of CamelCase -> snake_case

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -299,38 +299,23 @@
     AuthLDAPGroupAttribute <%= groupattr %>
       <%- end -%>
     <%- end -%>
-    <%- if directory['auth_ldap_group_attribute_is_dn'] == 'off' -%>
-    AuthLDAPGroupAttributeIsDN Off
-    <%- end -%>
-    <%- if directory['auth_ldap_group_attribute_is_dn'] == 'on' -%>
-    AuthLDAPGroupAttributeIsDN On
+    <%- unless directory['auth_ldap_group_attribute_is_dn'].nil? -%>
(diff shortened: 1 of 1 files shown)


templates/vhost/_directories.erb
```diff
@@ -305,6 +305,66 @@
     <%- if directory['auth_ldap_group_attribute_is_dn'] == 'on' -%>
     AuthLDAPGroupAttributeIsDN On
     <%- end -%>
+    <%- if directory['auth_ldap_initial_bind_pattern'] -%>
+    AuthLDAPInitialBindPattern <%= directory['auth_ldap_initial_bind_pattern'] %>
+    <%- end -%>
+    <%- if directory['auth_ldap_bind_as_user'] == 'off' -%>
+    AuthLDAPInitialBindAsUser Off
+    <%- end -%>
```
ekohl, line 313: Would it be better to use this pattern for booleans?
```suggestion
    <%- unless directory['auth_ldap_bind_as_user'].nil? -%>
    AuthLDAPInitialBindAsUser <%= scope.call_function('apache::bool2httpd', [directory['auth_ldap_bind_as_user']]) %>
    <%- end -%>
```

chutzimir: I just followed the pattern in the file, and I agree it would be lovely if we changed everything to a single call. Should this be a separate PR?

Ramesh7 (approved):

Ramesh7:
Thanks @chutzimir for your contribution, the changes looks good and ready to ship.

PR #2408, opened 2023-04-26
Remove has_key usage

evgeni:
In the next puppetlabs-stdlib major version this function will be dropped. Native Puppet is used instead.

https://github.com/puppetlabs/puppetlabs-stdlib/pull/1319 broke this


chelnak (approved):

marszip:
When will this be released so we can update to the last stdlib version?

PR #2415, opened 2023-06-05
use a dedicated Enum type for On/Off and accept lower case too

evgeni:


bastelfreak (dismissed):

bastelfreak (approved):

smortex (approved):

jordanbreen28:
Thanks @evgeni. I like this approach, are you able to rebase so we can get this merged?

jordanbreen28 (approved):
LGTM. thanks again @evgeni!

PR #2429, opened 2023-06-29
(MAINT) Enable legacy SP3 repository for SUSE

Ramesh7:
## Summary

Enable legacy SP3 repo so that we can leverage the old packages without any issues.

## Additional Context

New release of SUSE 15 introduced new version SP4 which deprecated old packages where `apache2-mod_php7` is one of them. As the package is deprecated with SP4 repo so user will not able to install.

## Related Issues (if any)
Due to package got deprecated from new repo, respective package manager is not able to install.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/mod/php.pp
+++ b/manifests/mod/php.pp
@@ -106,6 +106,16 @@ class apache::mod::php (
   }
 
   if $facts['os']['name'] == 'SLES' {
+    # Enable legacy repo to install apache2-mod_php7 package
+    # if SUSE OS major version is >= 15 and minor version is > 3
+    if ($_package_name == 'apache2-mod_php7' and versioncmp($facts['os']['release']['major'], '15') >= 0 and versioncmp($facts['os']['release']['minor'], '3') == 1) {
+      exec { 'enable legacy repos':
+        path    => '/bin:/usr/bin/:/sbin:/usr/sbin',
+        command => 'SUSEConnect --product sle-module-legacy/15.4/x86_64',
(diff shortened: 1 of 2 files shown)


manifests/mod/php.pp
```diff
@@ -106,6 +106,14 @@
   }
 
   if $facts['os']['name'] == 'SLES' {
+    if $_package_name == 'apache2-mod_php7' and versioncmp($facts['os']['release']['major'], '0') == '15' {
```
jordanbreen28, line 109: don't think this is the right usage of versioncmp, it returns an integer between -1 and 1.
you can either do:
```suggestion
    if $_package_name == 'apache2-mod_php7' and versioncmp($facts['os']['release']['major'], '15') >= 0 {
```
or like in ntp:
```suggestion
    if $_package_name == 'apache2-mod_php7' and $facts['os']['release']['major'] == '15' {
```

Ramesh7: `versioncmp($facts['os']['release']['major'], '15')` doesn't seems to be working, hence taking the ntp one :)

spec/setup_acceptance_node.pp
```diff
@@ -5,6 +5,13 @@
 
 case $facts['os']['family'] {
   'SLES', 'SUSE': {
+    if $facts['os']['release']['major'] == '15' {
```
Ramesh7, line 8: While running spec need to enable legacy repo for netstat package.

GSPatton (approved):

jordanbreen28 (approved):

PR #2436, opened 2023-07-28
CAT-1261-update_SUSE_repo_name

praj1001:
## Summary
Enable legacy SP5 repo so that we can leverage the old packages without any issues.

## Additional Context
Add any additional context about the problem here
New release of SUSE introduced new version SP5 which deprecated old package apache2-mod_php7. As the package is deprecated with old SP4 repo so user will not able to install.

## Related Issues (if any)
Due to package got deprecated from new repo, respective package manager is not able to install.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


Ramesh7:
Can you please update PR template as well as link this change to some ticket for better tracking. Thanks

Issue #2260, opened 2022-07-11
Add support for mod_log_forensic (ForensicLog)


Sebbo94BY:
## Use Case
I would like to be able to enable / install and configure the ``mod_log_forensic``, so that I can enable it using a Hiera variable in development/test for example, while it's disabled in production.

## Describe the Solution You Would Like
It would be great, when this project supports the setup / configuration of the ``ForensicLog`` line in the apache configuration file.

## Describe Alternatives You've Considered
I would need to "hack" the required configuration around this puppet project to make it possible to enable and configure the forensic module.

## Additional Context
Documentation: https://httpd.apache.org/docs/2.4/mod/mod_log_forensic.html

Sebi94nbg:
This is still required.

ekohl:
It's probably best to come up with a PR yourself. Until then I think it's ok to keep this issue open as a feature request.

PR #2442, opened 2023-08-14
(CAT-1283) - Enable forensic module

Ramesh7:
## Summary
Adding support for `log_forensic` module with apache.

## Additional Context
- https://github.com/puppetlabs/puppetlabs-apache/issues/2260

## Related Issues (if any)
- https://github.com/puppetlabs/puppetlabs-apache/issues/2260

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/manifests/mod/log_forensic.pp
+++ b/manifests/mod/log_forensic.pp
@@ -1,0 +1,9 @@
+# @summary
+#   Installs `mod_log_forensic`
+# 
+# @see https://httpd.apache.org/docs/current/mod/mod_log_forensic.html for additional documentation.
+#
+class apache::mod::log_forensic {
+  include apache
+  apache::mod { 'log_forensic': }
+}
--- a/spec/classes/mod/log_forensic_spec.rb
+++ b/spec/classes/mod/log_forensic_spec.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'apache::mod::log_forensic', type: :class do
+  ['Debian 11', 'RedHat 8'].each do |os|
+    context "on a #{os} OS" do
+      include_examples os
+
+      it { is_expected.to contain_class('apache::params') }
+      it { is_expected.to contain_class('apache::mod::log_forensic') }
+      it { is_expected.to contain_apache__mod('log_forensic') }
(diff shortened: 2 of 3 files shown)


spec/classes/mod/log_forensic_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'apache::mod::log_forensic', type: :class do
+  ['Debian 11', 'RedHat 8', 'FreeBSD 9', 'Gentoo'].each do |os|
+    context 'on a Debian OS' do
```
smortex, line 7: "Debian" here is suspect

spec/classes/mod/log_forensic_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'apache::mod::log_forensic', type: :class do
+  ['Debian 11', 'RedHat 8', 'FreeBSD 9', 'Gentoo'].each do |os|
```
smortex, line 6: "FreeBSD 9" is EOL (same for 10 and 11 and 12 in a few month), I guess this was copy-pasted, maybe we can adjust this (but I do not think any code is FreeBSD-version dependent so I would be surprised if it change anything)?

manifests/mod/log_forensic.pp
```diff
@@ -0,0 +1,9 @@
+# @summary
+#   Installs `mod_log_forensic`
+# 
+# @see https://httpd.apache.org/docs/current/mod/mod_log_forensic.html for additional documentation.
+#
+class apache::mod::log_forensic {
+  include apache
+  ::apache::mod { 'log_forensic': }
```
smortex, line 8:
```suggestion
  apache::mod { 'log_forensic': }
```

This legacy syntax is from the Puppet 3 era IIRC.  The module use it a lot, but maybe we can avoid it in new code?

smortex (commented):
Looks fine, only some cosmetics flaws we may want to fix?

GSPatton (approved):

smortex (approved):

PR #2444, opened 2023-08-18
Raise Puppet lower bound to >= 7.9.0

ekohl:
Similar to https://github.com/puppetlabs/puppetlabs-apache/commit/eb5b4c04e42c1711ef5d851e493a76ea4420ba63 this module uses [PUP-5704](https://www.puppet.com/docs/puppet/7/release_notes_puppet.html#enhancements_puppet_x-7-9-0-PUP-5704) which was only introduced in Puppet 7.10.0.

Fixes: 8070478d41a8c2f3a912b05904ac77a0b08bbe04 ("(CONT-772) Update metadata.json and Gemfile")


bastelfreak (approved):

smortex:
Looks fine (419078920b03900979e3d4b6a893c55f17a03b56 introduced such changes a while ago), but your link says it was part of Puppet 7.9.0, not 7.10.0.  Was there some issue that prevented this from working correctly in 7.9.0 or is it a typo?

ekohl:
Good catch. Not a typo, but I used `git tag --contains 59d045bb65260b7e9c5bf979da2bd7421cebfbaf` which doesn't sort like versions by default, so 7.10 shows up before 7.9. Updated now.

smortex (approved):

Ramesh7:
Wanted to get more context around `exec`'s `command` (only for *.nix) along with `unless` and `onlyif` arguments were already accepting array and as part of [419078](https://github.com/puppetlabs/puppetlabs-apache/commit/419078920b03900979e3d4b6a893c55f17a03b56) only see above arguments changed to array which was already supported in older version as well.
I am sure I am missing something here so please help me to get missing piece and also correct if my understanding is not right.

ekohl:
@Ramesh7 I'm not sure what you're exactly asking, but the ability to pass an array was only added in Puppet 7.9.0 (and Puppet 6.24.0). So you can't use this module with Puppet 7.8.0. The version_requirement should reflect that.

Ramesh7:
> @Ramesh7 I'm not sure what you're exactly asking, but the ability to pass an array was only added in Puppet 7.9.0 (and Puppet 6.24.0). So you can't use this module with Puppet 7.8.0. The version_requirement should reflect that.

Looks like I was checking on Puppet 6.28, which has this implemented and that's where I must have confused that as its already there.

Ramesh7 (approved):

ekohl:
It is confusing that it's >= 6.24 < 7, >= 7.9. I've removed the backwards-incompatible label because it's not. It's just clarifying what was previously implicit. I'd consider this a bugfix.

PR #2445, opened 2023-08-29
(CAT-1351) - Downgrading puppet-lint gem

Ramesh7:
## Summary
Recently rolled-out puppet-lint gem checking title as well for unsafe interpolation which seems to be a issue with feature so, downgrading the gem version to avoid the unsafe interpolation issue for title. 
Have reported the same to Team.

## Additional Context
- [x] Unsafe interpolation issue reported by puppet-lint (https://github.com/puppetlabs/puppet-lint/pull/142)

## Related Issues (if any)
- N/A

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


Ramesh7:
> 👎 on this. In [puppetlabs/puppet-lint#147](https://github.com/puppetlabs/puppet-lint/pull/147) it's proposed to drop the check instead, which would avoid requiring to do updates everywhere.

Thanks @ekohl, just now noticed that we are reverting the unsafe check. Closing the same.

PR #2455, opened 2023-09-08
changed server signature

praj1001:
## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


manifests/vhost.pp
```diff
@@ -2368,7 +2368,7 @@
   concat::fragment { "${name}-serversignature":
     target  => "${priority_real}${filename}.conf",
     order   => 90,
-    content => template('apache/vhost/_serversignature.erb'),
+    content => epp('apache/vhost/_serversignature.epp'),
```
smortex, line 2371: The file has no substitution, does it make sense to use a template rather than a file?

dinesh-gujar: we tried removing but it was failing the acceptance so created the file

ekohl: I think @smortex suggested to use `file()` instead of `epp()`. Then store it in `files/` instead of `templates`.

dinesh-gujar: sure will try that

Ramesh7:
Checked with author and the PR is no more needed, hence closing the same.

PR #2326, opened 2022-10-07
pdksync - (CONT-189) Remove support for RedHat6 / OracleLinux6 / Scientific6

david22swan:
CONT-189/remove_os_support
pdk version: `2.5.0`

PR #2462, opened 2023-09-20
(CAT-1424)-Removal of redhat/scientific/oraclelinux 6 for apache module

praj1001:
## Summary
With the removal of support for RHEL6 from our modules, work must be done to cleanup and remove any residual code.
This PR does the same for Apache repo module. More info can be found #2326 

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/spec/acceptance/itk_spec.rb
+++ b/spec/acceptance/itk_spec.rb
@@ -5,38 +5,22 @@ require 'spec_helper_acceptance'
 case os[:family]
 when 'debian', 'ubuntu'
   service_name = 'apache2'
-  variant = :prefork
 when 'redhat'
-  unless %r{^5}.match?(os[:release])
-    variant = (os[:release].to_i >= 7) ? :prefork : :itk_only
-    service_name = 'httpd'
-  end
+  service_name = 'httpd'
 when 'freebsd'
   service_name = 'apache24'
-  variant = :prefork
 end
 
 # IAC-787: The http-itk mod package is not available in any of the standard RHEL/CentOS 8.x repos. Disable this test
 # on those platforms until we can find a suitable source for this package.
 describe 'apache::mod::itk class', if: service_name && mod_supported_on_platform?('apache::mod::itk') do
   describe 'running puppet code' do
-    let(:pp) do
-      case variant
-      when :prefork
-        <<-MANIFEST
+    pp = <<-MANIFEST
             class { 'apache':
               mpm_module => 'prefork',
             }
             class { 'apache::mod::itk': }
-        MANIFEST
-      when :itk_only
-        <<-MANIFEST
-            class { 'apache':
-              mpm_module => 'itk',
-            }
-        MANIFEST
-      end
-    end
+    MANIFEST
 
     it 'behaves idempotently' do
       idempotent_apply(pp)
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -2020,8 +2020,7 @@ describe 'apache::vhost', type: :define do
             it { is_expected.to compile }
             it { is_expected.to contain_concat('25-rspec.example.com.conf') }
 
-            if (os_facts[:os]['family'] == 'RedHat' && os_facts[:os]['release']['major'].to_i > 6) ||
-               (os_facts[:os]['name'] == 'SLES' && os_facts[:os]['release']['major'].to_i > 11)
+            if os_facts[:os]['family'] == 'RedHat' || os_facts[:os]['name'] == 'SLES'
               it {
                 expect(subject).to contain_concat__fragment('rspec.example.com-directories').with(
                   content: %r{^\s+Require all granted$},
--- a/spec/util/apache_mod_platform_compatibility_spec.rb
+++ b/spec/util/apache_mod_platform_compatibility_spec.rb
@@ -12,10 +12,10 @@ describe ApacheModPlatformCompatibility do
   foobar_linux = 'foobar_linux'
 
   expected_compatible_platform_versions = {
-    'redhat' => [6, 7, 8],
-    'centos' => [6, 7, 8],
-    'oraclelinux' => [6, 7],
-    'scientific' => [6, 7],
+    'redhat' => [7, 8],
+    'centos' => [7, 8],
+    'oraclelinux' => [7],
+    'scientific' => [7],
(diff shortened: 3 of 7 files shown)


spec/acceptance/itk_spec.rb
```diff
@@ -7,10 +7,8 @@
   service_name = 'apache2'
   variant = :prefork
 when 'redhat'
-  unless %r{^5}.match?(os[:release])
     variant = (os[:release].to_i >= 7) ? :prefork : :itk_only
     service_name = 'httpd'
```
ekohl, line 11:
```suggestion
  variant = :prefork
  service_name = 'httpd'
```

However, with that there is nothing left that's the `itk_only` variant and that can also be cleaned up now.

templates/mod/security_crs.conf.erb
```diff
@@ -1,4 +1,4 @@
-<% if scope['facts']['os']['family'] == 'RedHat' and scope['facts']['os']['release']['major'].to_i <= 7 -%>
+<% if scope['facts']['os']['family'] == 'RedHat' and scope['facts']['os']['release']['major'].to_i == 7 -%>
```
ekohl, line 1: I'd keep this the same. It doesn't make any difference and fewer changes in git is easier to review.
```suggestion
<% if scope['facts']['os']['family'] == 'RedHat' and scope['facts']['os']['release']['major'].to_i <= 7 -%>
```

spec/defines/vhost_spec.rb
```diff
@@ -2020,7 +2020,7 @@
             it { is_expected.to compile }
             it { is_expected.to contain_concat('25-rspec.example.com.conf') }
 
-            if (os_facts[:os]['family'] == 'RedHat' && os_facts[:os]['release']['major'].to_i > 6) ||
+            if (os_facts[:os]['family'] == 'RedHat' && os_facts[:os]['release']['major'].to_i >= 7) ||
                (os_facts[:os]['name'] == 'SLES' && os_facts[:os]['release']['major'].to_i > 11)
```
ekohl, line 2024: This can be simplified to:
```suggestion
            if os_facts[:os]['family'] == 'RedHat' || os_facts[:os]['name'] == 'SLES'
```

But really, I think the whole `if` condition can be dropped. It's true for everything now.

manifests/params.pp
```diff
@@ -120,7 +120,7 @@
     $secpcrematchlimit = 1500
     $secpcrematchlimitrecursion = 1500
     $modsec_secruleengine = 'On'
-    if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '7') <= 0 {
+    if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '7') == 0 {
```
ekohl, line 123: I'd keep this, since it's no difference at runtime and fewer changes in git. Same below on line 266.

ekohl (changes requested):
I like cleaning up.

spec/acceptance/itk_spec.rb
praj1001: Oh I missed it. Thanks for pointing out @ekohl

spec/acceptance/itk_spec.rb
```diff
@@ -7,8 +7,8 @@
   service_name = 'apache2'
   variant = :prefork
 when 'redhat'
-    variant = (os[:release].to_i >= 7) ? :prefork : :itk_only
-    service_name = 'httpd'
+  variant = :prefork
```
ekohl, line 10: Note the second part I wrote: every one is now `:prefork`. So you can simplify the whole file.

spec/defines/vhost_spec.rb
```diff
@@ -2020,8 +2020,7 @@
             it { is_expected.to compile }
             it { is_expected.to contain_concat('25-rspec.example.com.conf') }
 
-            if (os_facts[:os]['family'] == 'RedHat' && os_facts[:os]['release']['major'].to_i > 6) ||
-               (os_facts[:os]['name'] == 'SLES' && os_facts[:os]['release']['major'].to_i > 11)
+            if os_facts[:os]['family'] == 'RedHat' || os_facts[:os]['name'] == 'SLES'
```
ekohl, line 2023: Note I said that you probably can remove the whole if and always check for the full bit, including content.

spec/util/apache_mod_platform_compatibility_spec.rb
```diff
@@ -12,10 +12,10 @@
...
   expected_compatible_platform_versions = {
-    'redhat' => [6, 7, 8],
-    'centos' => [6, 7, 8],
-    'oraclelinux' => [6, 7],
-    'scientific' => [6, 7],
+    'redhat' => [7, 8],
+    'centos' => [7, 8],
+    'oraclelinux' => [7],
+    'scientific' => [7],
     'debian' => [8, 9, 10],
```
ekohl, line 19: Is dropping 8 & 9 here scheduled for a future PR? As well as Ubuntu cleanup just below.

ekohl (changes requested):

spec/util/apache_mod_platform_compatibility_spec.rb
praj1001: We will be creating a separate thorough ticket just for removing OS support and their residual code wherever required. As for the part of this ticket, we have primarily focused on redhat family OS and that too Redhat , Scientific and OracleLinux 6.
This ticket is the successive ticket of https://github.com/puppetlabs/puppetlabs-apache/pull/2326/files
Anyway thanks for suggestion and reminding us.

spec/defines/vhost_spec.rb
praj1001: Thanks for the suggestion @ekohl , Also I have tried this with removing if condition which usually results in failing parallel specs.
Currently content part has been tested for Redhat and SLES OS, for other OS content part is not checked only concatenated fragment part is checked. So I believe we need this part.

Ramesh7 (approved):

spec/util/apache_mod_platform_compatibility_spec.rb
ekohl: Makes sense, thanks.

ekohl (approved):

PR #2463, opened 2023-09-21
Cat 1347 erb to epp conversion vhost dir

dinesh-gujar:
## Summary
Converting ERB to EPP


Ramesh7:
not needed, hence closing the same

PR #2469, opened 2023-09-22
Remove deprecated scriptaliases parameter

ekohl:
I haven't updated REFERENCE.md.

PR #2470, opened 2023-09-22
(CAT-1449) - Remove deprecated parameters for scriptaliases & passenger

Ramesh7:
## Summary
Cleanup for deprecated content : 
- scriptaliases parameter
- passanger params

## Additional Context
Thanks @ekohl for pointing out the deprecated content and raising the bar for apache module.
Also thanks https://github.com/puppetlabs/puppetlabs-apache/pull/2469 which I used it to fix scriptaliases spec fixes. 

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -646,7 +646,6 @@ class apache::mod::passenger (
   # - $passenger_sticky_sessions : since 4.0.45.
   # - $passenger_sticky_sessions_cookie_name : since 4.0.45.
   # - $passenger_thread_count : since 4.0.0.
-  # - $passenger_use_global_queue : since 2.0.4.Deprecated in 4.0.0.
   # - $passenger_user : since 4.0.0.
   # - $passenger_user_switching : since 3.0.0.
 
--- a/manifests/mod/passenger.pp
+++ b/manifests/mod/passenger.pp
@@ -1,5 +1,6 @@
 # @summary
 #   Installs `mod_pasenger`.
+#   > **Note**: This module support Passenger 4.0.0 and higher.
 #
 # @param manage_repo
 #   Toggle whether to manage yum repo if on a RedHat node.
@@ -253,35 +254,8 @@
 # @param passenger_user_switching
 #   Toggles whether to attempt to enable user account sandboxing, also known as user switching.
 #
-# @param rack_auto_detect
-#   This option has been removed in version 4.0.0 as part of an optimization. You should use PassengerEnabled instead.
-#
-# @param rack_autodetect
-#   This option has been removed in version 4.0.0 as part of an optimization. You should use PassengerEnabled instead.
-#
-# @param rack_base_uri
-#   Deprecated in 3.0.0 in favor of PassengerBaseURI.
-#
 # @param rack_env
 #   Alias for PassengerAppEnv.
-#
-# @param rails_allow_mod_rewrite
-#   This option doesn't do anything anymore since version 4.0.0.
-#
-# @param rails_app_spawner_idle_time
-#   This option has been removed in version 4.0.0, and replaced with PassengerMaxPreloaderIdleTime.
-#
-# @param rails_auto_detect
-#   This option has been removed in version 4.0.0 as part of an optimization. You should use PassengerEnabled instead.
-#
-# @param rails_autodetect
-#   This option has been removed in version 4.0.0 as part of an optimization. You should use PassengerEnabled instead.
-#
-# @param rails_base_uri
-#   Deprecated in 3.0.0 in favor of PassengerBaseURI.
-#
-# @param rails_default_user
-#   Deprecated in 3.0.0 in favor of PassengerDefaultUser
 #
 # @param rails_env
 #   Alias for PassengerAppEnv.
@@ -289,18 +263,6 @@
 # @param rails_framework_spawner_idle_time
 #   This option is no longer available in version 4.0.0. There is no alternative because framework spawning has been removed 
 #   altogether. You should use smart spawning instead.
-#
-# @param rails_ruby
-#   Deprecated in 3.0.0 in favor of PassengerRuby.
-#
-# @param rails_spawn_method
-#   Deprecated in 3.0.0 in favor of PassengerSpawnMethod.
-#
-# @param rails_user_switching
-#   Deprecated in 3.0.0 in favor of PassengerUserSwitching.
-#
-# @param wsgi_auto_detect
-#   This option has been removed in version 4.0.0 as part of an optimization. You should use PassengerEnabled instead.
 #
 # @note 
 #   In Passenger source code you can strip out what are all the available options by looking in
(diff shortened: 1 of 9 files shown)


HISTORY.md
```diff
@@ -1,3 +1,36 @@
+## 11.0.0
```
ekohl, line 1: This is not the right way. HISTORY.md is taken and then any newer version is added, so you'll wipe all the history between 11.0.0 and 3.2.0 this way. See https://github.com/github-changelog-generator/github-changelog-generator#migrating-from-a-manual-changelog for that.

What you're looking for is https://github.com/github-changelog-generator/github-changelog-generator#using-the-summary-section-feature to add some manual notes to a version.

manifests/mod/passenger.pp
```diff
@@ -1,5 +1,7 @@
 # @summary
 #   Installs `mod_pasenger`.
+#   > **Note**: This module does not install the passenger gem. You must install the gem yourself.
+#   > **Note**: This module support Passenger 5.0.0 and higher.
```
ekohl, line 4: EPEL 7 includes Passenger 4.0.53 now (https://src.fedoraproject.org/rpms/passenger).

manifests/mod/passenger.pp
```diff
@@ -750,14 +742,6 @@
...
-    if $passenger_use_global_queue {
-      if (versioncmp($passenger_installed_version, '4.0.0') > 0) {
-        fail('REMOVED PASSENGER OPTION :: passenger_use_global_queue :: -- no message on the current passenger reference webpage -- ')
-      }
-      if (versioncmp($passenger_installed_version, '2.0.4') < 0) {
-        fail("Passenger config option :: passenger_use_global_queue is not introduced until version 2.0.4 :: ${passenger_installed_version} is the version reported")
-      }
-    }
     if $passenger_user {
       if (versioncmp($passenger_installed_version, '4.0.0') < 0) {
```
ekohl, line 746: If you're raising the minimum level to 4.x you can drop this.

manifests/mod/passenger.pp
Ramesh7: The current version of Passenger is 6, thats where I am thinking to keep it N-1 and N. 
I know I am going aggressive but open for thoughts, do you propose to keep it 4.0.0 as min version to support?

manifests/mod/passenger.pp
Ramesh7: yeah, you are right, we should be cleaning up all < 5.0.0 based on the my approach as of now, coz we are expecting these are params coming from some older versions.

manifests/mod/passenger.pp
Ramesh7: Have updated with 4.0.0 as minimum version, open for thoughts.

ekohl: I'd indeed be a bit more conservative. Not sure how many users use it, but I tend to prefer distro packages. You can drop it next year when EPEL7 goes EOL.

Ramesh7: @ekohl Not sure if you got chance to look above comment but I have updated min passenger version to 4.0.0, I think that should be good?
Also the PR is ready for review, can you please review. Thanks

Ramesh7: Once everything looks good, I will release [11.0.0](https://github.com/puppetlabs/puppetlabs-apache/pull/2464), if we feel not good then will merge https://github.com/puppetlabs/puppetlabs-apache/pull/2469 with 11.0.0.

bastelfreak: I don't know the Puppet support policy (is there any?), but I would also be fine with only supporting what's in RHEL8/RHEL9. RHEL8 was released quite some time ago so one could argue that it's fine to Drop EL7. I'm fine with both.

bastelfreak (approved):

smortex (commented):
There seems to be something weird with `passenger_resist_deployment_errors` as I do not think it was removed. It is an "Enterprise only" feature and I have not tested it, but the doc still reference it.

smortex:
Also wondering if having one PR for passenger and a separate one for scriptaliases would make the changelog more readable?

manifests/mod/passenger.pp
```diff
@@ -1,5 +1,7 @@
 # @summary
 #   Installs `mod_pasenger`.
+#   > **Note**: This module does not install the passenger gem/package. You must install it yourself.
```
ekohl, line 3: I'd assume that if you pass `mod_package` that it does.

manifests/mod/passenger.pp
```diff
@@ -329,6 +315,12 @@
   Optional[String] $mod_package                                                              = undef,
   Optional[String] $mod_package_ensure                                                       = undef,
   Optional[String] $mod_path                                                                 = undef,
+  Optional[Integer] $passenger_max_request_queue_time                                        = undef,
```
ekohl, line 318: Adding new parameters in the same PR as removing deprecations is a bad practice IMHO. Makes it hard to follow what's going on.

manifests/mod/passenger.pp
```diff
@@ -963,26 +731,20 @@
...
-  # - $rails_base_uri : since unkown. Deprecated in 3.0.0.
-  # - $rails_default_user : since unkown. Deprecated in 3.0.0.
   # - $rails_env : since 2.0.0.
   # - $rails_framework_spawner_idle_time : since unkown. Deprecated in 4.0.0.
-  # - $rails_ruby : since unkown. Deprecated in 3.0.0.
-  # - $rails_spawn_method : since unkown. Deprecated in 3.0.0.
-  # - $rails_user_switching : since unkown. Deprecated in 3.0.0.
-  # - $wsgi_auto_detect : since unkown. Deprecated in 4.0.0.
   # - $rails_autodetect : this options is only for backward compatiblity with older versions of this class
   # - $rack_autodetect : this options is only for backward compatiblity with older versions of this class
```
ekohl, line 738: Does that mean it can be removed?

manifests/mod/passenger.pp
```diff
@@ -575,121 +485,40 @@
...
-    }
-    if $passenger_python {
-      if (versioncmp($passenger_installed_version, '4.0.0') < 0) {
-        fail("Passenger config option :: passenger_python is not introduced until version 4.0.0 :: ${passenger_installed_version} is the version reported")
-      }
-    }
     if $passenger_resist_deployment_errors {
       if (versioncmp($passenger_installed_version, '3.0.0') < 0) {
         fail("Passenger config option :: passenger_resist_deployment_errors is not introduced until version 3.0.0 :: ${passenger_installed_version} is the version reported")
       }
```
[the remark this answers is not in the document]
Ramesh7: We have checked this config and its available https://www.phusionpassenger.com/library/config/apache/reference/#passengerresistdeploymenterrors
Hope I am looking the right doc :)

templates/vhost/_scriptalias.erb
```diff
@@ -1,14 +1,2 @@
...
-<%- aliases.each do |salias| -%>
-  <%- if salias["path"] != '' -%>
-    <%- if salias["alias"] and salias["alias"] != '' -%>
-  ScriptAlias <%= salias['alias'] %> "<%= salias['path'] %>"
-    <%- elsif salias["aliasmatch"] and salias["aliasmatch"] != '' -%>
-  ScriptAliasMatch <%= salias['aliasmatch'] %> "<%= salias['path'] %>"
-    <%- end -%>
-  <%- end -%>
-<%- end -%>
+## Script alias directives
```
ekohl, line 1:
```suggestion
  ## Script alias directives
```

ekohl (changes requested):

manifests/mod/passenger.pp
Ramesh7: you are right, its does. Looks I missed that part totally. Thanks for catching this.

manifests/mod/passenger.pp
Ramesh7: sure, will put another PR for new params.

manifests/mod/passenger.pp
```diff
@@ -575,121 +485,40 @@
...
-    }
-    if $passenger_resolve_symlinks_in_document_root {
-      if (versioncmp($passenger_installed_version, '3.0.0') < 0) {
-        fail("Passenger config option :: passenger_resolve_symlinks_in_document_root is not introduced until version 3.0.0 :: ${passenger_installed_version} is the version reported")
+
+      if (versioncmp($passenger_installed_version, '5.2.0') > 0) {
+        fail('REMOVED PASSENGER OPTION :: passenger_resist_deployment_errors :: -- no message on the current passenger reference webpage -- ')
```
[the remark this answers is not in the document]
Ramesh7: I am keeping this untouched for now, will be covering while adding new params.
I have also removed newly added params and will be getting in new PR.

manifests/mod/passenger.pp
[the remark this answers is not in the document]
smortex: I meant the minimum version check can be removed: since we drop support for Passenger < 4, checking if we are using 3 or more recent is now useless.

manifests/mod/passenger.pp
Ramesh7: initially I kept it for reference but yeah it can be removed..

Ramesh7: done

templates/vhost/_scriptalias.erb
Ramesh7: done

Ramesh7:
> Also wondering if having one PR for passenger and a separate one for scriptaliases would make the changelog more readable?

Have removed new addition of params as part of this PR, just removed deprecated params.

manifests/mod/passenger.pp
```diff
@@ -963,22 +649,7 @@
   # - $passenger_use_global_queue : since 2.0.4.Deprecated in 4.0.0.
```
ekohl, line 649: If it's deprecated, should it be removed?

ekohl (approved):
Minor note, but looks good.

manifests/mod/passenger.pp
Ramesh7: done! Thanks @ekohl for your quick responses and review.

praj1001:
LGTM

praj1001 (dismissed):

praj1001:
LGTM

praj1001 (approved):

PR #2476, opened 2023-09-26
CAT-1347 erb rebase

dinesh-gujar:


Ramesh7:
don't think so we need this, hence closing the same.

PR #2481, opened 2023-10-10
Drop EoL Ubuntu code

bastelfreak:
This deletes Ubuntu 16 and older code. Those versions are already gone from metadata.json.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/spec/classes/mod/php_spec.rb
+++ b/spec/classes/mod/php_spec.rb
@@ -290,9 +290,8 @@ describe 'apache::mod::php', type: :class do
       end
 
       # all the following tests are for legacy php/apache versions. They don't work on modern ubuntu and redhat 8
-      next if (facts[:os]['release']['major'].to_i > 15 && facts[:os]['name'] == 'Ubuntu') ||
-              (facts[:os]['release']['major'].to_i >= 15 && facts[:os]['name'] == 'SLES')  ||
-              (facts[:os]['name'] == 'Debian')                                             ||
+      next if (facts[:os]['release']['major'].to_i >= 15 && facts[:os]['name'] == 'SLES')  ||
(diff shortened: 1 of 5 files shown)


ekohl (commented):
These conditions can be simplified:
```
spec/classes/mod/php_spec.rb
293:      next if (facts[:os]['release']['major'].to_i > 15 && facts[:os]['name'] == 'Ubuntu') ||

spec/defines/vhost_spec.rb
596:          if os_facts[:os]['release']['major'].to_i >= 18 && os_facts[:os]['name'] == 'Ubuntu'
```

spec/classes/mod/php_spec.rb
```diff
@@ -290,9 +290,8 @@
       end
 
       # all the following tests are for legacy php/apache versions. They don't work on modern ubuntu and redhat 8
-      next if (facts[:os]['release']['major'].to_i > 15 && facts[:os]['name'] == 'Ubuntu') ||
```
ekohl, line 293: I think this should be simplified to still test for Ubuntu, just like Debian.

bastelfreak: The tests were skipped for those and I updated the code below to match the Debian OS family. The logic should be the same now.

ekohl: Ah yes, I didn't read it carefully enough. Feels like we can also apply the same tricky to the Red Hat OS family below, but that can be done in another PR. Overall I think using `next` to skip tests below something is a poor practice. Indenting is much more reliable IMHO.

ekohl (approved):

PR #2483, opened 2023-10-11
Update load.epp to avoid loading a module already loaded

mz4r:
I encountered a problem during an Apache update. The scriplet executed during updates set up the default configuration of module loads. If the node with puppet agent no longer has the puppet configuration active then there is a duplicate problem in the loading of modules, example of warning returned: 
[so:warn] [pid 15619] AH01574: module dav_module is already loaded, skipping

The commit therefore makes it possible to load the module if it is not already loaded and thus avoid warnings on duplicates.

### DIFF
--- a/templates/mod/load.epp
+++ b/templates/mod/load.epp
@@ -4,4 +4,6 @@ LoadFile <%= $loadfile %>
 <% } -%>
 
 <% } -%>
+<IfModule <%= $_id %> >
 LoadModule <%= $_id %> <%= $_path %>
+</IfModule>
--- a/templates/mod/load.epp
+++ b/templates/mod/load.epp
@@ -4,6 +4,6 @@ LoadFile <%= $loadfile %>
 <% } -%>
 
 <% } -%>
-<IfModule <%= $_id %> >
+<IfModule !<%= $_id %> >
 LoadModule <%= $_id %> <%= $_path %>
 </IfModule>


ekohl (commented):
Can you share which OS + version you ran into this? I'd guess some EL based system since RPM can't deal with removed files (unlike DPKG for Debian), but then it would be useful to know which files. I think it's better to lay out files in a way that they overwrite native file locations instead of removing some file and creating it elsewhere.

mz4r:
> Can you share which OS + version you ran into this? I'd guess some EL based system since RPM can't deal with removed files (unlike DPKG for Debian), but then it would be useful to know which files. I think it's better to lay out files in a way that they overwrite native file locations instead of removing some file and creating it elsewhere.

OS : Oracle Linux 8 => 5.15.0-104.119.4.2.el8uek.x86_64

Examples files put by scriplet : 
01-cgi.conf
00-base.conf
00-dav.conf
00-lua.conf

Apache module configure "loadModule" in <module_name>.load ( example "ssl.load"), but apache scriplet set in "00-ssl.conf".
File list set by apache scriplet depend that installed modules.

"I think it's better to lay out files in a way that they overwrite native file locations instead of removing some file and creating it elsewhere."

Of course, i totaly agree, but the modules management is different. It would be necessary to modify the entire module management in Apache puppet module .

Scriplet apache can put many "LoadModule" in one file.
example /etc/httpd/conf.modules.d/00-proxy.conf :
LoadModule proxy_module modules/mod_proxy.so
LoadModule lbmethod_bybusyness_module modules/mod_lbmethod_bybusyness.so
LoadModule lbmethod_byrequests_module modules/mod_lbmethod_byrequests.so
LoadModule lbmethod_bytraffic_module modules/mod_lbmethod_bytraffic.so
LoadModule lbmethod_heartbeat_module modules/mod_lbmethod_heartbeat.so
LoadModule proxy_ajp_module modules/mod_proxy_ajp.so
LoadModule proxy_balancer_module modules/mod_proxy_balancer.so
LoadModule proxy_connect_module modules/mod_proxy_connect.so
LoadModule proxy_express_module modules/mod_proxy_express.so
LoadModule proxy_fcgi_module modules/mod_proxy_fcgi.so
LoadModule proxy_fdpass_module modules/mod_proxy_fdpass.so
LoadModule proxy_ftp_module modules/mod_proxy_ftp.so
LoadModule proxy_http_module modules/mod_proxy_http.so
LoadModule proxy_hcheck_module modules/mod_proxy_hcheck.so
LoadModule proxy_scgi_module modules/mod_proxy_scgi.so
LoadModule proxy_uwsgi_module modules/mod_proxy_uwsgi.so
LoadModule proxy_wstunnel_module modules/mod_proxy_wstunnel.so

But Apache module set in different file (proxy_ajp.load,proxy_balancer.load,proxy_connect.load ...)

Other one, the file 00-proxyhtml.conf : 
LoadModule xml2enc_module modules/mod_xml2enc.so
LoadModule proxy_html_module modules/mod_proxy_html.so

Apache module set then in xml2enc.load and proxy_html.load

mz4r:
Yes, so the easiest way to avoid this problem is to:
- Check if configuration files for modules are not missing with: rpm -qV httpd
- If the "purge_configs" variable is true, we create a file with empty content to avoid the duplicate problem and the implementation of default configuration not managed by the Apache module

So we avoid restarting Apache service multiple times.

If you agree, I can do the patch on my side and reopen a pull request.

Tell me if it suits you.

jst-cyr:
Hello @mz4r : The last comment from you on this suggested you were going to open a new PR (or update this PR) but I don't see a reference to another PR or further commits. Can we close this PR off?

PR #2485, opened 2023-10-17
fix wrong variable definition in params.pp for SUSE/openSUSE

carraroj:
## Summary
* changed the variable $vhost_dir and the $vhost_enable_dir
 
- [ ] Manually` verified. (For example `puppet apply`)


ekohl (changes requested):
A better description of this change would be nice. I have no idea what issue 7 is.

carraroj:
so i  changed the description, thing is the on installed package apache2 for SUSE Distros there is no directory `sites-enabled or sites-available or a file ports.conf , changed these parameters an my puppet apply with default class apache was running without errors and the directories are correct now.

PR #2499, opened 2023-11-22
pdksync - [CAT-1606] : Adding release_prep gem group

rajat-puppet:
[CAT-1606] : Adding release_prep gem group
pdk version: `3.0.0`


rajat-puppet:
Not required anymore.

PR #2501, opened 2023-11-28
(CAT-1608) - PDK update

Ramesh7:
## Summary

PDK update.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


spec/spec_helper.rb
```diff
@@ -25,7 +25,7 @@
   next unless File.exist?(f) && File.readable?(f) && File.size?(f)
 
   begin
-    default_facts.deep_merge!(YAML.safe_load(File.read(f), permitted_classes: [], permitted_symbols: [], aliases: true))
+    default_facts.merge!(YAML.safe_load(File.read(f), permitted_classes: [], permitted_symbols: [], aliases: true))
```
smortex, line 28: This may not be better.  There are some pitfalls with (deep)merging listed here:

https://github.com/voxpupuli/voxpupuli-test/commit/87d851ca58915117df3a06ff73810bb929dfc8b8

Ramesh7: Was just going through the existing code `deep_merge` and the change `merge`, looks like `deep_merge` never got succeded as I can see warning in [logs](https://github.com/puppetlabs/puppetlabs-apache/actions/runs/7040218061/job/19160763323#step:6:20), but now introducing to `merge` it started failing some of specs with `apache` and `mysql` modules : 
```
WARNING: Unable to load /home/runner/work/puppetlabs-apache/puppetlabs-apache/spec/default_facts.yml: undefined method `deep_merge!' for {:puppetversion=>"7.27.0", :facterversion=>"4.5.2"}:Hash
```

smortex: I think maybe @ekohl can elaborate because he worked on the fix linked above and told me a few times that I should not use `merge()` and use the `override_facts()` function instead.  This part of the patch has some info about the problem deep_merge pose:

https://github.com/voxpupuli/voxpupuli-test/commit/87d851ca58915117df3a06ff73810bb929dfc8b8#diff-69103b11f5a59727fe6d9de29156f298586f2a3ab7ff8758b356f691549cb07bR3-R8

Ramesh7: Anyways I think in this spec_helper we are trying to set some default facts which seems to be good, with override_facts make sense in actual spec tests where we already have implemented at multiple places.

rajat-puppet (approved):

smortex (approved):

PR #2504, opened 2023-12-12
Enable support for svn_parent_path in vhost directories

deepak7340:
## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -236,6 +236,9 @@
     <%- end -%>
     <%- if directory['dav'] -%>
     Dav <%= directory['dav'] %>
+    <%- if directory['svn_parent_path'] -%>
+    SVNParentPath <%= directory['svn_parent_path'] %>
+    <%- end -%>
     <%- if directory['dav_depth_infinity'] -%>
     DavDepthInfinity <%= scope.call_function('apache::bool2httpd', [directory['dav_depth_infinity']]) %>
     <%- end -%>


templates/vhost/_directories.erb
```diff
@@ -235,7 +235,10 @@
       <%- end -%>
     <%- end -%>
     <%- if directory['dav'] -%>
-    Dav <%= directory['dav'] %>
+    DAV <%= directory['dav'] %>
```
bastelfreak, line 238: Did you switch from `Dav` to `DAV` on purpose?

deepak7340: Wanted to enable the DAV svn in the location with SVNParentPath

deepak7340: Seems like its working on 'Dav svn'  too. Have removed that changes

smortex:
Can you please adjust the test suite so that this is tested?  These hashes are currently free-form which allow you to add a new item with minimal effort but at the cost of being super fragile because a typo would not be catch…

At some point, we will probably add data types to constraint what can be passed, but without tests this will make this needed step more stressful.

Thanks!

jst-cyr:
@deepak7340 : Do you still need this change? I've been reviewing old PRs and it looks like a request was made for tests to be added and no further changes were made.

PR #2512, opened 2024-01-02
Remove deprecated $proxy_ips and $trusted_proxy_ips parameters

ekohl:
Also simplifies the template by avoiding `Optional[Array] = undef` and instead treating empty arrays in that way.


smortex (approved):
LGTM

PR #2514, opened 2024-01-02
Fix extra newline at end of headers

smortex:
This slept through while converting templates form erb to epp.  Remove
it so that it is easier to spot regressions if the catalog is not
supposed to change but applying it in noop mode wants to change the
actual files.

### DIFF
--- a/templates/vhost/_file_header.epp
+++ b/templates/vhost/_file_header.epp
@@ -40,4 +40,4 @@ MDomain <%= $servername %>
 <% } -%>
 <% if $limitreqbody { -%>
   LimitRequestBody <%= $limitreqbody %>
-<% } %>
+<% } -%>


bastelfreak (approved):

smortex:
This issue was reported by @vchepkov on slack, I would like to see it released in a 12.0.1 version before we merge backward-incompatble change if you don't mind @ekohl?

vchepkov:
I concur, this fixes it for me

ekohl (approved):

ekohl:
> This issue was reported by @vchepkov on slack, I would like to see it released in a 12.0.1 version before we merge backward-incompatble change if you don't mind @ekohl?

I'm wondering if it's wise to do a major release so soon after. On the other hand, people can bump 2 releases at once

smortex:
@ekohl I personally do not really care about how ofter major release happen… they happen when they happen :shrug:

We can also have breaking changes labelled as such that wait not merged, and when we feel a new major release make sense merge them.  I think it is something we try to do with some voxpupuli modules (e.g. merge a PR that deprecate something and open a PR that remove the deprecation that is ready to merge).

Do you think we should trigger a 12.0.1 with this fix, or do we merge these backward-incompatible PRs and go far a 13.0.0?

ekohl:
12.0.1 now sounds good. Major versions are a pain in reverse dependencies because they require a release. In the Foreman context it means I need to do that in 3 modules. 3 commits to bump and 3 module releases

PR #2518, opened 2024-01-06
Correct handling of $serveraliases as string

### LINKED ISSUE #2517: serveraliases template creation fails on String
## Describe the Bug
If parameter serveraliases for a vhost is defined as a string instead of an array, the resulting entries in the vhost config file consist of each character of that string.
If the string contains a wildcard it turns into a unwanted catchall vhost for some other vhosts.
The example below results in something like this:
\## Server aliases
 ServerAlias *
 ServerAlias .
 ServerAlias s
 ServerAlias t
 ServerAlias r
 ServerAlias i
 ServerAlias n
 ServerAlias g
 ServerAlias .
 ServerAlias c
 ServerAlias o
 ServerAlias m

This bug was introduced with 12.0.0 and worked as expected until 11.1.0

## Expected Behavior
the config files should remain untouched after an upgrade from 11.1.0 to 12.0.0 or 12.0.1
in this case:
\## Server aliases
 ServerAlias *.string.com

## Steps to Reproduce
create a vhost like this:
 apache::vhost { 'string.com':
    servername        => 'string.com',
    serveraliases     => '*.string.com',
 }

run puppet on host, look in the resulting conf file.

## Workaround
change parameter to array:
    serveraliases     => ['*.string.com'],

## Environment
 - Version 12.0.0 / 12.0.1
 - Puppet Version 7.27.0
 - Platform OracleLinux 7

## Additional Context
refs:
https://forge.puppet.com/modules/puppetlabs/apache/reference#-apache--vhost--serveraliases
serveraliases
Data type: Variant[Array[String], String]
Sets the ServerAliases of the site.
Default value: []

ekohl:
Fixes #2517

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2559,7 +2559,7 @@ define apache::vhost (
     concat::fragment { "${name}-serveralias":
       target  => "${priority_real}${filename}.conf",
       order   => 210,
-      content => epp('apache/vhost/_serveralias.epp', { 'serveraliases' => $serveraliases }),
+      content => epp('apache/vhost/_serveralias.epp', { 'serveraliases' => [$serveraliases].flatten }),
     }
   }
 
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -776,7 +776,7 @@ describe 'apache::vhost', type: :define do
           }
 
           it { is_expected.to contain_concat__fragment('rspec.example.com-scriptalias') }
-          it { is_expected.to contain_concat__fragment('rspec.example.com-serveralias') }
+          it { is_expected.to contain_concat__fragment('rspec.example.com-serveralias').with_content(%r{^  ServerAlias test-example\.com$}) }
 
           it {
             expect(subject).to contain_concat__fragment('rspec.example.com-setenv')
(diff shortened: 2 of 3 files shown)


manifests/vhost.pp
```diff
@@ -2559,7 +2559,7 @@
     concat::fragment { "${name}-serveralias":
       target  => "${priority_real}${filename}.conf",
       order   => 210,
-      content => epp('apache/vhost/_serveralias.epp', { 'serveraliases' => $serveraliases }),
+      content => epp('apache/vhost/_serveralias.epp', { 'serveraliases' => Array($serveraliases) }),
```
smortex, line 2562: This does not fix the problem as `Array('foo') #=> ['f', 'o', 'o']`.  Did you meant:

```suggestion
      content => epp('apache/vhost/_serveralias.epp', { 'serveraliases' => [$serveraliases].flatten }),
```

spec/defines/vhost_spec.rb
```diff
@@ -1221,6 +1221,32 @@
           it { is_expected.not_to contain_concat__fragment('NameVirtualHost 127.0.0.1:8080') }
         end
 
+        describe 'serveraliases parameter' do
+          let :params do
+            {
+              'serveraliases' => serveraliases,
+            }
+          end
```
smortex, line 1229:
```suggestion
          let :params do
            {
              'docroot'       => '/path/to/docroot',
              'serveraliases' => serveraliases,
            }
          end
```

should help according to the failures bellow.

smortex (commented):
+1 for adding tests for this parameter, need some tweaking and should be good to go!

manifests/vhost.pp
ekohl: I was debating that. Perhaps I was confused by Ruby's `Array()` behavior, which may also be where the bug came from.

This is why I generally prefer to only allow arrays when you want an array value.

ekohl:
Test failure is unrelated.

smortex (approved):
LGTM!

rwaffen (approved):

smortex:
This feel safe so I will merge it.  Do you want to trigger a 12.0.2 release?

ekohl:
I can't trigger it from my phone

bastelfreak:
release PR: https://github.com/puppetlabs/puppetlabs-apache/pull/2519

PR #2521, opened 2024-01-12
Added cache_disk

dploeger:
## Summary
This deprecates apache::mod::disk_cache and instead adds support for apache::mod::cache_disk. Also adds parameters to both apache::mod::cache_disk and apache::mod::cache.

## Additional Context
Since Apache 2.4 mod_disk_cache is named mod_cache_disk. This officially supports mod_cache_disk and additionally adds configuration options for it. To be backwards compatible, the old module is still available and simply uses the new module.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/manifests/mod/disk_cache.pp
+++ b/manifests/mod/disk_cache.pp
@@ -28,15 +28,10 @@ class apache::mod::disk_cache (
 ) {
   deprecation('apache::mod::disk_cache', 'This class is deprecated; please use apache::mod::cache_disk')
 
-  $_cache_ignore_headers = $cache_ignore_headers ? {
-    undef   => [],
-    default => $cache_ignore_headers.split(' ')
-  }
-
   class { 'apache::mod::cache_disk':
     cache_root              => $cache_root,
     cache_enable            => ['/'],
-    cache_ignore_headers    => $_cache_ignore_headers,
+    cache_ignore_headers    => $cache_ignore_headers,
     cache_dir_length        => 1,
     cache_dir_levels        => 2,
     configuration_file_name => 'cache_disk.conf'
--- a/templates/mod/cache.conf.epp
+++ b/templates/mod/cache.conf.epp
@@ -25,6 +25,6 @@ CacheLock <%= $cache_lock %>
 <%- if $cache_ignore_cache_control { -%>
 CacheIgnoreCacheControl <%= $cache_ignore_cache_control %>
 <%- } -%>
-<%- if $cache_ignore_headers { -%>
-CacheIgnoreHeaders <%= $cache_ignore_headers.join(' ') %>
+<%- if ! empty($cache_ignore_headers) { -%>
+CacheIgnoreHeaders <%= $cache_ignore_headers.sort.join(' ') %>
 <%- } -%>
--- a/spec/classes/mod/cache_spec.rb
+++ b/spec/classes/mod/cache_spec.rb
@@ -18,13 +18,26 @@ describe 'apache::mod::cache', type: :class do
     describe 'with cache_ignore_headers' do
       let(:params) do
         {
-          cache_ignore_headers: 'Set-Cookie',
+          cache_ignore_headers: ['Set-Cookie'],
         }
       end
 
       it {
         expect(subject).to contain_file('cache.conf')
           .with(content: %r{CacheIgnoreHeaders Set-Cookie})
+      }
+    end
+
+    describe 'with cache_ignore_headers' do
+      let(:params) do
+        {
+          cache_ignore_headers: ['Set-Cookie', 'X-Forwarded-For', 'Cross-Origin-Embedder-Policy', 'Expires', 'Access-Control-Allow-Headers'],
+        }
+      end
+
+      it {
+        expect(subject).to contain_file('cache.conf')
+          .with(content: %r{CacheIgnoreHeaders Access-Control-Allow-Headers Cross-Origin-Embedder-Policy Expires Set-Cookie X-Forwarded-For})
       }
     end
 
--- a/spec/classes/mod/disk_cache_spec.rb
+++ b/spec/classes/mod/disk_cache_spec.rb
@@ -14,7 +14,7 @@ describe 'apache::mod::disk_cache', type: :class do
 
     let :pre_condition do
       'class{ "apache":
-        default_mods   => ["cache"],
+        default_mods   => ["cache", "disk_cache"],
         mod_dir        => "/tmp/junk",
        }'
     end
--- a/templates/mod/cache_disk.conf.epp
+++ b/templates/mod/cache_disk.conf.epp
@@ -1,9 +1,10 @@
 <% |
-  Optional[String] $cache_root          = undef,
-  Array[String] $cache_enable           = [],
-  Optional[Integer] $cache_dir_length   = undef,
-  Optional[Integer] $cache_dir_levels   = undef,
-  Optional[Integer] $cache_max_filesize = undef,
+  Optional[String] $cache_root            = undef,
+  Array[String] $cache_enable             = [],
+  Optional[Integer] $cache_dir_length     = undef,
+  Optional[Integer] $cache_dir_levels     = undef,
+  Optional[Integer] $cache_max_filesize   = undef,
+  Optional[String] $cache_ignore_headers  = undef,
 | -%>
 <%- if $cache_enable { -%>
   <%- $cache_enable.each |$enable| { -%>
@@ -19,4 +20,7 @@ CacheDirLength <%= $cache_dir_length %>
 <%- } -%>
 <%- if $cache_max_filesize { -%>
 CacheMaxFileSize <%= $cache_max_filesize %>
+<%- } -%>
+<%- if $cache_ignore_headers { -%>
+CacheIgnoreHeaders <%= $cache_ignore_headers -%>
 <%- } -%>


manifests/mod/cache_disk.pp
```diff
@@ -0,0 +1,121 @@
...
+  # - $cache_ignore_no_lastmod
+  # - $cache_header
+  # - $cache_lock
+  # - $cache_ignore_cache_control
+  # - $cache_max_filesize
+  file { $_configuration_file_name:
+    ensure  => file,
+    path    => "${apache::mod_dir}/${_configuration_file_name}",
+    mode    => $apache::file_mode,
+    content => template('apache/mod/cache_disk.conf.erb'),
```
ekohl, line 116: Please use EPP for new templates.

manifests/mod/cache_disk.pp
```diff
@@ -0,0 +1,121 @@
...
+  Optional[Enum['Off', 'On']] $cache_ignore_cache_control = undef,
+  Optional[Integer] $cache_max_filesize                   = undef,
+) {
+  include apache
+  if $cache_root {
+    $_cache_root = $cache_root
+  }
+  elsif versioncmp($apache::apache_version, '2.4') >= 0 {
+    $_module_name = 'cache_disk'
+    $_cache_root = $::osfamily ? {
```
ekohl, line 77: Please use `$facts['os']['family']`

ekohl (changes requested):
We've dropped support for Apache < 2.4 so you can simplify a few things.

manifests/mod/cache_disk.pp
dploeger: Oh, you're quick. 😆  The PR is in draft. I'm currently optimizing and adding tests. Thanks!

manifests/mod/cache_disk.pp
dploeger: Oh, okay. I'll change that.

manifests/mod/cache_disk.pp
ekohl: I saw that and didn't look too closely at details, just wanted to point out some general things that jumped out and can affect writing tests.

manifests/mod/cache.pp
```diff
@@ -1,8 +1,58 @@
...
+#   Enable the thundering herd lock.
+#
+# @param cache_ignore_cache_control
+#   Ignore request to not serve cached content to client
+#
 # @see https://httpd.apache.org/docs/current/mod/mod_cache.html for additional documentation.
 #
-class apache::mod::cache {
+class apache::mod::cache (
+  Optional[String] $cache_ignore_headers                               = undef,
```
smortex, line 28: This allow multiple values, we should use:

```suggestion
  Array[String[1]] $cache_ignore_headers                               = [],
```

manifests/mod/cache_disk.pp
```diff
@@ -0,0 +1,91 @@
...
+# @param configuration_file_name
+#   DEPRECATED Name of module configuration file (used for the compatibility layer for disk_cache)
+#
+# @note
+#   Apache 2.2, mod_disk_cache installed. On Apache 2.4, mod_cache_disk installed.
+#
+# @see https://httpd.apache.org/docs/2.4/mod/mod_cache_disk.html for additional documentation on version 2.4.
+#
+class apache::mod::cache_disk (
+  Optional[String] $cache_root              = undef,
```
smortex, line 40: Are relative path acceptable?  If yes, a non-null `String` should be expected:

```suggestion
  Optional[String[1]] $cache_root              = undef,
```

otherwise:

```suggestion
  Optional[Stdlib::Absolutepath] $cache_root              = undef,
```

smortex (commented):
There are a few minor things that we can improve.  I added in-line comments. Thanks!

dploeger:
@smortex  Thanks. Done.

manifests/mod/disk_cache.pp
```diff
@@ -17,44 +16,24 @@
...
-    path    => "${apache::mod_dir}/disk_cache.conf",
-    mode    => $apache::file_mode,
-    content => epp('apache/mod/disk_cache.conf.epp', $parameters),
-    require => Exec["mkdir ${apache::mod_dir}"],
-    before  => File[$apache::mod_dir],
-    notify  => Class['apache::service'],
+  class { 'apache::mod::cache_disk':
+    cache_root              => $cache_root,
+    cache_enable            => ['/'],
+    cache_ignore_headers    => $cache_ignore_headers,
```
smortex, line 34: For backwards compatibility we may need something like:

```suggestion
    cache_ignore_headers    => $cache_ignore_headers.split(' '),
```

(I don't think it's worth breaking backwards compatibility to improve a legacy interface)

spec/classes/mod/cache_disk_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    it { is_expected.to contain_class('apache::mod::cache_disk') }
+    it { is_expected.to contain_class('apache::mod::cache').that_comes_before('Class[Apache::Mod::Cache_disk]') }
+    it { is_expected.to contain_apache__mod('cache_disk') }
+
+    default_config = %r{CacheEnable disk /\nCacheRoot "/var/cache/apache2/mod_cache_disk"}
+
+    it {
+      expect(subject).to contain_file('cache_disk.conf')
+                           .with(content: default_config)
+    }
```
smortex, line 32: Not sure about the suggestion from rubocop :upside_down_face: 

The line length feel reasonable with current editors and screen so would not mind seeing everything on the same line:

```suggestion
    it { is_expected.to contain_file('cache_disk.conf').with(content: default_config) }
```


Note that I replaced `expect(subject)` by `is_expected`: it does the same but makes the whole test [almost] readable as a sentence.

dploeger:
@smortex Should I take a look at the other Rubocop findings?

smortex:
> @smortex Should I take a look at the other Rubocop findings?

Yes please :smile:  They are admittedly mostly gratuitous, but allow some consistency across the module which makes maintenance easier.  Rubocop has to pass for the unit and integration tests to run.

As it is now, it looks quite good, CI will maybe help us spot some remaining issues.  I am just wondering if the unit tests for the legacy class should be updated to check that it instantiate the new class with the right parameters and not more.

dploeger:
Fixed Rubocop

manifests/mod/cache_disk.pp
```diff
@@ -0,0 +1,91 @@
...
+#   This will install an configure the proper module depending on the used apache version, so
+#   - mod_cache_disk for apache version >= 2.4
+#   - mod_disk_cache for older apache versions
+# 
+# @param cache_root
+#   Defines the name of the directory on the disk to contain cache files.
+#   Default depends on the Apache version and operating system:
+#   - Debian: /var/cache/apache2/mod_cache_disk
+#   - FreeBSD: /var/cache/mod_cache_disk
+#   - Red Hat, Apache 2.4: /var/cache/httpd/proxy
```
ekohl, line 14: You can assume Apache 2.4 only and I don't see Apache 2.2 references in this class

manifests/mod/cache_disk.pp
```diff
@@ -0,0 +1,91 @@
...
+#   The maximum size (in bytes) of a document to be placed in the cache
+#
+# @param cache_ignore_headers
+#   DEPRECATED Ignore request to not serve cached content to client (included for compatibility reasons to support disk_cache)
+#
+# @param configuration_file_name
+#   DEPRECATED Name of module configuration file (used for the compatibility layer for disk_cache)
+#
+# @note
+#   Apache 2.2, mod_disk_cache installed. On Apache 2.4, mod_cache_disk installed.
```
ekohl, line 35: I don't see the code for Apache 2.2.

manifests/mod/cache_disk.pp
```diff
@@ -0,0 +1,91 @@
...
+#
+# @see https://httpd.apache.org/docs/2.4/mod/mod_cache_disk.html for additional documentation on version 2.4.
```
ekohl, line 37: I think this is sufficient
```suggestion
# @see https://httpd.apache.org/docs/2.4/mod/mod_cache_disk.html
```

manifests/mod/cache_disk.pp
```diff
@@ -0,0 +1,91 @@
...
+      'debian'  => '/var/cache/apache2/mod_cache_disk',
+      'redhat'  => '/var/cache/httpd/proxy',
+      'freebsd' => '/var/cache/mod_cache_disk',
+    },
+    default => $cache_root,
+  }
+  $_configuration_file_name = $configuration_file_name ? {
+    undef   => 'cache_disk.conf',
+    default => $configuration_file_name
+  }
```
ekohl, line 68: Isn't this what `pick()` does?
```suggestion
  $_configuration_file_name = pick($configuration_file_name, 'cache_disk.conf')
```

manifests/mod/cache_disk.pp
dploeger: Oh right, I missed that comment.

manifests/mod/cache_disk.pp
dploeger: True, but I wanted to rely on built in features instead of stdlib uses. Should I change it to pick?

ekohl: What's one more:
```console
$ rg pick
manifests/custom_config.pp
97:  $_file_mode = pick($file_mode, $apache::file_mode)

manifests/mod/userdir.pp
42:    'userdir'         => pick($userdir, $path),

manifests/mod/mime_magic.pp
13:  $_magic_file = pick($magic_file, "${apache::conf_dir}/magic")

manifests/mod/mime.pp
21:  $_mime_types_additional = pick($mime_types_additional, $apache::mime_types_additional)
```

dploeger: Okay. 🤷🏽

templates/mod/cache.conf.epp
```diff
@@ -0,0 +1,30 @@
...
+CacheHeader <%= $cache_header %>
+<%- } -%>
+<%- if $cache_lock { -%>
+CacheLock <%= $cache_lock %>
+<%- } -%>
+<%- if $cache_ignore_cache_control { -%>
+CacheIgnoreCacheControl <%= $cache_ignore_cache_control %>
+<%- } -%>
+<%- if $cache_ignore_headers { -%>
+CacheIgnoreHeaders <% $cache_ignore_headers.each |$header| { %><%= $header %> <% } %>
```
smortex, line 29: In this case (list of words), prefer `join` which avoid a trailing space after the last item and is a bit more readable:

```suggestion
CacheIgnoreHeaders <% $cache_ignore_headers.join(' ') %>
```

ThomasMinor: Nice. Added a missing '='.

manifests/mod/disk_cache.pp
```diff
@@ -17,44 +16,24 @@
...
-    path    => "${apache::mod_dir}/disk_cache.conf",
-    mode    => $apache::file_mode,
-    content => epp('apache/mod/disk_cache.conf.epp', $parameters),
-    require => Exec["mkdir ${apache::mod_dir}"],
-    before  => File[$apache::mod_dir],
-    notify  => Class['apache::service'],
+  class { 'apache::mod::cache_disk':
+    cache_root              => $cache_root,
+    cache_enable            => ['/'],
+    cache_ignore_headers    => $cache_ignore_headers.split(' '),
```
smortex, line 34: This cause trouble if `$cache_ignore_headers` is `undef`.  We can default to something like:

```suggestion
    cache_ignore_headers    => if $cache_ignore_headers {
                                 $cache_ignore_headers.split(' ')
                               } else {
                                 []
                               },
```

smortex (commented):
It also look like the `apache::mod::disk_cache` is included in `manifests/default_mods.pp` for FreeBSD.

I just checked on my FreeBSD box, and it seems the default config does not include the `cache_disk` module, so I think we can also remove these lines:

https://github.com/puppetlabs/puppetlabs-apache/blob/9f159df9c66807168c3ff495be2b5aeac4319b9b/manifests/default_mods.pp#L72-L73

(In fact, all these lists seems outdated, maybe some global rework would be great, but that is out-of-scope of this PR)

manifests/mod/disk_cache.pp
ThomasMinor: I actually prefer the selector variant here.

ThomasMinor: But actually the splitting is wrong here. I'll asap commit a fixed version.

ThomasMinor:
It seems the failures during acceptance testing do not relate to the changes in this PR. Any idea when these might get fixed? Can I offer some support here?

smortex:
Puppet 8 seems to add a check for duplicates that fail.  I opened #2525 to address the issue.

smortex:
@dploeger can you please rebase your changes on top of the main branch?  This should fix CI.

From your working directory:

```
git fetch origin       # Download the latest code we have here
git rebase origin/main # Move your commits on top of the main branch
git push -f            # Send the changes (-f is required because we re-wrote history)
```

ThomasMinor:
Hi @smortex, I just merged the latest changes from main to this repo. Since the 2 failing tests do not seem to be related to the changes in this PR, how do we proceed from here?

smortex:
@ThomasMinor From my point of view this look okay.  The repository require multiple approval before merging so the next move is probably to have some feedback from @puppetlabs maybe?

I re-approved the CI run (any change you do to your commit invalidate the test), I expect it to all pass except the 2 known-bad tests which I think we can ignore.

ThomasMinor:
Hi @smortex, it seems another test joined the club and failed with connection failures. Can I assist solving the problem?

ThomasMinor:
Seems I looked wrong, 2 fail, one test was skipped.

smortex:
@ThomasMinor the Ubuntu ARM tests are broken and mend seems to only run when the PR is from a branch in the repo itself.

`spec/classes/mod/disk_cache_spec.rb` is maybe testing too much things given the corresponding code is only a wrapper around `apache::mod::cache_disk` (it act as an integration test rather than a unit test).

Also there is a merge commit in the branch which is not nice. Rebasing on top of master would fix this.

These two annoyances are not stoppers for me.

ThomasMinor:
How about a final review, @ekohl & @bastelfreak?
Then we might be able to close this one.

ekohl (approved):

PR #2525, opened 2024-02-15
Fix `mod_auth_openidc` parameters

smortex:
A typo in the data type cause an exception with recent Puppet 8
versions:

```
Error: The key '["IDTokenEncryptedResponseAlg"]' is declared more than once (file: /etc/puppetlabs/code/environments/production/modules/apache/types/oidcsettings.pp, line: 41, column: 57) on node b4d45d8831d5.mtgrj3ika5nejlcpzgeeecg0la.cx.internal.cloudapp.net
```

According to the upstream repository configuration file, this is a
copy-paste error.


ekohl (approved):

ekohl:
It was introduced in 9f7f38ff21036c9a1ce4d669ccaea816941209ca. How did we not find this before? I see we don't have a test for the type in `spec/type_aliases` so would have caught it?

smortex:
Lack of tests might not help indeed.  The new Puppet 8 ensuring there is no duplicate allows us to find such issues now, so I guess this is okay.

The ARM builds seems currently broken (or at least not reliable as they do not start), and I think we can ignore them.  I will merge this so that we can move on on other PR broken by this typo.

PR #2531, opened 2024-02-27
bugfix: fix the vhost security configuration

JGodin-C2C:
The modified variable was never passed to the template, resulting in a bad configuration of the vhost security if secrule were removed.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2170,28 +2170,22 @@ define apache::vhost (
   }
 
   ## Create a global LocationMatch if locations aren't defined
-  if $modsec_disable_ids {
-    if $modsec_disable_ids =~ Array {
-      $_modsec_disable_ids = { '.*' => $modsec_disable_ids }
-    } else {
-      $_modsec_disable_ids = $modsec_disable_ids
-    }
+  if $modsec_disable_ids =~ Array {
+    $_modsec_disable_ids = { '.*' => $modsec_disable_ids }
+  } else {
+    $_modsec_disable_ids = $modsec_disable_ids
   }
 
-  if $modsec_disable_msgs {
-    if $modsec_disable_msgs =~ Array {
-      $_modsec_disable_msgs = { '.*' => $modsec_disable_msgs }
-    } else {
-      $_modsec_disable_msgs = $modsec_disable_msgs
-    }
+  if $modsec_disable_msgs =~ Array {
+    $_modsec_disable_msgs = { '.*' => $modsec_disable_msgs }
+  } else {
+    $_modsec_disable_msgs = $modsec_disable_msgs
   }
 
-  if $modsec_disable_tags {
-    if $modsec_disable_tags =~ Array {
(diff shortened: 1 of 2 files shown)


templates/vhost/_security.epp
```diff
@@ -1,6 +1,8 @@
 <IfModule mod_security2.c>
 <% if $modsec_disable_vhost {-%>
   SecRuleEngine Off
+<% } else { -%>
+  SecRuleEngine On
 <% } -%>
```
ekohl, line 6: I'm not a fan of double negatives, but I think this is better than renaming it now.
```suggestion
  SecRuleEngine <%= apache::bool2httpd(!$modsec_disable_vhost) %>
```

manifests/vhost.pp
```diff
@@ -2832,10 +2832,10 @@
     $security_params = {
       'modsec_disable_vhost'              => $modsec_disable_vhost,
       'modsec_audit_log_destination'      => $modsec_audit_log_destination,
-      '_modsec_disable_ids'               => $modsec_disable_ids,
+      '_modsec_disable_ids'               => $_modsec_disable_ids,
```
ekohl, line 2835: The EPP file shouldn't have underscores. What a poor API but I guess 31a137bcb490e6a9041959e8f7115dceac91d4c0 is to blame for that. Not something you need to fix now, but more of a general complaint.

templates/vhost/_security.epp
JGodin-C2C: Should i embark this change in my commit or using the github suggestion commit is enought ?

ekohl: I think in this repo there's no need to fix up your commits.

JGodin-C2C:
Ok, i will need to have a better understanding

JGodin-C2C:
Can we launch the failed tests ? seems a problem on the CI side ?

stevegreengvl:
Seems to also cause a problem if the modsecure_disable_ids parameter is an Array, as the Hash Array built never gets passed to the "epp" processing

JGodin-C2C:
@stevegreengvl I dont get it , 
The _modsec_disable_ids seems to be  passed to the Hash "security_params" and then injected into the epp. 
Is there anything i missed ?

stevegreengvl:
@JGodin-C2C we have this in our PUppet code
apache::vhost { 'XXX':
    ...
    modsec_disable_ids => [ array of IDs ],
    ...

Up to Version 11.1.0 of the Module all was fine in the generated vhost Config file, from Version 12.0 the Block below is missing from the generated file

<LocationMatch .*>
    SecRuleRemoveById XXXXXX
    SecRuleRemoveById YYYYYY
  </LocationMatch>

ekohl (approved):
I think this is correct, but it would be nice to have a test for this.

jst-cyr:
Hey there @JGodin-C2C, I'm going through old requests and am wondering if you still need this fix? If so, are you able to rebase and add some tests in? Thank you!

jst-cyr:
For now, I'm closing this PR until a need comes up for it in the future.

PR #2535, opened 2024-03-17
Add EnableSendfile per directory

kajinamit:
## Summary
Support configuring the EnableSendfile option under Directory directives

## Additional Context
The EnableSendfile option can be configured in a Directory directive to disable/enable the snedfile feature for specific paths as is described in https://httpd.apache.org/docs/2.4/en/mod/core.html#enablesendfile .

## Related Issues (if any)
N/A

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/templates/vhost/_directories.erb
+++ b/templates/vhost/_directories.erb
@@ -19,8 +19,11 @@
       <%- path = directory['path'] -%>
 
   <<%= provider %> "<%= path %>">
-      <%- if directory['headers'] -%>
-        <%- Array(directory['headers']).each do |header| -%>
+    <%- if directory['enable_sendfile'] -%>
+    EnableSendfile <%= directory['enable_sendfile'] %>
+    <%- end -%>
+    <%- if directory['headers'] -%>
+      <%- Array(directory['headers']).each do |header| -%>
     Header <%= header %>
       <%- end -%>
     <%- end -%>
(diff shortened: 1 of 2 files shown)


templates/vhost/_directories.erb
```diff
@@ -19,8 +19,11 @@
       <%- path = directory['path'] -%>
 
   <<%= provider %> "<%= path %>">
-      <%- if directory['headers'] -%>
-        <%- Array(directory['headers']).each do |header| -%>
+    <%- if directory['sendfile'] -%>
+    EnableSendfile <%= directory['sendfile'] %>
```
smortex, line 23:
```suggestion
    EnableSendfile <%= directory['enable_sendfile'] %>
```

smortex (commented):
That looks good, just using the actual configuration setting name for the key might make it easier to use it (we trend to do this in the rest of the module, with some exceptions unfortunately).

templates/vhost/_directories.erb
kajinamit: I selected `sendfile` because that name is used in the apache class, but I agree making it consistent with the actual option would be better. Fixed in the latest version.

smortex (approved):
Thank you for the update. LGTM!

ekohl (approved):

PR #2540, opened 2024-04-04
Add missing brackets for function call

gerlingsm:
## Summary
Two brackets have been added around the arguments of the second scope.call_function

## Additional Context
- [x] Root cause and the steps to reproduce. (If applicable)

The problem here is or was that the argument was not passed correctly to the function "authz_core_config" and you then got an error message from it that an incorrect variable type was passed.
"parameter 'config' expects a Hash value, got Tuple"
thanks to @hufschmidt for helping to find the error and solve the problem

For error reproduction it is sufficient to use "authz_core" in some way within a directory in a vhost.
For example as described in the ReadMe: https://github.com/puppetlabs/puppetlabs-apache/blob/b9ac34ca8a17c1bd507de01fc109d8fe73bf110c/manifests/vhost.pp#L1369-L1392

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/spec/acceptance/vhost_spec.rb
+++ b/spec/acceptance/vhost_spec.rb
@@ -452,6 +452,25 @@ describe 'apache::vhost define' do
               auth_require => 'valid-user',
               satisfy => 'Any',
             },
+            {
+              path => '/var/www/files/authz',
+              auth_type => 'Basic',
+              auth_name => 'Basic Auth',
+              authz_core => {
+                require_all => {
+                  require_any => {
+                    require => [
+                      '127.0.0.1'
+                      '10.10.10.10'
+                    ],
+                    require_all => {
+                      auth_user_file => ['/var/www/htpasswd'],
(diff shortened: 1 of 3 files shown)


bastelfreak:
@gerlingsm can you maybe add a unit test for this to ensure that this works now?

gerlingsm:
> @gerlingsm can you maybe add a unit test for this to ensure that this works now?

I will try tomorrow or early next week yes.
However, this is my first pull request for a Puppet project and I haven't dealt with unit tests before....

gerlingsm:
> > @gerlingsm can you maybe add a unit test for this to ensure that this works now?
> 
> I will try tomorrow or early next week yes. However, this is my first pull request for a Puppet project and I haven't dealt with unit tests before....

@bastelfreak Could you have a look? Is that sufficient?

PR #2542, opened 2024-04-20
Add support for `IndexHeadInsert` for directories

trefzer:
## Summary
add option IndexHeadInsert for directories

## Additional Context
Option was missing ;)

### DIFF
--- a/spec/defines/vhost_spec.rb
+++ b/spec/defines/vhost_spec.rb
@@ -150,7 +150,8 @@ describe 'apache::vhost', type: :define do
                   'directoryindex' => 'disabled',
                   'options' => ['Indexes', 'FollowSymLinks', 'MultiViews'],
                   'index_options' => ['FancyIndexing'],
-                  'index_style_sheet' => '/styles/style.css' },
+                  'index_style_sheet' => '/styles/style.css',
+                  'index_head_insert' => ['<base target=_blank>'] },
                 { 'path' => '/var/www/files/output_filtered',
                   'set_output_filter' => 'output_filter' },
                 { 'path' => '/var/www/files/input_filtered',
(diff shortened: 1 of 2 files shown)


smortex (approved):

smortex:
This option is also availabile in the global server config and virtual hosts.  Do you want to add the corresponding parameters to this PR?

trefzer:
@smortex there is no need to add this as a parameter to init.pp or vhost.pp since it is a value in the directory hash. (or do I missunderstand youre statement ?)

But another thing, that I realized, is that the options are only applied for provider == Directory (templates/vhost/_directories.erb line 33) which I think is wrong (at least the Index options seem to work for provider == Location on apache but not in puppet).

smortex:
> @smortex there is no need to add this as a parameter to init.pp or vhost.pp since it is a value in the directory hash. (or do I missunderstand youre statement ?)

I am referring to https://httpd.apache.org/docs/current/mod/mod_autoindex.html#indexheadinsert, more specifically:

> [Context:](https://httpd.apache.org/docs/current/mod/directive-dict.html#Context)	server config, virtual host, directory, .htaccess 

I understand that you can also set this at the vhost or server level so that you do not need to repeat the configuration for each directory.  It may make sense to add another new parameter to both the `apache` and `apache::vhost` classes and the corresponding templates to allow setting this parameter at any supported location.

> But another thing, that I realized, is that the options are only applied for provider == Directory (templates/vhost/_directories.erb line 33) which I think is wrong (at least the Index options seem to work for provider == Location on apache but not in puppet).

I am not sure to follow. `<Location ...>`, `<Directory ...>` etc are handled by the module through the `directories` parameter where each entry indicate which `provider` to use.  It probably allows to build invalid configuration for apache that is accepted by Puppet (e.g. adding a `IndexHeadInsert` into a `<Proxy ...>`), but it goes beyond the scope of this PR and I don't think we should worry about this.

jst-cyr:
Hi @trefzer, my apologies that this hasn't gotten addressed yet. I've been going through older PRs, came across this one, and wanted to know if this is still a feature you are needing to be added to the apache module?

trefzer:
@jst-cyr  of course it's still in my custom branch of this module that I use for production. So yes I appreciate merging this !

jst-cyr:
Thank you @trefzer , then I'll leave this PR open for now.

PR #2545, opened 2024-05-04
Finish conversion to EPP templates

smortex:
A lot of work was done to convert the module templates form ERB to EPP, but a few templates where still to be converted.

Along with various benefits, EPP templates offer better detection for access to undefined variables.  This refactoring therefore fix a few issues that where reported while converting.  Also a bunch of outdated comments about which template use which variable where removed no that this usage is explicit.

The extensive test suite helped ensure the conversion was not introducing regressions.


manifests/mod/php.pp
```diff
@@ -32,7 +32,7 @@
   Optional[String] $path         = undef,
   Array $extensions              = ['.php'],
   Optional[String] $content      = undef,
-  String $template               = 'apache/mod/php.conf.erb',
+  String $template               = 'apache/mod/php.conf.epp',
```
smortex, line 35: This change should be backwards-compatible. See related commit for details.

bastelfreak (dismissed):

templates/mod/status.conf.erb
```diff
@@ -1,14 +0,0 @@
...
-    <%-# "scope.function_template(["apache/mod/<Template>"])" -%>
-    <%-# with: -%>
-    <%-# "scope.call_function('template', ["apache/mod/<Template>"])" -%>
-<%= scope.function_template(["apache/mod/_require.erb"]) -%>
-</Location>
-ExtendedStatus <%= @extended_status %>
-
-<IfModule mod_proxy.c>
-    # Show Proxy LoadBalancer status in mod_status
-    ProxyStatus On
```
Dan33l, line 13: Probably not critical and perhaps i missed something.
I did not found in epp templates a place where `ProxyStatus` is handled.

Dan33l (commented):
Huge PR. I tried to add my eyes, and added an inline comment.

templates/mod/status.conf.erb
smortex: `ProxyStatus` is handled in the EPP template just like in the ERB one: no direct user control on this setting, when mod_status is enabled, and if mod_proxy is also loaded, the setting is enabled.

In these mod_status files, the conversion to EPP was started but not finished (the template was not used), so I just completed to pass the expected parameters, removed the legacy comments that caused extra space to be output with EPP.  The GitHub diff is cut before the relevant lines, but clicking on `...` you should see the original code :wink:

binford2k:
Hallo @smortex! Fill out this form if you'd like to claim your May the Source Be With You sticker! https://forms.office.com/r/Cn55uJmWMH

templates/mod/_require.epp
ekohl, line 1: Unrelated, but wtf is `String(type($_requires, 'generalized')).index('Array') == 0`? Isn't that equal to `$_requires =~ Array`?

Would it make sense to also use native types to enforce a structure? I think the resulting template is easier to understand:
```epp
<%- | Variant[
        String[0],
        Array[String[1]],
        Struct[
          {
            enforce  => Optional[Enum['all', 'none', 'any']],
            requires => Array[String[1]],
          },
        ],
      ] $requires,
| -%>
<% if $requires.downcase in ['', 'unmanaged'] { -%>
<% elsif $requires =~ String { -%>
    Require <%= $requires %>
<% } elsif $requires =~ Array -%>
  <%- $requires.each |$req| { -%>
    Require <%= $req %>
  <%- } -%>
<% } $requires =~ Hash { -%>
  <%- if $requires['enforce'] { -%>
    <%- $enforce_str = "Require${requires['enforce'].capitalize}>\n" -%>
    <%- $enforce_open = "    <${enforce_str}" -%>
    <%- $enforce_close = "    </${enforce_str}" -%>
    <%- $indentation = '    ' -%>
  <%- } else { -%>
    <%- $enforce_open = '' -%>
    <%- $enforce_close = '' -%>
    <%- $indentation = '' -%>
  <%- } -%>
<%# %><%= $enforce_open -%>
      <%- $requires['requires'].each |$req| { -%>
<%# %>    <%= $indentation -%>Require <%= $req %>
      <%- } -%>
<%# %><%= $enforce_close -%>
<% } -%>
```

I'd even be tempted to rewrite the last bit to avoid variables:
```epp
<% } $requires =~ Hash { -%>
  <%- if $requires['enforce'] { -%>
    <Require${requires['enforce'].capitalize}>
      <%- $requires['requires'].each |$req| { -%>
        Require <%= $req %>
      <%- } -%>
    </Require${requires['enforce'].capitalize}>
  <%- } else { -%>
      <%- $requires['requires'].each |$req| { -%>
    Require <%= $req %>
      <%- } -%>
  <%- } -%>
<% } -%>
```

templates/mod/status.conf.epp
```diff
@@ -1,10 +1,6 @@
 <Location <%= $status_path %>>
     SetHandler server-status
-    <%# From Puppet 4.2 up, replace: -%>
-    <%# "scope.function_template(["apache/mod/<Template>"])" -%>
-    <%# with: -%>
-    <%# "scope.call_function('template', ["apache/mod/<Template>"])" -%>
-<%= epp("apache/mod/_require.epp") -%>
+<%= epp("apache/mod/_require.epp", { 'requires' => $requires, 'requires_defaults' => $requires_defaults }) -%>
```
ekohl, line 3: So this would become:
```suggestion
<%= epp("apache/mod/_require.epp", { 'requires' => pick($requires, $requires_defaults) }) -%>
```

templates/vhost/_php.epp
```diff
@@ -0,0 +1,16 @@
+<% if $php_values and !$php_values.empty { -%>
+  <%- $php_values.stdlib::sort_by |$m| { $m[0] }.each |$key, $value| { -%>
```
ekohl, line 2: Isn't hash sorting always by key anyway?

templates/vhost/_php.epp
```diff
@@ -0,0 +1,16 @@
...
+    <%- } else { -%>
+  php_value <%= $key %> <%= $value %>
+    <%- } -%>
+  <%- } -%>
+<% } -%>
+<% if $php_flags and !$php_flags.empty { -%>
+  <%- $php_flags.stdlib::sort_by |$m| { $m[0] }.each |$key, $flag| { -%>
+    <%# normalize flag -%>
+    <%- if $flag.downcase =~ /true|yes|on|1/ { $_flag = 'on' } else { $_flag = 'off' } -%>
+  php_flag <%= $key %> <%= $_flag %>
```
ekohl, line 14: I think you can avoid the additional variable with little loss of readability here
```suggestion
  php_flag <%= $key %> <%= if $flag.downcase =~ /true|yes|on|1/ { 'on' } else { 'off' } %>
```

templates/vhost/_php_admin.epp
```diff
@@ -0,0 +1,12 @@
+<% if $php_admin_values and !$php_admin_values.empty { -%>
+  <%- $php_admin_values.stdlib::sort_by |$m| { $m[0] }.each |$key, $value| { -%>
+  php_admin_value <%= $key %> <%= $value %>
+  <%- } -%>
+<% } -%>
+<% if $php_admin_flags and !$php_admin_flags.empty { -%>
+  <%- $php_admin_flags.stdlib::sort_by |$m| { $m[0] }.each |$key, $flag| { -%>
+    <%# normalize flag -%>
+    <%- if $flag.downcase =~ /true|yes|on|1/ { $_flag = 'on' } else { $_flag = 'off' } -%>
```
ekohl, line 9: Given the repetition, perhaps a helper would be useful? There's a lot of overlap with `$_php.epp` anyway so other comments I had there apply here too.

templates/vhost/_setenv.epp
```diff
@@ -0,0 +1,17 @@
+<% unless $setenv.empty { -%>
```
ekohl, line 1: Can we move the `empty()` check to `vhost.pp` and only include the concat fragment if it's non-empty? In fact, we already have `$use_env_mod = !empty($setenv)` so you can reuse that variable.

templates/vhost/_ssl.epp
```diff
@@ -0,0 +1,68 @@
+<% if $ssl { -%>
```
ekohl, line 1: Like `_setenv.epp`, I'd prefer to keep the `if` in the Puppet code and avoid rendering the whole template.

ekohl (commented):
I wish we had more type enforcement in templates and EPP allows that. If this is backwards incompatible anyway, it may be a good moment to do so.

manifests/mod/security.pp
```diff
@@ -316,49 +316,30 @@
...
-      'content_types'               => $content_types,
-      'restricted_extensions'       => $restricted_extensions,
-      'restricted_headers'          => $restricted_headers,
-      'paranoia_level'              => $paranoia_level,
-      'executing_paranoia_level'    => $executing_paranoia_level,
-      'enable_dos_protection'       => $enable_dos_protection,
-      'dos_burst_time_slice'        => $dos_burst_time_slice,
-      'dos_counter_threshold'       => $dos_counter_threshold,
-      'dos_block_timeout'           => $dos_block_timeout,
+      '_secdefaultaction'          => $_secdefaultaction,
```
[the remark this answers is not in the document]
smortex: I decided to go this way in order to limit the changes to just ERB -> EPP. IMHO, we should pass `secdefaultaction => $secdefaultaction` and move the code that does the transformation into the template, but it would be hard to track the move with all other changes, and as you say this kind of fix would be welcome in other places.

So for now, my proposal is to keep it that way.

templates/vhost/_setenv.epp
smortex: Yeah, I saw that but kept it as it was before.  I will add a commit to avoid the double-test.

manifests/mod/security.pp
```diff
@@ -316,49 +316,30 @@
...
+      'allowed_methods'            => $allowed_methods,
+      'content_types'              => $content_types,
+      'restricted_extensions'      => $restricted_extensions,
+      'restricted_headers'         => $restricted_headers,
+      'paranoia_level'             => $paranoia_level,
+      'executing_paranoia_level'   => $executing_paranoia_level,
+      'enable_dos_protection'      => $enable_dos_protection,
+      'dos_burst_time_slice'       => $dos_burst_time_slice,
+      'dos_counter_threshold'      => $dos_counter_threshold,
+      'dos_block_timeout'          => $dos_block_timeout,
```
sbrowne-godaddy, line 337: For CRS2, the `$security_crs_parameters` also needs a `anomaly_score_blocking` defined (oddly, as shown in the diff of the `security_crs.conf.epp`).

The following additional line in this block fixed it for me.

```
      'anomaly_score_blocking'     => downcase($modsec_secruleengine),
```

The rule is only ever for the lowercase string of `on`, so the defaults pass through cleanly.

jst-cyr:
@smortex : It looks like there is quite a bit of feedback to be resolved on this PR, and conflicts with the current main branch. Is this change still needed? Otherwise, perhaps we should close it off until needed in the future?

jst-cyr:
Closing this PR for now until a need rises for this in the future at which point we can re-open!

PR #2552, opened 2024-06-07
Stop using apache::apache_version

ekohl:
In cedd45b63be89ea54bd2a596e6cd3a3f60d4faf8 this was dropped and the whole module is now only for Apache 2.4+. This means the parameter is redundant. The parameter is kept in so this can be released in a bugfix release, but the next major version should drop it.

Fixes: cedd45b63be89ea54bd2a596e6cd3a3f60d4faf8 ("Drop Apache 2.2 support")


bastelfreak (approved):

jst-cyr:
Hey @ekohl and @bastelfreak ... Found this old PR that got approved but never merged up. Do we think this could still be useful in the apache module? Or should we close this out?

ekohl:
Probably still a good cleanup because it simplifies the code and makes sure the tests actually run.

jst-cyr:
Alrighty! I will rebase, get it into the list for engineering, and kick off some acceptance testing just to see where we are at with the latest codebase.

jst-cyr:
Reran the unit test CI and there are some failures, so I haven't kicked off the full acceptance testing. Those will need to be addressed first.

PR #2553, opened 2024-06-10
Wsgi hash documentation

### LINKED ISSUE #2537: using wsgi_daemon_process_options logs a deprecation warning but the docs only show the deprecated usage
With any recent version of `puppetlabs-apache`, in `manifests/vhost.pp` there is this section of code:

```
  if $wsgi_daemon_process_options {
    deprecation('apache::vhost::wsgi_daemon_process_options', 'This parameter is deprecated. Please add values inside Hash `wsgi_daemon_process`.')
  }
```

This triggers a (huge) deprecation warning log message in the puppetserver.log.

The intent seems to be to switch from using the related `wsgi_daemon_process` option as a string, and instead make _it_ a Hash, likely with the key/value pairs that were being passed in `wsgi_daemon_process_options`.

The problem is that the documentation doesn't reflect the deprecation.  The reference documentation only shows the deprecated usage, not the preferred usage.

Without updated documentation, it's not clear how the String that was the argument to the deprecated use of `wsgi_daemon_process` is supposed to be incorporated into preferred format Hash.

virot:
## Summary
The code says wsgi_daemon_process_options is depreciated

Has been since https://github.com/puppetlabs/puppetlabs-apache/commit/0d5e0bef88c741e964efc3e0c91fe4764182803f

The documentation for that was reverted with https://github.com/puppetlabs/puppetlabs-apache/commit/98cf15c01cc7467f9d493a479db063b6a8812ec5

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1190,8 +1190,8 @@
 #     docroot                     => '/var/www/pythonapp',
 #     wsgi_daemon_process         => {
 #       'wsgi' => {
-#         processes    => '2',
-#         threads      => '15',
+#         processes    => 2,
+#         threads      => 15,
 #         display-name => '%{GROUP}',
 #       },
 #       'foo' => {},
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -1188,12 +1188,14 @@
 #   apache::vhost { 'wsgi.example.com':
 #     port                        => 80,
 #     docroot                     => '/var/www/pythonapp',
-#     wsgi_daemon_process         => 'wsgi',
-#     wsgi_daemon_process_options =>
-#       { processes    => 2,
-#         threads      => 15,
+#     wsgi_daemon_process         => {
+#       'wsgi' => {
+#         processes    => '2',
+#         threads      => '15',
 #         display-name => '%{GROUP}',
 #       },
+#       'foo' => {},
+#     },
 #     wsgi_process_group          => 'wsgi',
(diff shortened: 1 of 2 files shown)


manifests/vhost.pp
```diff
@@ -1188,19 +1188,22 @@
...
+#       'foo' => {},
+#     },
 #     wsgi_process_group          => 'wsgi',
 #     wsgi_script_aliases         => { '/' => '/var/www/demo.wsgi' },
 #     wsgi_chunked_request        => 'On',
 #   }
 #   ```
 #
 # @param wsgi_daemon_process_options
+#   Depricated, move to wsgi_daemon_process
```
smortex, line 1206:
```suggestion
#   Deprecated, move to wsgi_daemon_process
```

manifests/vhost.pp
```diff
@@ -1188,19 +1188,22 @@
...
 #     port                        => 80,
 #     docroot                     => '/var/www/pythonapp',
-#     wsgi_daemon_process         => 'wsgi',
-#     wsgi_daemon_process_options =>
-#       { processes    => 2,
-#         threads      => 15,
+#     wsgi_daemon_process         => {
+#       'wsgi' => {
+#         processes    => '2',
+#         threads      => '15',
```
smortex, line 1194: Do you know if we can use integers?  If so, this may be preferable for consistency (even if there is no actual check of the data type).

virot:
I updated the documentation to Integers instead of Strings.

PR #2556, opened 2024-07-18
Unify Debian MPM management with "mpm_" prefix

### LINKED ISSUE #2555: mpm_event module management is not consistent
## Describe the Bug
If `mpm_event` is enabled with `apache::mod::event`, using `apache::mpm::disable_mpm_event` will not disable it. Internally, at least on Debian 12, it looks like the `event` and `mpm_event` modules are identical, aside from the `mpm_event` module specifically warning about conflicts with other MPM modules. As a result, I believe it is a bug that the module's method to enable the module is not compatible with the way to disable the module.

Note: I will add a pull request with a patch to fix this behavior; I wanted an issue to link it against.

## Expected Behavior
The expected behavior is that the Puppet module can both enable or disable `mpm_event`, based on variables such as host facts. 

## Steps to Reproduce
1. Apply a manifest to a system that includes `mpm_event`:
```
include apache::mod::event
```
2. Later, apply a different manifest to the same system that disables the same module:
```
include apache::mpm::disable_mpm_event
```
3. Module will still be enabled, as the class to enable uses the name without the `mpm_` prefix, while the class to disable the module does use the `mpm_` prefix.

## Environment
 - Version 12.1
 - Platform Debian 12

## Additional Context
The inconsistency is caused by the module being enabled with `${mpm}` as `event``:
https://github.com/puppetlabs/puppetlabs-apache/blob/58dada69b92cc5b5b9155068db4f82fbb2ae9cac/manifests/mpm.pp#L49
https://github.com/puppetlabs/puppetlabs-apache/blob/58dada69b92cc5b5b9155068db4f82fbb2ae9cac/manifests/mpm.pp#L57

While the module is disabled with:
https://github.com/puppetlabs/puppetlabs-apache/blob/58dada69b92cc5b5b9155068db4f82fbb2ae9cac/manifests/mpm/disable_mpm_event.pp#L3

This is inconsistent with prefork and worker, which are enabled and disabled without the `mpm_` prefix.

chacunda:
## Summary
This change unifies the management of MPM modules using the "mpm_" prefix on the Debian platform. Previously, there were inconsistencies in whether MPM modules were enabled or disabled with the title `$mpm` or `mpm_${mpm}`. The changes to `mpm.pp` and the related individual mod manifests make it so that on the Debian platform, MPM modules are always referred to with the "mpm_" prefix. This is consistent with how the modules are installed and addressed by the Apache package maintainers for the Debian platform.

With this change implemented, an administrator is able to enable or disable MPM modules without introducing conflicts to the Apache service. Prior to this change, if the "prefork" or "worker" modules were enabled under the package maintainer provided name "mpm_worker" or "mpm_prefork", the `disable_mpm_*.pp` would fail to disable the modules. Similarly, if the "mpm_event" module was enabled by this Puppet module, the associated `disable_mpm_event.pp` would fail to disable it, as it was looking for the module under the name "mpm_event," but the module was enabled as "event".

The changes to accomplish this are:

- Adding a switch to `event.pp`, `itk.pp`, `prefork.pp` and `worker.pp` that changes the conf file name to `mpm_*.conf` on Debian hosts.
- Changes `mpm.pp` to create load files that match the name and content as provided by the Debian Apache package maintainers.
- Modifies `mpm.pp` to refer to load and conf files on Debian by a relative symlink and with the "mpm_" prefix, to match behavior of the modules installed via APT and changes made by `a2enmod`.
- Adds mod names with and without "mpm_" prefixes to `disable_mpm_*.pp` to make sure that the module is disabled, regardless of name. Additionally, adds the `mpm_itk` to `disable_mpm_prefork.pp` as `mpm_itk` must be disabled for disabling `mpm_prefork` to succeed.
- All relevant spec tests are updated to pass with the modifications made, specifically, matching the new filenames and new content for the load files.

## Additional Context
The initial problem leading to this fix is that if you use the module to enable `mpm_event` and then later try to enable a different MPM module, applying the catalog will fail. This change fixes that issue, and makes sure behavior and naming is consistent and matches what the Debian Apache package maintainers provide for module load definitions.

## Related Issues (if any)
This resolves https://github.com/puppetlabs/puppetlabs-apache/issues/2555.
## Checklist
- [x] 🟢 Spec tests. All spec tests pass.
- [ ] 🟢 Acceptance tests. Debian 11 and 12 pass, errors about package availability encountered on SLES 15 from the litmusimage repo.
- [x] Manually verified. (For example `puppet apply`) Tested and verified on Debian 12 hosts in my Puppet environment.

### DIFF
--- a/manifests/mod/event.pp
+++ b/manifests/mod/event.pp
@@ -86,7 +86,12 @@ class apache::mod::event (
     'listenbacklog'           => $listenbacklog,
   }
 
-  file { "${apache::mod_dir}/event.conf":
+  $eventconffile = $facts['os']['family'] ? {
+    'Debian' => "${apache::mod_dir}/mpm_event.conf",
+    default  => "${apache::mod_dir}/event.conf",
+  }
+
+  file { $eventconffile:
     ensure  => file,
     mode    => $apache::file_mode,
     content => epp('apache/mod/event.conf.epp', $parameters),
--- a/manifests/mod/itk.pp
+++ b/manifests/mod/itk.pp
@@ -79,7 +79,12 @@ class apache::mod::itk (
     'enablecapabilities'  => $enablecapabilities,
   }
 
-  file { "${apache::mod_dir}/itk.conf":
+  $itkconffile = $facts['os']['family'] ? {
+    'Debian' => "${apache::mod_dir}/mpm_itk.conf",
+    default  => "${apache::mod_dir}/itk.conf",
+  }
+
+  file { $itkconffile:
     ensure  => file,
     mode    => $apache::file_mode,
     content => epp('apache/mod/itk.conf.epp', $parameters),
--- a/manifests/mod/prefork.pp
+++ b/manifests/mod/prefork.pp
@@ -79,7 +79,12 @@ class apache::mod::prefork (
     'listenbacklog'           => $listenbacklog,
   }
 
-  file { "${apache::mod_dir}/prefork.conf":
+  $preforkconffile = $facts['os']['family'] ? {
+    'Debian' => "${apache::mod_dir}/mpm_prefork.conf",
+    default  => "${apache::mod_dir}/prefork.conf",
+  }
+
+  file { $preforkconffile:
     ensure  => file,
     content => epp('apache/mod/prefork.conf.epp', $parameters),
     require => Exec["mkdir ${apache::mod_dir}"],
--- a/manifests/mod/worker.pp
+++ b/manifests/mod/worker.pp
@@ -90,7 +90,12 @@ class apache::mod::worker (
     'maxrequestworkers'   => $maxrequestworkers,
   }
 
-  file { "${apache::mod_dir}/worker.conf":
+  $workerconffile = $facts['os']['family'] ? {
+    'Debian' => "${apache::mod_dir}/mpm_worker.conf",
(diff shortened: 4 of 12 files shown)


jst-cyr:
Hi @chacunda , I'm sorry that there hasn't been a response on this PR before now. I'm going through older PRs and wondering if this is still an issue that you need resolved?

chacunda:
Hi @jst-cyr, thanks for the response. I have been running my team's Puppet instance with these changes since I submitted this PR originally, so I don't have an immediate need for any action from the upstream repository. If there aren't resources available to validate and merge this change, I wouldn't consider it a priority.

However, I still think the change to unify how the manifests enable and disable MPM modules on Debian would be valuable for anyone who manages workstations or other frequently changing hosts running Apache via Puppet. If you have any questions or would like me to otherwise update the change to prepare for a merge given the passage of time and changes to the upstream code, I can do so. Please let me know.

jst-cyr:
Thank you for the offer @chacunda ! I've pulled in a rebase and will rerun the CI so we can do a check of these changes against latest. If the CI looks good, we can get some acceptance tests running and then eventually the engineering review.

Thanks again!

jst-cyr:
@chacunda : Looks like the spec tests found some syntax issues.

PR #2559, opened 2024-08-16
Automatically include mod_expires if required via vhost directories

ekohl:
## Summary

Similar to other options passed in the vhost's directory this now inspects the directories entry and if expires_* is passed, mod_expires` is included.

## Additional Context

This is a pattern that I've started to apply more and more. It allows passing complex directories (also via Hiera). On our servers we disable the default mods and this ensures we have the correct mods enabled.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


smortex (approved):

manifests/vhost.pp
```diff
@@ -2362,6 +2362,10 @@
       if 'setenv' in $directory {
         include apache::mod::env
       }
+
+      if 'expires_active' in $directory or 'expires_default' in $directory or 'expires_by_type' in $directory {
+        include apache::mod::expires
```
ekohl, line 2367: One thing to note: `apache::mod::expires` by default writes down a configuration file (unlike other modules). That could be a behavior difference and break setups. Is that a problem?

evgeni: Don't we have other modules that do that and thus we include them differently?

evgeni: Like
```
::apache::mod { 'expires': }
```

ekohl: That is exactly the point I tried to make and I'm surprised that CI didn't fail on this:
https://github.com/puppetlabs/puppetlabs-apache/blob/6d218cfcbcb1c58cbf77ccd5669524792172c7c0/manifests/default_mods.pp#L63
https://github.com/puppetlabs/puppetlabs-apache/blob/6d218cfcbcb1c58cbf77ccd5669524792172c7c0/manifests/default_mods.pp#L93

It feels ugly, but perhaps we should have some helper class that is only the module definition without the configuration file.

Issue #2561, opened 2024-09-03
Service fails to start on Amazon Linux 2023


rjd1:
## Describe the Bug
The httpd service fails to start on Amazon Linux 2023 due to the systemd module not being present.

## Expected Behavior
The service is able to start and run normally via systemd. 

## Steps to Reproduce
Include the apache class on AL2023.

## Environment
 - Version: 12.1.0
 - Platform: Amazon Linux 2023

## Additional Context
There is existing logic to include the systemd module specifically for Amazon Linux 2. We can simply modify it a bit to include on AL2023 as well.

daveseff:
+1 Please merge soon

PR #2562, opened 2024-09-03
Include systemd module for Amazon Linux 2023

rjd1:
## Summary
Include the systemd module for both AL2 and AL2023.

## Additional Context
This resolves an issue where the service fails to start on AL2023.

## Related Issues (if any)
https://github.com/puppetlabs/puppetlabs-apache/issues/2561

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x ] Manually verified. (For example `puppet apply`)

### DIFF
--- a/manifests/default_mods.pp
+++ b/manifests/default_mods.pp
@@ -16,7 +16,7 @@ class apache::default_mods (
       if $facts['os']['name'] != 'Amazon' and $use_systemd {
         ::apache::mod { 'systemd': }
       }
-      if ($facts['os']['name'] == 'Amazon' and $facts['os']['release']['full'] == '2') {
+      if ($facts['os']['name'] == 'Amazon' and $facts['os']['release']['major'] =~ /^(2|2023)$/) {
         ::apache::mod { 'systemd': }
       }
       ::apache::mod { 'unixd': }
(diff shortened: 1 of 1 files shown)


manifests/default_mods.pp
```diff
@@ -16,7 +16,7 @@
       if $facts['os']['name'] != 'Amazon' and $use_systemd {
         ::apache::mod { 'systemd': }
       }
-      if ($facts['os']['name'] == 'Amazon' and $facts['os']['release']['full'] == '2') {
+      if ($facts['os']['name'] == 'Amazon' and $facts['os']['release']['major'] =~ /^(2|2023)$/) {
```
bastelfreak, line 19: does it make more sense to always include it when systemd is the default init system?

rjd1: Yeah probably @bastelfreak 
I think the original logic here was added to support AL1 & 2, with the former not using systemd. But since that's been EOL for a while now, the Amazon specific code shouldn't really be needed anymore.

rjd1: I went ahead and updated the PR.

ekohl (approved):

daveseff:
Can this get merged soon please?

malikparvez (approved):

PR #2583, opened 2024-11-28
Adding ModSecurity parameter for audit log format.

Tamerz:
## Summary
Adds a parameter to set the audit log format to `JSON`

https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#user-content-SecAuditLogFormat

## Additional Context
Since as of the current version of ModSecurity, this can only have two values `Native` and `JSON`, and `Native` is default, I decided only add the line if someone specifically sets the value to `JSON`. This means it shouldn't touch any existing config, especially on versions where this parameter may not exist.

## Checklist
- [X] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [X] Manually verified. (For example `puppet apply`)

### DIFF
--- a/manifests/mod/security.pp
+++ b/manifests/mod/security.pp
@@ -34,7 +34,7 @@
 #   Defines the type of audit logging mechanism to be used.
 #
 # @param audit_log_format
-#   Defines what format the logs should be written in. Accepts `Native` and `JSON`.
+#   Defines what format the logs should be written in.
 # 
 # @param audit_log_storage_dir
 #   Defines the directory where concurrent audit log entries are to be stored. This directive is only needed when concurrent audit logging is used.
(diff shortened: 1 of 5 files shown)


manifests/params.pp
```diff
@@ -36,6 +36,7 @@
 
   $modsec_audit_log_parts = 'ABIJDEFHZ'
   $modsec_audit_log_type = 'Serial'
+  $modsec_audit_log_format = 'Native'
```
smortex, line 39: Don't put the default value in params.pp

manifests/mod/security.pp
```diff
@@ -143,6 +146,7 @@
   String $audit_log_relevant_status                            = '^(?:5|4(?!04))',
   String $audit_log_parts                                      = $apache::params::modsec_audit_log_parts,
   String $audit_log_type                                       = $apache::params::modsec_audit_log_type,
+  Enum['Native', 'JSON'] $audit_log_format                     = $apache::params::modsec_audit_log_format,
```
smortex, line 149: Put the default value here instead.

That way, the default value is correctly documented in the generated documentation.

manifests/mod/security.pp
```diff
@@ -32,6 +32,9 @@
 # 
 # @param audit_log_type
 #   Defines the type of audit logging mechanism to be used.
+#
+# @param audit_log_format
+#   Defines what format the logs should be written in. Accepts `Native` and `JSON`.
```
smortex, line 37: No need to list accepted values, as the documentation include the data type (and the default value).

smortex (commented):
LGTM but see inline comment for some doc improvements.  Thank!

smortex (approved):
LGTM, I regenerated REFERENCE.md so that it reflect your changes. Thanks!

PR #2586, opened 2025-01-09
Allow to use the same domain multiple times with mod_md

smortex:
When using mod_md to manage TLS certificates, a domain can only appear
once as a parameter of a MDomain configuration.

When a single node configue multiple Virtual Hosts to serve the same
website on different IP Addresses or on different ports, and we want to
use mod_md to manage the TLS certificate, the current code produce a
MDomain entry in each virtual host, leading to configuration error and
preventing apache from starting.

This commit rework how the MDomain setting is emitted, and ensure it is
only output once even if multiple Virtual Hosts configure the same
domain.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2239,7 +2239,6 @@ define apache::vhost (
   $file_header_params = {
     'comment'               => $comment,
     'nvh_addr_port'         => $nvh_addr_port,
-    'mdomain'               => $mdomain,
     'servername'            => $servername,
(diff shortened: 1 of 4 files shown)


amitkarsale:
@smortex can you rebase this PR please.

smortex:
> @smortex can you rebase this PR please.

@amitkarsale done.  Looks like SLES 15 packages are broken (does not find `/usr/sbin/apache2ctl` in CI)?

Issue #2588, opened 2025-02-25
Display of RewriteRule in vhost config files


teluq-pbrideau:
## Use Case
When I define multiple `RewriteRule`, they are not very well displayed in the config file, there is too much space between them and there is no sense they are related to each other…
```
  apache::vhost { 'example.com':
    rewrites => [
      {
        comment => 'Example rules 1',
        rewrite_rule => [
          '"\/(foo\/)" - [F]',
          '"\/(bar\/)" - [F]',
          '"\/(baz\/)" - [F]',
          '"\/(qux\/)" - [F]',
          '"\/(quux\/)" - [F]',
        ],
      },
      {
        comment => 'Example rules 2',
        rewrite_rule => [
          '"\/(another_foo\/)" - [F]',
          '"\/(another_bar\/)" - [F]',
          '"\/(another_baz\/)" - [F]',
          '"\/(another_qux\/)" - [F]',
          '"\/(another_quux\/)" - [F]',
        ],
      },
    ],
  }
```

```
[...]
  # Example rules 1
  RewriteRule "(\/foo\/)" - [F]

  RewriteRule "(\/bar\/)" - [F]

  RewriteRule "(\/baz\/)" - [F]

  RewriteRule "(\/qux\/)" - [F]

  RewriteRule "(\/quux\/)" - [F]

  # Example rules 2
  RewriteRule "(\/another_foo\/)" - [F]

  RewriteRule "(\/another_bar\/)" - [F]

  RewriteRule "(\/another_baz\/)" - [F]

  RewriteRule "(\/another_qux\/)" - [F]

  RewriteRule "(\/another_quux\/)" - [F]
```

## Describe the Solution You Would Like
I feel these `RewriteRule` should not be this much spaced.  I think this would be much more readable:
```
  # Example rules 1
  RewriteRule "(\/foo\/)" - [F]
  RewriteRule "(\/bar\/)" - [F]
  RewriteRule "(\/baz\/)" - [F]
  RewriteRule "(\/qux\/)" - [F]
  RewriteRule "(\/quux\/)" - [F]

  # Example rules 2
  RewriteRule "(\/another_foo\/)" - [F]
  RewriteRule "(\/another_bar\/)" - [F]
  RewriteRule "(\/another_baz\/)" - [F]
  RewriteRule "(\/another_qux\/)" - [F]
  RewriteRule "(\/another_quux\/)" - [F]
```

## Describe Alternatives You've Considered
[...]

PR #2589, opened 2025-02-25
fix: better spacing for RewriteRule

teluq-pbrideau:
## Summary
Fix spacing for RewriteRule

## Related Issues (if any)
#2588 

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


smortex (approved):
+1 for generating better VHost even if this is somewhat a breaking change.


Testing the whole generated fragments as a single test instead of each line as currently done in this module and a lot of other puppet modules would not only run faster, but make it obvious that the generated config is changed.  It would also help spotting this kind of issues when they are introduced, or spot unexpected changes when reworking templates.

PR #2593, opened 2025-03-10
Install mod_http2 on EL if required

ekohl:
## Summary

In EL8 httpd had a hard requirement on mod_http2 but in EL9 it turned into a weak dependency. While most installs default to installing weak dependencies, it can be disabled by users (by setting `install_weak_deps=0` in `/etc/dnf/dnf.conf`). OracleLinux even defaults to disabled.

EL7 doesn't have http2 at all so the whole class can't be used with it.

## Additional Context
- [x] Root cause and the steps to reproduce. (If applicable)
- [x] Thought process behind the implementation.

Still a draft, but having a PR makes it easy for me to test.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -202,6 +202,7 @@ class apache::params inherits apache::version {
       'authnz_pam'            => 'mod_authnz_pam',
       'fcgid'                 => 'mod_fcgid',
       'geoip'                 => 'mod_geoip',
+      'http2'                 => 'mod_http2',
       'intercept_form_submit' => 'mod_intercept_form_submit',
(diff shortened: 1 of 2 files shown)


kenyon (approved):

smortex (approved):

bastelfreak (dismissed):

smortex (dismissed):

ekohl:
Trivial rebase done. I'd like to get this merged now.

bastelfreak (approved):

PR #2337, opened 2022-10-13
Automatically enable mod_http2 if needed

ekohl:
Following https://httpd.apache.org/docs/current/howto/http2.html the first step is to load the http2 module. By inspecting the protocols argument users can enable HTTP/2 via Hiera.

PR #2598, opened 2025-04-04
Detect module dependencies when modules are not default-loaded

gcoxmoz:
## Additional Context
If you accept `default_mods => true` you get a -lot- of apache modules automatically.  In my experience it's too many, which has led to us doing `default_mods => []` to reduce the sprawl, and then adding back in what we need.  That leads to a bit of whack-a-mole upon new server buildout, as a lot of this module's code assumes that you have included the default modules and it doesn't need to check.  It also leads to "I'm not sure if I can remove this module" later, if you remove directives.

## Summary
This covers off the worst offenders I've hit in our environment: `mod_dir`, `mod_autoindex`, `mod_expires`, `mod_dav`, and `mod_cgi`.  If you use vhost directory parameters that will invoke these module directives, vhost will try to make sure the module is loaded.  If you're on `default_mods => true`, this is a 'wasted' extra load; if you don't have the module loaded (say, `default_mods => false`) then this brings the module in for you.

## Related Issues (if any)
N/A

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2340,6 +2340,10 @@ define apache::vhost (
         include apache::mod::dir
       }
 
+      if 'expires_active' in $directory {
+        include apache::mod::expires
+      }
+
       if 'gssapi' in $directory {
         include apache::mod::auth_gssapi
       }
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2337,18 +2337,16 @@ define apache::vhost (
       }
 
       if 'options' in $directory {
-        if !('-ExecCGI' in $directory['options']) {
-          if 'ExecCGI' in $directory['options'] {
-            case $apache::mpm_module {
-              'prefork': {
-                include apache::mod::cgi
-              }
-              'worker': {
-                include apache::mod::cgid
-              }
-              default: {
-                # do nothing
-              }
+        if !('-ExecCGI' in $directory['options']) and 'ExecCGI' in $directory['options'] {
+          case $apache::mpm_module {
+            'prefork': {
+              include apache::mod::cgi
+            }
+            'worker': {
+              include apache::mod::cgid
+            }
+            default: {
+              # do nothing
             }
           }
         }
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2336,6 +2336,24 @@ define apache::vhost (
         include apache::mod::authz_groupfile
       }
 
+      if 'options' in $directory {
+        if !('-ExecCGI' in $directory['options']) {
+          if 'ExecCGI' in $directory['options'] {
+            case $apache::mpm_module {
+              'prefork': {
+                include apache::mod::cgi
+              }
+              'worker': {
+                include apache::mod::cgid
+              }
+              default: {
+                # do nothing
+              }
+            }
+          }
+        }
+      }
+
       if 'dav' in $directory {
         include apache::mod::dav
         if $directory['dav'] == 'On' {
--- a/manifests/vhost.pp
+++ b/manifests/vhost.pp
@@ -2357,6 +2357,10 @@ define apache::vhost (
         include apache::mod::auth_gssapi
       }
 
+      if 'index_options' in $directory or 'index_order_default' in $directory or 'index_style_sheet' in $directory {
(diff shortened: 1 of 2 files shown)


binford2k:
Should this magically load dependencies, or should it display errors so people know to correct their code? I'm not sure I have a strong opinion there. Perhaps it could load, but also print an `INFO` message.

gcoxmoz:
There's plenty of prior art of the puppet-module "acting on your behalf" in grabbing needed apache-modules.  One for a quick example:  https://github.com/puppetlabs/puppetlabs-apache/blob/1c720758bf4ac605c58a1cb0e1f181566af377b8/manifests/vhost.pp#L2261-L2263 is "if your vhost directive asked for `virtual_docroot` (which defaults to `false`), jump in and grab the apache-module that will make it work."  This parallels what I'm looking for: 'demand an apache module when it is needed'.  The difference is that a lot of the puppet module is built on the assumption that some of the apache modules are always there... and they don't have to be.  Which is actually a bit odd, since the spec test https://github.com/puppetlabs/puppetlabs-apache/blob/1c720758bf4ac605c58a1cb0e1f181566af377b8/spec/defines/vhost_spec.rb#L11 assumes `default_mods => false` while most of the code is built with an unspoken assumption of `default_mods => true`

That all happens without an `INFO` or an explicit apache-module declaration, so IMO autoloading the apache-modules (quietly, as needed) is the way to go here.

binford2k (approved):

manifests/vhost.pp
```diff
@@ -2335,10 +2336,49 @@
...
+                # do nothing
+              }
+            }
+          }
+        }
+      }
+
+      if 'dav' in $directory {
+        include apache::mod::dav
+        if $directory['dav'] == 'On' {
```
ekohl, line 2359: I'm not sure we really enforce this to be `On`. Could be `on` or a boolean too, right? Would it be better to reverse the logic and first check for `svn` and then:
```suggestion
        if bool2httpd($directory['dav']) == 'On' {
```

ekohl (commented):
I've been pushing in this direction. Like https://github.com/puppetlabs/puppetlabs-apache/pull/2255 and https://github.com/puppetlabs/puppetlabs-apache/pull/2337. Exactly for the reason you mention: I want to run the server with `default_mods => false` but not think about which modules to include.

My ideal is that we actually define a proper data type for `$directories` and enforce it.

ekohl (commented):
I want to let you know that I've essentially stopped reviewing anything that is Perforce maintained since they are very unclear on outside maintainers. There used to be https://puppet.com/ecosystem/trusted-contributors/ but that's now gone.

jst-cyr:
@gcoxmoz : Thanks for the PR! Sorry for the long delays in getting to this. I'm going through old PRs and wanted to check if this was still an issue for you? It seems like a helpful change, but wanted to double-check.

gcoxmoz:
Yes, still valid.   I re-rebased up to main.

jst-cyr:
Running CI for this PR here: https://github.com/puppetlabs/puppetlabs-apache/pull/2641

jst-cyr:
Acceptance test results are here: https://github.com/puppetlabs/puppetlabs-apache/actions/runs/31515279283/job/93930997451?pr=2641

Mostly green, with these platforms failing acceptance tests:
- Debian 13: Failed with same `apache2 fails to start via systemd` error as the nightly CI failures against main
- SLES 12: Provisioning succeeded on second run, but still failed with the same errors as seen in nightly CI against main
- SLES 15: Failed with the same `Could not find command '/usr/sbin/apache2ctl` error as the nightly CI failures against main

The PR does not seem to be introducing any new failures from an acceptance testing perspective.

gcoxmoz:
So, like, are we just stuck here til someone greens up main?

jst-cyr:
@gcoxmoz : The step I ran was to get acceptance tests going, which is part of the review process. From my standpoint, tests for this PR don't show any additional issues being introduced, so I think our acceptance testing gate is probably fine for now. We'll likely need to run these again if more changes happen in main and we need to rebase.

The next step is getting the engineering team to do a review of the changes. That could be some time.

PR #2602, opened 2025-08-04
FreeBSD loads apache::mod::info by default - there's no good reason for this

raybellis:
## Summary
On FreeBSD (and only on FreeBSD) `apache::mod::info` is loaded by default.   I can discern no good reason for why this should be the case, and it's trivial to add if required.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [X] Manually verified. (For example `puppet apply`)

### DIFF
--- a/manifests/default_mods.pp
+++ b/manifests/default_mods.pp
@@ -68,7 +68,6 @@ class apache::default_mods (
         include apache::mod::authn_core
         include apache::mod::filter
         include apache::mod::headers
-        include apache::mod::info
         include apache::mod::mime_magic
         include apache::mod::reqtimeout
         include apache::mod::rewrite


raybellis:
By and large we're fine with the default modules, but I found an issue specifically with `apache::mod::info` where it interacts with other features leaving it wide open ignoring any IP ACL.   I've sent a report to security@perforce.

shubhamshinde360:
Hi @raybellis,
I think this looks good.

I have created https://github.com/puppetlabs/puppetlabs-apache/pull/2610 with you as co-author since the nightlies didn't run for this PR and also added some info for changelog when releasing.

Thank you for your contribution!

PR #2628, opened 2026-05-05
Fix SELinux context for Apache configuration files

archanaserver:
Add `seltype => 'httpd_config_t'` to all file resources that create Apache httpd configuration files. This ensures files are created with the correct SELinux context on RHEL-based systems.

Without this fix, files inherit `etc_t` context from parent directory instead of `httpd_config_t`, causing SELinux policy violations.

### DIFF
--- a/manifests/mod.pp
+++ b/manifests/mod.pp
@@ -126,6 +126,7 @@ define apache::mod (
     owner   => 'root',
     group   => $apache::params::root_group,
     mode    => $apache::file_mode,
+    seltype => 'httpd_config_t',
     content => epp('apache/mod/load.epp', $parameters),
     require => [
       Package['httpd'],
@@ -144,6 +145,7 @@ define apache::mod (
       owner   => 'root',
       group   => $apache::params::root_group,
       mode    => $apache::file_mode,
+      seltype => 'httpd_config_t',
       require => [
         File[$_loadfile_name],
         Exec["mkdir ${enable_dir}"],
@@ -162,6 +164,7 @@ define apache::mod (
         owner   => 'root',
         group   => $apache::params::root_group,
         mode    => $apache::file_mode,
+        seltype => 'httpd_config_t',
         require => [
           File["${mod}.conf"],
           Exec["mkdir ${enable_dir}"],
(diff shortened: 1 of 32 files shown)


amitkarsale:
Hi @archanaserver,

Thanks for putting this together — appreciate the effort. After going through it carefully though, I don't think we can take it as-is. Sharing the reasoning so it's useful if you want to revise.

### The cross-platform issue

`httpd_config_t` is defined by the SELinux **targeted policy on RHEL-family distributions** — it doesn't apply elsewhere. This module supports Debian/Ubuntu (AppArmor, not SELinux), FreeBSD (no SELinux), and Gentoo/SUSE (different policies). Applying the type unconditionally pushes a Red Hat-specific string onto every supported platform; on Debian's optional SELinux refpolicy, `httpd_config_t` isn't even guaranteed to be a defined type and could fail rather than silently no-op. That's the main blocker.

### Even on RHEL, this is likely solving a non-problem

The RHEL targeted policy already auto-assigns `httpd_config_t` to files created under `/etc/httpd(/.*)?`, `/etc/httpd/conf.d(/.*)?`, and `/etc/httpd/conf.modules.d(/.*)?` via path-based file context rules. On a stock install with the policy intact, Puppet-managed files in those paths already get the right type. If you were seeing `etc_t`, something specific to your environment caused it — worth tracking down the root cause.

### How to fix it in your environment without changing the module

Two standard options that don't require touching upstream:

1. **Re-apply the policy:** `restorecon -Rv /etc/httpd/` relabels every Apache config file per the existing policy. Wrap it in your own profile as an `Exec` refreshed by `Class['apache']`.
2. **Non-standard path?** Teach the policy first: `semanage fcontext -a -t httpd_config_t "/your/path(/.*)?"` then `restorecon`. The `voxpupuli/puppet-selinux` module wraps both as resource types if you'd rather keep it declarative.

That keeps the SELinux logic in your site-specific code where it belongs, rather than pinning it into the upstream module.

### If you do want to push something upstream

The module's pattern for OS-specific defaults is a parameter on `apache`, defaulted via `apache::params` per `$facts['os']['family']`. A mergeable version would add an `apache::mod_seltype` parameter (defaulting to `'httpd_config_t'` on RedHat and `undef` elsewhere), reference it once per file resource instead of hardcoding 32 string literals, cover the missing manifests (`vhost.pp`, `init.pp`, `custom_config.pp`, and several `mod/*.pp` files), and include tests.

Closing for now — happy to look at a parameter-based version if you'd like to take another pass.

Thanks again!

PR #2643, opened 2026-08-18
Fix: Ensure mod_md `tls-alpn-01` challenge is usable

JonasVerhofste:
The challenge is already configurable in apache::mod::md, but it needs the `acme-tls/1` protocol to properly work (see "https: Challenges"): https://httpd.apache.org/docs/2.4/mod/mod_md.html

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -774,7 +774,7 @@ Default value: `$apache::params::ports_file`
 
 ##### <a name="-apache--protocols"></a>`protocols`
 
-Data type: `Array[Enum['h2', 'h2c', 'http/1.1']]`
+Data type: `Array[Enum['h2', 'h2c', 'http/1.1', 'acme-tls/1']]`
 
 Sets the [Protocols](https://httpd.apache.org/docs/current/en/mod/core.html#protocols)
 directive, which lists available protocols for the server.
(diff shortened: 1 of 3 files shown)


smortex (approved):
LGTM

JonasVerhofste:
That looks like a CI bug, not a failed test. :sweat_smile:

jst-cyr:
@JonasVerhofste : You are right, the acceptance tests don't run correctly on forks. Initial spec tests look good, I'll run some acceptance tests so we can get platform-specific feedback.

jst-cyr:
Running CI tests for this PR here: https://github.com/puppetlabs/puppetlabs-apache/pull/2644

jst-cyr:
Acceptance test results here: https://github.com/puppetlabs/puppetlabs-apache/actions/runs/32295187776?pr=2644

All failures in PR acceptance testing match existing failures in the Nightly build against main:

- SLES-12 fails during provisioning (SSH connection refused before the host comes up) — same on nightly.
- Debian-13 fails in apache parameters default_confd_files => false doesnt do anything — apache2 fails to start via systemd (mktemp error in the init script) — same on nightly.
- SLES-15 fails in apache::custom_config define valid config adds the config — apache2ctl not found — same on nightly.

These are pre-existing issues on main, not caused by this change.

Next step will require an engineering review when the team is available, but tests look good!

PR #2644, opened 2026-08-19
Fix: Ensure mod_md `tls-alpn-01` challenge is usable

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-apache/pull/2643)


jst-cyr:
Mostly green, failures look like the same as what is happening in the nightly's on main. This PR does not seem to be introducing any new issues.

Feedback will be provided on the PR.

PR #2618, opened 2025-12-18
add Ubuntu 24.04 support

mxr0:
## Summary
add changes needed for Ubuntu 24.04 Support

## Additional Context
module does not work on Ubuntu 24.04

PR #2585, opened 2024-12-30
Add support for Ubuntu 24.04

fetzerms:
## Summary
This adds support for Ubuntu 24.04

## Additional Context
n/a

## Related Issues (if any)
I see #2577 which did not add the acceptance tests nor the determination of the according php version.

## Checklist
- [X] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [X] Manually verified. (For example `puppet apply`)

PR #2646, opened 2026-08-25
add Ubuntu 24.04 support

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-apache/pull/2618)


jst-cyr:
Testing has mostly passed, but specifically the Ubuntu 24.04 testing does not pass. The changes in this PR are not sufficient to add Ubuntu 24.04 support. 

PR #2585 has these changes and an additional change and seems to be passing acceptance tests. I think we should close PR #2618 and go forward with PR #2585 as the base for Ubuntu 24.04.

